PR #43, opened 2023-08-23
fix(audit): event meta list add export result set

krihy:
## Summary

audit meta list lack of export result set

## Solution Description

add export result set to audit meta list


yhilmare (approved):
LGTM

PR #44, opened 2023-08-23
[Fix]: fix get wrong filename when export result set

LuckyPickleZZ:
#### Defect Description

When dump result set on alibaba cloud, an wrong filename would be exported.

#### Fix Solution

Use `File#getName` as filename instead.

#### Testing Instructions

Test env must be alibaba cloud or ob-cloud

#### Workaround

N/A

#### Other Information

N/A


yizhouxw (approved):
LGTM

PR #45, opened 2023-08-23
[Fix]: Generated insert into statement error after executing alter table add column statement

PeachThinking:
## Defect Description
After adding new columns to the table, when inserting table data,  the generated insert into statement still only has the original columns.
image
Backend get `columnList` from cache, so when the user adds a field to a table, the `columnList `does not have the new field. The front end relies on the `columnList` field to obtain the field information of the table, and all subsequent components pass information through this field.

## Fix Solution
Delete `tableColumnsCache` and `viewColumnsCache`.
The cache was used before because the oracle statement querying constraint information was time-consuming, but the current `tryToGetTableColumnsFromCache` method has removed this logic, only querying the column information, and the query of the column information is not time-consuming, so there is no need to use this cache.


LuckyPickleZZ (approved):
LGTM

PR #49, opened 2023-08-24
delete: Remove useless log to keep the log output clean

smallsheeeep:
## Summary

Unnecessary log printing can be removed to keep the log output clean

## Solution Description

Remove useless log to keep the log output clean


yhilmare (approved):
LGTM

PR #51, opened 2023-08-24
build: set eclipse formatter as idea project setting

yizhouxw:
## Summary
set eclipse formatter as idea project setting


yhilmare (approved):
LGTM

PR #54, opened 2023-08-24
Partition plan adaptation to database management.

guowl3:
#### What type of PR is this?
type-feature

Partition plan adaptation to database management.

Fixes #

```docs

```


server/odc-migrate/src/main/resources/migrate/common/V_4_2_0_42__alter_partition_plan.sql
```diff
@@ -0,0 +1,6 @@
+ALTER TABLE `table_partition_plan` MODIFY COLUMN `flow_instance_id` bigint(20) COMMENT '任务 ID' NULL;
```
yizhouxw, line 1: if a column should not be null, use `NOT NULL`

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/PartitionPlanController.java
```diff
@@ -28,38 +27,36 @@
...
 @RestController
-@ConditionalOnProperty(name = "odc.feature.partitionplan.enabled", havingValue = "true")
 public class PartitionPlanController {
 
     @Autowired
     private PartitionPlanService partitionPlanService;
 
-    @RequestMapping(value = "/api/v2/patitionplan/ConnectionPartitionPlan", method = RequestMethod.GET)
-    public SuccessResponse<ConnectionPartitionPlan> getConnectionPartitionPlan(@RequestParam Long connectionId,
+    @RequestMapping(value = "/api/v2/patitionplan/DatabasePartitionPlan/", method = RequestMethod.GET)
```
yizhouxw, line 43: seems no RESTful, what is the relationship between database and partition plan ?

may `api/v2/partitionplan/PartitionPlans/{id}` ?

server/odc-migrate/src/main/resources/migrate/common/V_4_2_0_42__alter_partition_plan.sql
guowl3: done

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/PartitionPlanController.java
guowl3: The API has been modified to 'api/v2/partitionplan/partitionPlans'

krihy (approved):
LGTM

PR #55, opened 2023-08-25
reactor(osc): reactor client result by different error code

krihy:
#### What type of PR is this?

reactor osc code 

because ghana return different error code, we need to identify this error terminal task or not 

```docs

```


server/odc-core/src/main/resources/i18n/ErrorMessages.properties
```diff
@@ -134,6 +134,11 @@ com.oceanbase.odc.ErrorCodes.ExportExcelFileFailed=Export Excel failed, we recom
 com.oceanbase.odc.ErrorCodes.OscSqlTypeInconsistent=The input sql type {0} is inconsistent with expected type {1}, sql: {2}
 com.oceanbase.odc.ErrorCodes.NoUniqueKeyExists=There is no primary key or unique key exists in table {0}
 com.oceanbase.odc.ErrorCodes.OscNotEnabled=Online schema change is not enabled on current instance {0}, please contact administrator
+com.oceanbase.odc.ErrorCodes.OscDataCheckInconsistent=Online schema change data check is inconsistent
+com.oceanbase.odc.ErrorCodes.OmsGhanaOperateFailed=Oms ghana operate error
```
smallsheeeep, line 138: Could people understand what is "Oms"? What about "OceanBase Migration Service (OMS)"?

krihy: good idea

krihy: done

smallsheeeep (approved):
LGTM

PR #61, opened 2023-08-28
refine(workflow): add new workflow for building artifact and refine config

smallsheeeep:
#### What type of PR is this?

type-feature

Refine workflow config,and add a new workflow dedicated to build artifacts:
1. rename build_test to build_daily;
2. add a new workflow dedicated to build artifacts.

```docs

```

### DIFF
--- a/.github/workflows/build_artifact.yaml
+++ b/.github/workflows/build_artifact.yaml
@@ -1,24 +1,39 @@
 ###
-# This workflow is used to build all artifacts (jar, rpm, image and client) for testing
+# This workflow is used to quickly build artifacts (jar, rpm, image and client)
 # Triggers:
-#   1. Daily trigger at 04:00 (Beijing, UTC+08:00)
-#   2. Manual trigger
+#   1. Manual trigger
 # Jobs:
-#   1. Check Code Format
-#   2. PMD Scan
-#   3. Calculation Version Number
-#   4. Build Web Artifact (only x86_64 for now)
-#   5. Build Client Artifact
+#   1. Calculation Version Number
+#   2. Build Web Artifact (only x86_64 for now)
+#   3. Build Client Artifact
 ###
 
-name: Build Test
-run-name: Build Test triggered by ${{ github.actor }} 🛠️
+name: Build Artifact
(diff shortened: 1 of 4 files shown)


.github/workflows/build_daily.yaml
```diff
@@ -3,6 +3,7 @@
 # Triggers:
 #   1. Daily trigger at 04:00 (Beijing, UTC+08:00)
 #   2. Manual trigger
+#   3. Push tag with prefix "release"
```
yhilmare, line 6: How about changing the name to build_release.yaml?

.github/workflows/build_daily.yaml
```diff
@@ -242,33 +246,11 @@ jobs:
             ${{ runner.os }}-pnpm-store-
       - name: Build artifact
         env:
-          # MACOS_CERTIFICATE: ${{ secrets.PROD_MACOS_CERTIFICATE }}
```
yhilmare, line 245: These comments may be needed for subsequent front-ends, and it may be better to add them back

.github/workflows/build_artifact.yaml
```diff
@@ -114,8 +95,9 @@ jobs:
         with:
           node-version: "16"
       - name: Build front static resources
+        if: ${{ github.event.inputs.build_jar == 'true' || github.event.inputs.build_rpm == 'true' || github.event.inputs.build_docker == 'true'}}
         run: |
-          echo "Current directory: "`pwd`
+          echo "Current directory: " `pwd`
```
yhilmare, line 100: why empty space?

smallsheeeep: typo, fixed!

.github/workflows/build_daily.yaml
smallsheeeep: This workflow has nothing to do with release，but only auto triggerred when a release tag is pushed. 
Based on the conclusions of the previous meeting，the name is build_daily.

.github/workflows/build_daily.yaml
smallsheeeep: okay!

yhilmare (approved):

PR #62, opened 2023-08-28
refactor(osc): schedule task improve stable 

krihy:
#### What type of PR is this?

module online schema  change type refactor 

improve osc stableness 

 1.add osc expired time when exceed oscTaskExpiredAfterSeconds
 2.release oms project before start next schedule task
 3.set current schedule task id quartz data map and reschedule job before start next

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/quartz/QuartzJobService.java
```diff
@@ -149,13 +147,20 @@ public Trigger getTrigger(TriggerKey key) throws SchedulerException {
...
+    public CreateQuartzJobReq getQuartzJobReqOfJob(JobKey key) throws SchedulerException {
+        JobDetail jobDetail = scheduler.getJobDetail(key);
+        CreateQuartzJobReq req = new CreateQuartzJobReq();
+        req.setJobDataMap(jobDetail.getJobDataMap());
+        req.setType(JobType.valueOf(key.getGroup()));
+        req.setScheduleId(Long.parseLong(key.getName()));
+        return req;
+    }
+
+    public void updateJob(CreateQuartzJobReq req) throws SchedulerException {
```
guowl3, line 159: maybe you can use ScheduleService#update method to update the job parameters.

krihy: @see updateTriggerDataMap#updateTriggerDataMap , in this method update trigger data map

Issue #63, opened 2023-08-28
[Feature]: data-transfer task should display export objects before completion


LuckyPickleZZ:
### Is your feature request related to a problem?

The objects of Import and Export tasks will only be displayed after the task is completed. This leads to the inability to see the object during approval and the inability to approve it.

### Describe the solution you'd like

During approval, user can see the selected import and export objects

PR #73, opened 2023-08-28
feat(data-transfer): support saving export objects 

LuckyPickleZZ:
#### What type of PR is this?
type-bug
module-datatransfer

For now export objects could only be shown after task completion. Therefore, the approver cannot see which objects will be exported during approval. Resulting in inability to approve.

In this PR, the above problem is fixed. Approver can see all export objects immediately afer task created.
Plus, this PR also supports export 'type' objects.

https://github.com/oceanbase/odc/issues/63

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowTaskInstanceService.java
```diff
@@ -214,7 +216,7 @@ public List<? extends FlowTaskResult> getResult(@NotNull Long id) throws IOExcep
             return getAsyncResult(taskEntity);
         } else if (taskEntity.getTaskType() == TaskType.MOCKDATA) {
             return getMockDataResult(taskEntity);
-        } else if (taskEntity.getTaskType() == TaskType.EXPORT || taskEntity.getTaskType() == TaskType.IMPORT) {
```
yhilmare, line 217: how about export task?

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/DataTransferRuntimeFlowableTask.java
```diff
@@ -86,7 +86,7 @@ protected Void start(Long taskId, TaskService taskService, DelegateExecution exe
             odcInternalFileService.getExternalImportFiles(taskEntity, submitter, config.getImportFileName());
         }
         context = dataTransferService.create(taskId + "", config);
-        taskService.start(taskId, DataTransferTaskResult.of(context));
```
yhilmare, line 89: why don't we initialize the result of it here, but add a hook at flowinstanceService

LuckyPickleZZ: Because when the program reaches the start method of this class, it indicates that the approval has been passed and it has entered the execution phase. But our goal is to show the results to users during approval.
So we can only add a hook to initialize the results when creating the task.

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowTaskInstanceService.java
LuckyPickleZZ: Export task was processed above in line#206

yhilmare (approved):
LGTM

PR #75, opened 2023-08-29
feat(workflow): add checkbox in workflow to decide weather to install db-browser and ob-sql-parser

yhilmare:
#### What type of PR is this?
 
type-feature, type-refactor

We publish db-browser and ob-sql-parser by maven central repo, and we cannot publish to the central warehouse during the R&D process.

So, here is the solution. We install them into local machine which we run our workflow, so that we can build a jar that include what we modify on db-browser and ob-sql-parser

Fixes #

```docs

```


smallsheeeep (approved):
LGTM

Issue #80, opened 2023-08-29
[Bug]: failed to mask sensitive columns for ResultSetExport task


LuckyPickleZZ:
### ODC version

4.2.0

### OB version

*

### What happened?

When a sensitive column was set, the data will be masked while querying. But it won't be effective for result-set-export task. Origin data will be exported !

### What did you expect to happen?

The data of sensitive column can be masked.

### How can we reproduce it (as minimally and precisely as possible)?

Exporting any sensitive column will reproduce this.

LuckyPickleZZ:
Fixed.

runzi389205:
PASS


![Image](https://github.com/oceanbase/odc/assets/141202865/70462c75-e5de-4ce3-a64b-9884ad794279)

PR #84, opened 2023-08-29
fix(result-set): fix NPE by setting column accessor during session initialization

LuckyPickleZZ:
#### What type of PR is this?
type-bug
module-resultset

Because a variable in the ConnectionSession has not been initialized, the task of exporting the result set fails to obtain sensitive columns. 
As a result, the data has not been desensitized. It is a serious bug.

Fixes  https://github.com/oceanbase/odc/issues/80

```docs

```


yhilmare (approved):
LGTM

PR #88, opened 2023-08-29
feat(schema-plugin): schema-plugin access service layer

### LINKED ISSUE #82: [Bug]: String data type do not show “length” when viewing the column information of a table
### ODC version

4.2.0

### OB version

all verison

### What happened?

String data type do not show “length” when viewing the column information of a table，both in ob oracle mode and ob mysql mode：
image

show length of a String data type like odc 4.1.2 verison：
image

create String data type column in both ob mysql mode and ob oracle mode

_No response_

### Cloud

_No response_

### LINKED ISSUE #86: [Feature]: Performance Guarantee for SQL Development of ODC in the Scenario of Millions of Database Objects

Original user voice:
Views are used to obtain structural information. In some scenarios (such as a large number of tables), querying views takes a long time, resulting in slow viewing of table structures.
It is hoped to optimize the way of obtaining the object structure (such as getDDL) to solve the problem of slow viewing of the table structure in some scenarios.
Need to consider: OB 4.2 will drop the constraint name, and the impact of this change on the implementation plan of this requirement needs to be considered;

Build the scene:
1. Million-level tables, including wide tables;
2. One million partitions;
3. The table object needs to have information such as indexes and constraints;

Improve product usability

### Additional context

1. Test thoroughly (object list, structure view, SQL execution, SQL editing, result set editing, etc...);
2. If the performance problem of the kernel view is a strong stuck point, investigate the bypass solution;

### LINKED ISSUE #87: [Feature]:  Plug-in transformation of database object viewing and editing

None yet

Viewing and creating new database objects on the left is handled by schema-plugin-api.

### Additional context

QA test note：
Test whether the creation and viewing of various database objects under different data sources are consistent with before.

PeachThinking:
#### What type of PR is this?
type-feature
module-Database object

1. Viewing and creating new database objects is handled by `schema-plugin-api`.
2. Time-consuming queries methods of `DBSchemaAccessor` are replaced by `GetDBTableByParser` in ob oracle mode: `listIndexes` 、`getTableDDL`、`listConstraints`、`getPartition`
3. OB MySQL use `OBMySQLGetDBTableByParser#getPartition` to replace `DBSchemaAccessor#getPartition` in order to remove the dependency on SYS
4. remove `JdbcOperationsUtil` in `connect-plugin-ob-mysql` to `odc-core` module in order to remove all plugin modules' dependence on `connect-plugin-ob-mysql`
5. getTable interface in `DBTableController` return `DBTable` Object like before odc version 420，set table size in `DBTableStats` through `schema-plugin` layer，this fix issue：#82

Fixes #87
Fixes #86 
Fixes #82

RD has self-tested various database object viewing and editing

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/db/DBFunctionService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/db/DBFunctionService.java
@@ -19,18 +19,18 @@ import java.util.List;
 import java.util.stream.Collectors;
 
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.jdbc.core.ConnectionCallback;
 import org.springframework.stereotype.Service;
 
 import com.oceanbase.odc.core.authority.util.SkipAuthorize;
 import com.oceanbase.odc.core.session.ConnectionSession;
+import com.oceanbase.odc.core.session.ConnectionSessionConstants;
+import com.oceanbase.odc.plugin.schema.api.FunctionExtensionPoint;
 import com.oceanbase.odc.service.common.model.ResourceSql;
-import com.oceanbase.odc.service.db.browser.DBSchemaAccessors;
+import com.oceanbase.odc.service.plugin.SchemaPluginUtil;
 import com.oceanbase.odc.service.session.ConnectConsoleService;
 import com.oceanbase.tools.dbbrowser.model.DBFunction;
-import com.oceanbase.tools.dbbrowser.schema.DBSchemaAccessor;
-import com.oceanbase.tools.dbbrowser.template.DBObjectTemplate;
-import com.oceanbase.tools.dbbrowser.template.mysql.MySQLFunctionTemplate;
-import com.oceanbase.tools.dbbrowser.template.oracle.OracleFunctionTemplate;
+import com.oceanbase.tools.dbbrowser.model.DBPLObjectIdentity;
 
 import lombok.NonNull;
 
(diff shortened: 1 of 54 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/db/DBFunctionService.java
```diff
@@ -41,33 +41,37 @@ public class DBFunctionService {
...
-        String ddl = template.generateCreateObjectTemplate(function);
-        return ResourceSql.ofSql(ddl);
+        return ResourceSql.ofSql(session.getSyncJdbcExecutor(
+                ConnectionSessionConstants.BACKEND_DS_KEY)
+                .execute((ConnectionCallback<String>) con -> getFunctionExtensionPoint(session)
+                        .generateCreateTemplate(function)));
     }
 
+    private FunctionExtensionPoint getFunctionExtensionPoint(@NonNull ConnectionSession session) {
+        return SchemaPluginUtil.getFunctionExtension(session.getDialectType());
```
yhilmare, line 75: how to handle npe? `OdcPluginManager` get a logical flaw:
```Java
// OdcPluginManager#getSingleton
private <T extends ExtensionPoint> T getSingleton(List<T> collection, V object) {
        if (collection.size() > 1) {
            String message = MessageFormat.format("Expect single extension for {}, but got "
                    + "{}, extension : {}", object, collection.size(),
                    collection.stream().map(t -> t.getClass().getSimpleName()).collect(Collectors.toList()));
            throw new IllegalStateException(message);
        }
       // if collection's size is equal to 0?
        return collection.get(0);
    }
```
maybe we can solve this npe by enhancing the robustness of the `OdcPluginManager#getSingleton`

server/plugins/schema-plugin-api/src/main/java/com/oceanbase/odc/plugin/schema/api/TableExtensionPoint.java
```diff
@@ -43,4 +43,6 @@ public interface TableExtensionPoint extends ExtensionPoint {
     String generateCreateDDL(Connection connection, DBTable table);
 
     String generateUpdateDDL(Connection connection, DBTable oldTable, DBTable newTable);
+
+    String generateUpdateDDLWithoutRenaming(Connection connection, DBTable oldTable, DBTable newTable);
```
yhilmare, line 47: why should we create this method?

yhilmare (commented):
does this pr pass all unit cases?

server/plugins/schema-plugin-api/src/main/java/com/oceanbase/odc/plugin/schema/api/TableExtensionPoint.java
PeachThinking: because `DBTableService#generateUpdateDDLWithoutRenaming` call this method，but it is only used in ShadowTableComparingTask at present.
However, considering that the schema-plugin is only for ODC to view database objects, it is indeed unnecessary to add this method here, so I would delete this method.

server/odc-service/src/main/java/com/oceanbase/odc/service/db/DBFunctionService.java
PeachThinking: 👌 add code：
```if (collection.size() > 1) {
            String message = MessageFormat.format("Expect single extension for {}, but got "
                    + "{}, extension : {}", object, collection.size(),
                    collection.stream().map(t -> t.getClass().getSimpleName()).collect(Collectors.toList()));
            throw new IllegalStateException(message);
        } else if (collection.isEmpty()) {
            throw new UnsupportedException(String.format("Feature extension point is not supported for %s", object));
        }

File: server/plugins/schema-plugin-api/src/main/java/com/oceanbase/odc/plugin/schema/api/TableExtensionPoint.java (hunk 2 above)
Comment (yhilmare) [reply server/plugins/schema-plugin-api/src/main/java/com/oceanbase/odc/plugin/schema/api/TableExtensionPoint.java]: we may move `DBTableService#generateUpdateDDLWithoutRenaming` to `ShadowTableComparingTask`, which is only used by `ShadowTableComparingTask`.

Comment (PeachThinking) [reply server/plugins/schema-plugin-api/src/main/java/com/oceanbase/odc/plugin/schema/api/TableExtensionPoint.java]: 👌，`DBTableService#listTables` and `DBTableService#isLowerCaseTableName` are also only used in `ShadowTableComparingTask`，but will not move them to `ShadowTableComparingTask` because these method can be commonly used in the future by other method.

**Comment by PeachThinking** [comment]
> does this pr pass all unit cases?

The unit tests related to this PR have all been run, and then several problems introduced by the previous PR that caused the unit test to fail were fixed. Deleted the unpassable unit test `encryptIfRequires_ValueContainsENC`, because the files involving sensitive information required by the unit test have been deleted.

File: server/plugins/schema-plugin-ob-oracle/src/main/java/com/oceanbase/odc/plugin/schema/oboracle/parser/OBOracleGetDBTableByParser.java [hunk 3]
```diff
@@ -328,8 +328,8 @@ private CreateIndex parseIndexDDL(String ddl) {
         return statement;
     }
 
-    private static String readFile(String strFile) throws IOException {
-        try (InputStream input = new FileInputStream(strFile)) {
+    private String readFile(String strFile) throws IOException {
+        try (InputStream input = OBOracleGetDBTableByParser.class.getClassLoader().getResourceAsStream(strFile)) {
```
krihy, line 332: OBOracleGetDBTableByParser.class.getClassLoader may be replaced by getClass().getLassLoader in instance method

PeachThinking: 👌

krihy (approved):
LGTM

PR #89, opened 2023-08-30
fix(unit-test): read properties from .env and system environment variables

krihy:
fix unit test properties
#### What type of PR is this?

type-bug

Some test information is sensitivity, but test case need it , so we can put them in .env file or system environments

fix properties  about test case

```docs

```


smallsheeeep (approved):
LGTM

PR #96, opened 2023-08-31
reafctor: move build time and build version into InfoAdaptor interface

MarkPotato777:
#### What type of PR is this?
type-refactor

This PR moves getBuildTime and getBuildVersion into InfoAdaptor interface. The reason for this change is making any other starters can define their own version and pass it to front-end


yhilmare (approved):
LGTM

PR #99, opened 2023-08-31
feat(encryption): support asymmetric encryption

### LINKED ISSUE #56: [Feature]: Use asymmetric encryption instead of symmetric encryption

The current symmetric encryption transmits the encryption key in plain text, which poses a security risk.

Use asymmetric encryption instead of symmetric encryption.

### Additional context

_No response_

smallsheeeep:
#### What type of PR is this?
type-feature

Use asymmetric encryption instead of symmetric encryption.

Close #56 

The PR is WIP and merge will be enabled after self-testing.

### DIFF
--- a/server/odc-common/src/main/java/com/oceanbase/odc/common/crypto/RsaBytesEncryptor.java
+++ b/server/odc-common/src/main/java/com/oceanbase/odc/common/crypto/RsaBytesEncryptor.java
@@ -56,17 +56,17 @@ public class RsaBytesEncryptor implements BytesEncryptor {
     public RsaBytesEncryptor(@NonNull RsaEncryptorType type, String publicKeyBase64, String privateKeyBase64) {
         this.type = type;
         if (type == RsaEncryptorType.ENCRYPT_MODE) {
-            Validate.notNull(publicKeyBase64);
+            Validate.notBlank(publicKeyBase64);
             this.encryptor = createCipher(Cipher.ENCRYPT_MODE, transformToPublicKey(publicKeyBase64));
             this.decryptor = null;
         } else if (type == RsaEncryptorType.DECRYPT_MODE) {
-            Validate.notNull(privateKeyBase64);
+            Validate.notBlank(privateKeyBase64);
             this.encryptor = null;
             this.decryptor = createCipher(Cipher.DECRYPT_MODE, transformToPrivateKey(privateKeyBase64));
         } else {
-            Validate.notNull(publicKeyBase64);
+            Validate.notBlank(publicKeyBase64);
             this.encryptor = createCipher(Cipher.ENCRYPT_MODE, transformToPublicKey(publicKeyBase64));
-            Validate.notNull(privateKeyBase64);
+            Validate.notBlank(privateKeyBase64);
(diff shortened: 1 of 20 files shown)


server/odc-common/src/main/java/com/oceanbase/odc/common/crypto/RsaBytesEncryptor.java
```diff
@@ -0,0 +1,175 @@
...
+public class RsaBytesEncryptor implements BytesEncryptor {
+
+    private static final String ALGORITHM_NAME = "RSA";
+    private static final int DEFAULT_KEY_SIZE = 1024;
+
+    private final RsaEncryptorType type;
+    private final Cipher encryptor;
+    private final Cipher decryptor;
+
+    public RsaBytesEncryptor(@NonNull RsaEncryptorType type, String publicKeyBase64, String privateKeyBase64) {
```
krihy, line 57: A RsaBytesEncryptor instance should deal encryp and decrypt at same time, this construct looks like it can separated by mode

server/odc-common/src/main/java/com/oceanbase/odc/common/crypto/RsaBytesEncryptor.java
```diff
@@ -0,0 +1,175 @@
...
+        this.type = type;
+        if (type == RsaEncryptorType.ENCRYPT_MODE) {
+            Validate.notBlank(publicKeyBase64);
```
yizhouxw, line 60: Validate.notBlank should contains readable message

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionHelper.java
```diff
@@ -38,12 +35,9 @@
 @Validated
 public class ConnectionHelper {
 
-    @Autowired
-    private SensitivePropertyHandler sensitiveHandler;
-
     public ConnectionStringParseResult parseConnectionStr(@NotNull @Valid ParseConnectionStringReq req) {
         ConnectionStringParseResult result = MySQLClientArgsParser.parse2(req.getConnStr());
-        result.setPassword(sensitiveHandler.encrypt(result.getPassword()));
+        result.setPassword(null);
```
yizhouxw, line 40: can we just remove password field , or mark the password filed as WRITEONLY ?

smallsheeeep: Yes we can. btw, this api is deprecated and will be deleted after this merge.

server/odc-common/src/main/java/com/oceanbase/odc/common/crypto/RsaBytesEncryptor.java
smallsheeeep: I got you!

server/odc-common/src/main/java/com/oceanbase/odc/common/crypto/RsaBytesEncryptor.java
smallsheeeep: Yeah, asymmetric encryption algorithm can separated by mode, because the encryptor and decryptor use different key and can work individually.

PR #113, opened 2023-09-01
fix(sql-check): fix can not detect table&column comment when execute a sqlscript

### LINKED ISSUE #108: [Bug]: sql check gives an error result on table&column comment when execute a sql
### ODC version

4.2.0

### OB version

not related

### What happened?

I want to execute a sql as follow:
```SQL
create table date_test_tbl(
    id varchar2(64) primary key,
    col date
);
comment on table date_test_tbl is 'asdasdasd';
```
The database change panel is popped up:
截屏2023-09-01 10 51 57

I do write the table comment on the table, and there is the same problem on column comment check. What I expect is that I can execute these sqls directly.

follow "What happened"

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

this pr fix a bug, and you can get more details in related.

Fixes #108

```docs

```


smallsheeeep (approved):
LGTM

PR #117, opened 2023-09-01
fix(osc): fix oracle rename table failed

krihy:
#### What type of PR is this?
type-bug
module-Online schema change

fix oracle rename table failed 

1. fix oracle rename table sql don't  contain database name
2. fix oracle white account add connection config user with upper case
3. refactor build new table name and rename table name , create oracle new table name  prefix is "" and mysql new table name prefix is "_"
4. delete new table table after task failed 

```docs

```


PeachThinking (approved):
LGTM

Issue #120, opened 2023-09-01
[Bug]: can not add any new sql check rules


yhilmare:
### ODC version

4.2.0

### OB version

not related

### What happened?

I add a new sql check rule at `regulation-rule-metadata.yaml` and `regulation-rule-applying.yaml` , and I start up odc. Exception will be thrown:
<img width="787" alt="截屏2023-09-01 18 13 05" src="https://github.com/oceanbase/odc/assets/22992296/37c954dd-b591-4567-aa4a-744bbe4104e4">

### What did you expect to happen?

I can see the sql check rule which I just created

### How can we reproduce it (as minimally and precisely as possible)?

QAer will not care about this issue

PR #121, opened 2023-09-01
[WIP] fix(sql-check): fix sql-check will detect primary key's name

### LINKED ISSUE #102: [Bug]: remove name check for primary key def
### ODC version

4.2.0

### OB version

not related

### What happened?

we create a table as follow:
```SQL
create table test_tbl(
    id varchar(64) primary key
);
```
SQL-Check will give a violation message: "Index definition is not named"
截屏2023-08-31 17 42 16

primary key's def is different from general index def, and we can create a new sql-check rule to check if user's create table def give a name to a primary key.

follow "what happened"

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

Index name exists check rule will be work at primary key. you may hit this rule with such a sql executing:
```SQL
create table abcd(
    id int primary key
)
```
This is a very common way of writing, it is not very reasonable to be detected, so that I separate index naming detection from primary key naming detection.

Fixes #102 

```docs

```


yhilmare:
waiting for this issue #120 fixed

PR #126, opened 2023-09-04
fix(syncDatabase): fix filter out the internal schema when synchronizing the database

### LINKED ISSUE #95: [Bug]: filter out the internal schema when synchronizing the database
### ODC version

ODC 4.2.0

### OB version

OB 3.2.3.1

### What happened?

The internal schema is shown when the database is synchronized
image

Filter out LBACSYS、ORAAUDITOR

Create the data source for the sys tenant and synchronize the database

_No response_

### Cloud

_No response_

PeachThinking:
#### What type of PR is this?
type-bug

filter out the internal schema when synchronizing the database:`PUBLIC、LBACSYS、ORAAUDITOR、__public`

Fixes #95

```docs

```


MarkPotato777 (approved):
LGTM

PR #128, opened 2023-09-04
fix(osc): fix wrap invalid ddl error message  when sql parsed failed

krihy:
#### What type of PR is this?

module-online schema change
type-bug
current exception is unexpected exception message, this is cannot readable
  
when input sql in valid throws BadArgumentException replace to SyntaxErrorException

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/validator/OnlineSchemaChangeValidator.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/validator/OnlineSchemaChangeValidator.java
@@ -19,6 +19,7 @@ import static com.oceanbase.tools.dbbrowser.model.DBIndexType.UNIQUE;
 
 import java.io.StringReader;
 import java.util.List;
+import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/validator/OnlineSchemaChangeValidator.java
```diff
@@ -67,25 +70,24 @@ public class OnlineSchemaChangeValidator {
...
-                            parameter.getDelimiter());
-            for (String sql : sqls) {
-                Statement statement =
-                        (connectionConfig.getDialectType().isMysql() ? new OBMySQLParser() : new OBOracleSQLParser())
-                                .parse(new StringReader(sql));
-
-                validateType(sql, getSqlType(statement), parameter.getSqlType());
+        List<Statement> statements = parseStatements(parameter, connectionConfig, sqls);
 
+        ConnectionSession session = new DefaultConnectSessionFactory(connectionConfig).generateSession();
```
PeachThinking, line 88: the created session should be `session.expire();` in `try {} finally {}`

PR #129, opened 2023-09-04
WIP: feat(schema-plugin): add check table exist in table extension point

krihy:
#### What type of PR is this?

type-feature, module-schema-plugin

add check table exist in table extension point

when check table exist in schema by table name,  the input name may be lower case,  when ob oracle  is not case sensitive, the query result should be exist. but for ob mysql  case sensitive take not effect, input parameters show transfer to lower case

```docs

```

### DIFF
--- a/server/plugins/schema-plugin-api/src/main/java/com/oceanbase/odc/plugin/schema/api/TableExtensionPoint.java
+++ b/server/plugins/schema-plugin-api/src/main/java/com/oceanbase/odc/plugin/schema/api/TableExtensionPoint.java
@@ -36,6 +36,8 @@ public interface TableExtensionPoint extends ExtensionPoint {
 
     List<String> showNamesLike(Connection connection, String schemaName, String tableNameLike);
 
+    boolean checkTableExist(Connection connection, String schemaName, String tableName);
+
     DBTable getDetail(Connection connection, String schemaName, String tableName);
 
     void drop(Connection connection, String schemaName, String tableName);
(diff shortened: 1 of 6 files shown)


yhilmare:
why should we need it? does db-browser can do this?

PeachThinking:
1. The `schema-plugin` does not access the service layer in the odc 4.2.1 version. Your PR is merged into 4.2.1. Will this affect it?
2. The `schema-plugin` is only designed for viewing and editing of odc database objects. You may add these code to call for the task of `online-schema-change task`. It feels like it is inappropriate to put it in the `schema-plugin layer`.

krihy:
> why should we need it? does db-browser can do this? if we need it, how can we handle procedure, view or other db object?

Current moment ob versions is only sensitive on  table name case in oracle model, name of other objects type  is not supported which is stored by default case,  oracle model is upper case and mysql model is lower case, if `checkTableExists` function is only used in osc module, we can put it in osc service

krihy:
> 1. The `schema-plugin` does not access the service layer in the odc 4.2.1 version. Your PR is merged into 4.2.1. Will this affect it?
> 2. The `schema-plugin` is only designed for viewing and editing of odc database objects. You may add these code to call for the task of `online-schema-change task`. It feels like it is inappropriate to put it in the `schema-plugin layer`.

The same  comment as above  https://github.com/oceanbase/odc/pull/129#issuecomment-1706478173

PR #130, opened 2023-09-04
fix(osc): fix sql of alter replace table name not correct

krihy:
#### What type of PR is this?

type-debug
type-online schema change
fix alter type execute failed 

1. oracle new table name with suffix with  "OSC_NEW_" and renamed table suffix with "OSC_OLD_"
2. fix alter type sql replaced by new table name
3. fix oracle constraints name with uuid, old renamed style may overrun 128 characters

```docs

```


PeachThinking (approved):
LGTM

PR #132, opened 2023-09-04
refactor(ob-sql-parser): add several new syntaxes which added in OceanBase 4.1.0

### LINKED ISSUE #123: [Bug]: GET DIAGNOSTICS Statement does not supported for mysql mode
### ODC version

4.2.0

### OB version

4.1.0

### What happened?

does not support syntax `GET DIAGNOSTICS`.

while use  ` GET DIAGNOSTICS CONDITION 1 res_desc= MESSAGE_TEXT;`

result into syntax error, SQL Check result:
You have an error in your SQL syntax; check the manual for the right syntax to use near 'N\n\n GET DIAGNOSTICS...' at line 67, col 12

 and the parameters will be empty.
image

no  syntax error, the parameters can shown in detail page of procedure

smply use below statement.

```
GET DIAGNOSTICS @p1 = NUMBER, @p2 = ROW_COUNT;
```

refer more from https://dev.mysql.com/doc/refman/5.7/en/get-diagnostics.html 

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-refactor

OceanBase 4.1.0 added some new syntaxes, such as `get_diagnostics_stmt`. Current ob-sql-parser isn't adapt to new syntax, and cause some syntax error, so that we should add these new syntaxes.

Fixes #123 
Closes #123 

```docs

```

### DIFF
--- a/libs/ob-sql-parser/src/main/resources/obmysql/sql/OBLexer.g4
+++ b/libs/ob-sql-parser/src/main/resources/obmysql/sql/OBLexer.g4
@@ -2650,8 +2650,8 @@ SIMPLE
     : S I M P L E
     ;
 
-BEGI
-    : B E G I
+BEGIN
+    : B E G I N
     ;
 
 VARIABLES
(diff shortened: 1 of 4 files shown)


smallsheeeep (approved):
LGTM

PR #133, opened 2023-09-04
fix(db-session): fix can not get latest query sql when list all sessions

### LINKED ISSUE #104: [Bug]: session management do not show the SQL being executed
### ODC version

4.2.0

### OB version

4.1.0 MySQL mode

### What happened?

1、To execute a SQL statement in a window，for example :select sleep(600);
2、To view the corresponding session in session management；
image

Session Management  show the SQL being execute

1、To execute a SQL statement in a window，for example :select sleep(600);
2、To view the corresponding session in session management；

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

we forget setting `OdcDBSession#sql` when convert from `DBSession` to `OdcDBSession` which cause a bug that user can not see the sql executing when list all db sessions

Fixes #104 
Close #104 

```docs

```


smallsheeeep (approved):
LGTM， but the `odcdbsession` may not match the param name specification

PR #134, opened 2023-09-04
fix(flow): flow's status is illegal when failed to submit a task

### LINKED ISSUE #79: [Bug]: flow's status is not right when service task failed to submit
### ODC version

4.1.1

### OB version

unrelated

### What happened?

Flow's status is not correct, which is 'CREATED'

The concurrent number of task is limited. Task will be denied when user submit too many tasks, and the default strategy is dropping task.

Which we expect is that we should convert flow's status from 'CREATED' to 'FAILED' 

QAer have not to care about this issue which is hard to reproduce~

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

this is a bug that flow's status will be illegal when task is rejected by thread pool.

Flow framework will eat the exception thrown by thread pool and do nothing, and 'CREATED' will be the final status of the flow instance which always confuse user.

Fixes #79 
Close #79 

This pr does things as follow:
1. increase the size of the thread pool.
2. set the capacity for queue of the thread pool to throw `RejectionException` when too many tasks are submitted.
3. add log print and status changing when too many tasks is submitted(`RejectionException` is thrown).

`OdcAsyncJobExecutor.java` gives such task submission logic:
1. get `processDefinitionId` by job (which is given by flowable`org.flowable.job.service.impl.persistence.entity.AbstractJobEntityImpl#getProcessDefinitionId`), and get `flowInstanceId` by it.
2. find out what task type does the flow contains(search by `flowInstanceId`).
3. mock data, import/export task and the others task will be submitted to different thread pool(`mockDataExecutorService`, `loaderDumperExecutorService` and `executorService` respectively).
4. flow's status will be updated to `FAILED` when `RejectionException` is thrown to avoid the task status remaining in 'CREATED'.

The cause of the problem may be that the user submitted too many tasks so that the thread pool rejected the submission. However, ODC does not change the status of the flow when rejection happened. 

Our solution is that update the status of the flow from 'CREATED' to 'FAILED' when rejection happened, this method can explicitly prompt the user that the task is rejected.

```docs

```


yizhouxw (approved):
LGTM

PR #135, opened 2023-09-04
fix(osc): fix duplicate foreign key constraint name when execute new table create ddl

krihy:
#### What type of PR is this?

type-bug
type-Online schema change
occur duplicate key error when create a new table contain  foreign key with the same constraint name  exists in database

replace the old constraint name by TokenStreamRewriter when antlr invoke enterConstraint_name

```docs

```


yhilmare (approved):
lgtm

PR #137, opened 2023-09-04
feat(iam): add a built-in role 'individual_space'

MarkPotato777:
#### What type of PR is this?
type-feature
module-iam

Add a built-in role 'individual_space' which contains individual space management permissions, making user easier to grant permissions to users who need to use individual space.

Other changes: move some migration resource Yamls to the 'web' package which will be migrated during bootstrap other than login. This change will not have effects on users who upgrade from V4.2.0 because allow_duplicate is false in these Yamls.


yhilmare (approved):
lgtm

PR #148, opened 2023-09-05
security: fix unmatched items in .secignore

MarkPotato777:
#### What type of PR is this?
type-security

fix some URLs in .secignore to pass the data security scan


smallsheeeep (approved):
LGTM

PR #168, opened 2023-09-05
fix(pl-debug): add exception prompt when debugging errors

### LINKED ISSUE #162: [Bug]: failed to debug and execute an anonymous block ending with '/' at anonymous block window
### ODC version

4.2.0

### OB version

OceanBase 4.1.0.1

### What happened?

debug an anonymous block:
```SQL
declare 
  var_ename varchar2(50);
  var_job varchar2(50);
  cursor cur_emp 
      is select ename,job from employees;
  begin 
      open cur_emp;
      fetch cur_emp into var_ename,var_job; 
      if cur_emp%found then 
          dbms_output.PUT_LINE('游标现在的行号: '||cur_emp%found);
          dbms_output.PUT_LINE('编号是7499的雇员名称为: '||var_ename||'，职务是: '||var_job);
          fetch cur_emp into var_ename,var_job;
      else 
          dbms_output.PUT_LINE('无数据记录: '||cur_emp%notfound);
      end if;
      
      if cur_emp%ISOPEN THEN 
          dbms_output.PUT_LINE('游标是否开启：'||cur_emp%ISOPEN);
          CLOSE cur_emp; -- 关闭游标
          dbms_output.PUT_LINE('游标关闭成功！');
      end if;
  end;
/
```
and some errors happened:
image1
image2

Debugging can proceed normally

same as "What happened"

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug
module-PL debugger

this pr throw an exception when failed to starting debugging which gives a clear error information.

Fixes #162 
Close #162 

Executing and debugging an anonymous block inside an anonymous block fails because there is a '/' at the end of the anonymous block, which causes a syntax error. 

This is not a problem in itself, because the anonymous block window is designed not to split sentences, causing the '/' at the end of the anonymous block to be sent to the server for execution. 

The problem is that when the debugging failed, the user was not informed of the specific cause of the failure. This PR will inform the user of the specific cause of the error

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/pldebug/session/DebuggerSession.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/pldebug/session/DebuggerSession.java
@@ -49,6 +49,7 @@ import com.oceanbase.odc.service.pldebug.model.DBPLError;
 import com.oceanbase.odc.service.pldebug.model.PLDebugBreakpoint;
 import com.oceanbase.odc.service.pldebug.model.PLDebugErrorCode;
 import com.oceanbase.odc.service.pldebug.model.PLDebugPrintBacktrace;
+import com.oceanbase.odc.service.pldebug.model.PLDebugResult;
 import com.oceanbase.odc.service.pldebug.model.PLDebugStatusReason;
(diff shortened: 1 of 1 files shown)


yizhouxw (approved):
LGTM

PR #173, opened 2023-09-06
fix(osc): fix old running task throw npe when enable full verify  

krihy:
#### What type of PR is this?

type-bug
module-online schema change

log password and check full verify result cause npe

1. fix check full verify result may cause npe when verify result is not arrived
2. fix clean up password in log when request oms create datasource failed

```docs

```


smallsheeeep (approved):
LGTM

PR #175, opened 2023-09-06
fix(data-security): create sensitive columns with case insensitive same column and table names

### LINKED ISSUE #161: [Bug]: "*.*.*"The path recognition rules are abnormal when scanning desensitized columns
### ODC version

4.2.1

### OB version

any schema

### What happened?

when the path recognition rule is: `*.*.*`, scan a database
 report errror：{
    "code": "DataAccessError",
    "durationMillis": 524,
    "error": {
        "code": "DataAccessError",
        "details": [],
        "message": "访问元数据库错误，请联系管理员"
    },
    "httpStatus": "INTERNAL_SERVER_ERROR",
    "message": "访问元数据库错误，请联系管理员",
    "requestId": "IG04KXONDBNU83NKK8KW3",
    "server": "sqaappecsv62s2011162217057.sa128",
    "successful": false,
    "timestamp": 1693907117.016000000,
    "traceId": "286262410fb74e6f"
}

no error

-

-

### Cloud

-

smallsheeeep:
#### What type of PR is this?
type-bug
module-Data security

The ODC metaDB is case insensitive when calculating unique key. So, if there are two columns with the case insensitive same table name and column name, and add both of them to sensitive columns, exception will be throw.

Solution: Make sensitive columns case insensitive. if there are two columns with the case insensitive same table name and column name, then they cannot be both add to sensitive columns.

Fixes #161


yizhouxw (commented):
unit testing missed

smallsheeeep:
> unit testing missed

Done!

PR #232, opened 2023-09-07
fix(connection): check database type when test connection.

guowl3:
#### What type of PR is this?
type-bug

Avoid database synchronization failure caused by user inputting the wrong database type.

Check database type when test connection.

```docs

```


server/plugins/connect-plugin-api/src/main/java/com/oceanbase/odc/plugin/connect/api/InformationExtensionPoint.java
```diff
@@ -27,4 +29,8 @@
 public interface InformationExtensionPoint extends ExtensionPoint {
 
     String getDBVersion(Connection connection);
+
+    String getDBVersionComment(Connection connection);
```
krihy, line 33: Oracle and other db may not provide version comment, this method can move to OBMySQLInformationExtensionPoint for present

krihy (approved):
LGTM

PR #243, opened 2023-09-07
refactor(migrate): speed up resource migration and add transaction control

yhilmare:
#### What type of PR is this?

type-refactor

1. The efficiency of the resource migration framework is low, and the migration takes a lot of time. This pr improves the migration efficiency.
Before refactor:
```shell
obclient> select type, sum(execution_millis) from migrate_schema_history group by type;
+----------+-----------------------+
| type     | sum(execution_millis) |
+----------+-----------------------+
| SQL      |                 36270 |
| JDBC     |                 77963 |
| RESOURCE |                 52505 |
+----------+-----------------------+
3 rows in set (0.00 sec)
```
After refactor:
```SQL
obclient> select type, sum(execution_millis) from migrate_schema_history group by type;
+----------+-----------------------+
| type     | sum(execution_millis) |
+----------+-----------------------+
| SQL      |                 39402 |
| JDBC     |                 82199 |
| RESOURCE |                 19997 |
+----------+-----------------------+
3 rows in set (0.00 sec)
```
as you can see: 60% reduction in time consumption

2. Added transaction control for resource migration to prevent data from being in an inconsistent state due to exceptions during the migration process.

Fixes #

1. speed up resource migration, the resource migration framework previously needed to execute 3 SQL statements for each record migration:
    1. query by unique keys to check if this record is present.
    2. insert the record.
    3. query it by unique keys to full fill the generated id. 

    Step 1 and step 3 is not necessary.

    If a record allows duplicates, we don't have to do step 1, and we can insert the record regardless of whether they already exist.

    We can get generated keys by jdbc directly rather than query it.

2. Added transaction control in `ResourceMigrator`

```docs

```


yizhouxw (approved):
👍👍 
nice improvment!

PR #275, opened 2023-09-08
refactor(migrates): add some abstract methods for migrates

yhilmare:
#### What type of PR is this?

type-refactor

as the increment of the team organization and individual organization, the time cost of resource migration is more and more unacceptable, so that we should migrate all resources when user login or organization created.

We refactor the migrates framework to archive the goal mentioned, so that we can reuse the ability of migrate framework when migrate organization or user's data.

Fixes #

```docs

```

### DIFF
--- a/server/odc-core/src/main/java/com/oceanbase/odc/core/migrate/BootstrapSchemaHistoryRepository.java
+++ b/server/odc-core/src/main/java/com/oceanbase/odc/core/migrate/BootstrapSchemaHistoryRepository.java
@@ -1,0 +1,156 @@
+/*
+ * Copyright (c) 2023 OceanBase.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


server/odc-core/src/main/java/com/oceanbase/odc/core/migrate/BootstrapSchemaHistoryRepository.java
```diff
@@ -0,0 +1,156 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * schema history persistent, the table schema define in migrate_schema_history_table_template.sql
+ * 
+ * @author yizhou.xw
+ * @version : SchemaHistoryRepository.java, v 0.1 2021-03-26 14:26
+ */
+@Slf4j
+public class BootstrapSchemaHistoryRepository implements SchemaHistoryRepository {
```
yizhouxw, line 51: how about JdbcSchemaHistoryRepository ?

yhilmare: which is based on scenario, `BootstrapSchemaHistoryRepository` is used on `migrate_schema_history`, the other impl of `SchemaHistoryRepository` can also be based on jdbc but different table.

MarkPotato777: LGTM

MarkPotato777 (dismissed):
LGTM

MarkPotato777 (approved):

PR #285, opened 2023-09-08
feat(table-index):  show index is available or unavailable

### LINKED ISSUE #235: [Feature]:  Check the index to show whether the index is available

no

image
For the case where the index is not available, can the ODC not display or display the index unavailable status, otherwise the user will think that the index has been created successfully.
image

### Additional context

_No response_

PeachThinking:
#### What type of PR is this?
type-feature
type-Database Object

Displaying whether the index is available (whether the index is created successfully) will not be used as a new column to view index information (because it is available in most cases, and it cannot be edited), but if the index is unavailable a prompt will be given to the user (similar to the prompt when viewing the PL invalid status in the object tree on the left).

Displays whether a index is available when viewing the index. If the index has been successfully created, it is in an available state, and if it has not been created successfully, it is in an unavailable state. `DBTableIndex.available` represents whether the index is available.

- Oracle judges whether it is available through the `STATUS` field of `SYS.ALL_INDEXES`.

- mysql judges whether it is available through the `COMMENT` field of `information_schema.statistics`.
mysql official documentation：
image
ob mysql official documentation：
image

Fixes #235

UED：Only use hover prompts, do not add the "status" column
image

```docs

```


libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoGreaterThan5740SchemaAccessor.java
```diff
@@ -953,6 +955,14 @@ public List<DBTableIndex> listTableIndexes(String schemaName, String tableName)
         return new ArrayList<>(indexName2Index.values());
     }
 
+    protected void handleIndexAvailability(DBTableIndex index, String availability) {
```
yhilmare, line 958: how about 'enable' status?

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/OBMySQLSchemaAccessor.java
```diff
@@ -179,6 +179,15 @@ public List<DBTableIndex> listTableIndexes(String schemaName, String tableName)
         return indexList;
     }
 
+    @Override
+    protected void handleIndexAvailability(DBTableIndex index, String availability) {
```
yhilmare, line 183: duplicated code

PeachThinking: `OBMySQLSchemaAccessor#handleIndexAvailability`  override `MySQLNoGreaterThan5740SchemaAccessor#handleIndexAvailability`,  the implementation of these two methods is different.

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoGreaterThan5740SchemaAccessor.java
PeachThinking: 👌

yhilmare (approved):
lgtm

PR #287, opened 2023-09-08
fix(pldebug): fix debug obtain connection info from direct connection config

krihy:
#### What type of PR is this?

type-bug
module-pldebug

fix debug obtain connection info from direct connection config

acquire host and port info should from connection config when config without cluster

```docs

```


smallsheeeep (approved):
LGTM

PR #290, opened 2023-09-09
refactor(migration): extract data migration interface

MarkPotato777:
What type of PR is this?
type-refactor
module-migration

Extract a migration interface and make the old implementation the default one, because other starters may have other implementations while migration.

Other changes: 1. Optimize log when sync data sources failed


yizhouxw (approved):
LGTM

PR #291, opened 2023-09-09
fix(migration): missing user organization relation migration

MarkPotato777:
#### What type of PR is this?

type-bug
module-migration

Decoupling user organization migration and organization resources migration

Use a single class UserOrganizationMigrator to record user and organization relations, and fix the timing of recording user organization relation


guowl3 (approved):
LGTM

PR #292, opened 2023-09-10
security: bad horizontal access control

MarkPotato777:
#### What type of PR is this?
type-security

Some method does not check horizontal permission, which may cause broken horizontal access control. Use HorizontalDataPermissionValidator to validate permission


guowl3 (approved):
LGTM

PR #303, opened 2023-09-11
fix(recyclebin): fix failed to generate flashback sql

### LINKED ISSUE #295: [Bug]: failed to get flashback sql on specific db object
### ODC version

4.2.0

### OB version

not related

### What happened?

I want to generate flash back sql on some db objects, and I got nothing as you can see as follow:
截屏2023-09-11 10 23 48

I can get the generated flashback sql.

try to generate a flashback sql

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

this pr fix a bug that odc can not generate flashback sql on some types, just like `normal index`.

Fixes #295 
Close #295 

OceanBase only supports some specific types for flashback, `TABLE`, `VIEW`, `TENANT`, `DATABASE` and `SCHEMA`.
ODC will give up generating sql when recycle object's type is not in available types, that is why the api returns nothing.

To fix this bug, we generate a comment to tell the user that your recycle object's type is invalid rather than ignore it

```docs

```

### DIFF
--- a/server/odc-core/src/main/java/com/oceanbase/odc/core/shared/constant/ErrorCodes.java
+++ b/server/odc-core/src/main/java/com/oceanbase/odc/core/shared/constant/ErrorCodes.java
@@ -223,6 +223,7 @@ public enum ErrorCodes implements ErrorCode {
 
     ObPreCheckDdlFailed,
     ObCopySchemaFailed,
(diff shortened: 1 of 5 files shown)


MarkPotato777 (approved):
LGTM

PR #327, opened 2023-09-13
fix: data masking failed with multiple join clause in SQL

### LINKED ISSUE #319: [Bug]: Data masking fails when using multiple JOIN clauses in SQL
### ODC version

4.2.1

### OB version

Any

### What happened?

When execute:
```sql
SELECT
  s.*
FROM
  student s
  INNER JOIN student_teacher st ON s.id = st.student_id
  INNER JOIN teacher t ON st.teacher_id = t.id;
```
Data masking is failed.

Some columns of result set should be masked.

Execute above SQL.

_No response_

### Cloud

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
module-Data masking

Fix bug during data masking

Fixes #319 

Unit test is covered


yhilmare (approved):
lgtm

Issue #313, opened 2023-09-12
[Feature]: External integration support extract XML response


smallsheeeep:
### Is your feature request related to a problem?

The response of external system may be XML, but ODC cannot support it yet.

### Describe the solution you'd like

1. Allow user choose different type of response, it may be a new config item;
2. Based on the response type config, using different ways to extract response.

PR #338, opened 2023-09-13
feat(integration): support retrieve xml format response

smallsheeeep:
#### What type of PR is this?
type-feature
module-integration

Support retrieve xml format response

Closing #313 

Unit test has covered

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/integration/model/IntegrationProperties.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/integration/model/IntegrationProperties.java
@@ -53,7 +53,6 @@ public class IntegrationProperties {
         private Body body;
         private boolean requestEncrypted = false;
         private boolean responseEncrypted = false;
-        private ResponseContentType responseContentType = ResponseContentType.JSON;
         @NotBlank
         private String requestSuccessExpression = "true";
     }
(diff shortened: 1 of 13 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/integration/model/IntegrationProperties.java
```diff
@@ -53,13 +53,14 @@ public static class ApiProperties {
         private Body body;
         private boolean requestEncrypted = false;
         private boolean responseEncrypted = false;
+        private ResponseType responseType = ResponseType.JSON;
```
yizhouxw, line 56: it is not 'response type', but format of the response body.

similar concept may 'contentType'.

however, we may got content type from response header 'Content-Type', so if the content type is not defined,
ODC may detect from response header.

smallsheeeep: I don't think we need to pay extra attention to the `Content-type` in response header. (Yes we can, but not necessary)

Apache http-core will parse response to `String` based on the Content-type automatically. So we should not care about different type of response content, just process the `String` according to the format specified by the integrator.

smallsheeeep: Rename `responseType` to `responseContentType`

yizhouxw (approved):
LGTM

PR #340, opened 2023-09-14
fix(connection):adapter result has been overwritten.

guowl3:
#### What type of PR is this?
type-bug

The password of the connection will be overwritten if you modify it in the adapter.

Fix the issue of the adapter result has been overwritten.

```docs

```


MarkPotato777 (approved):
LGTM

Issue #343, opened 2023-09-15
[Bug]: Desktop version startup error


runzi389205:
### ODC version

mac m1
odc: 4.2.1

### OB version

Not involved

### What happened?

![image](https://github.com/oceanbase/odc/assets/141202865/e3f10f4d-1cf1-4a6d-8eb9-e1175694071f)

![image](https://github.com/oceanbase/odc/assets/141202865/75c6df35-4137-4200-afa0-da8be1515782)
[odc.log](https://github.com/oceanbase/odc/files/12616605/odc.log)

### What did you expect to happen?

Enter desktop version

### How can we reproduce it (as minimally and precisely as possible)?

Overwrite the installation and start the desktop version

Huangxiao-mas:
test passed

yizhouxw:
fixed in v4.2.1

PR #345, opened 2023-09-15
fix(clientMode): odc fail to start on clientMode

LuckyPickleZZ:
#### What type of PR is this?
type-bug

ODC failed to start on clientMode for ` No qualifying bean of type 'com.oceanbase.odc.service.collaboration.OrganizationResourceMigrator' available:`
After remove the `@Profile` annotation, ODC can start normally.

Fixes https://github.com/oceanbase/odc/issues/343

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/collaboration/DefaultOrganizationResourceMigrator.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/collaboration/DefaultOrganizationResourceMigrator.java
@@ -22,6 +22,7 @@ import java.util.Optional;
 import javax.sql.DataSource;
 
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.annotation.Profile;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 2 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/collaboration/DefaultOrganizationResourceMigrator.java
```diff
@@ -44,7 +43,6 @@
  * @Date: 2023/5/30 10:53
  * @Description: []
  */
-@Profile("alipay")
```
MarkPotato777, line 47: add clientMode into @Profile rather than just remove @Profile as  some other profile has its own implementation

LuckyPickleZZ: done. added DesktopOrganizationResourceMigrator for clientMode only

yhilmare (approved):
LGTM

PR #347, opened 2023-09-15
fix(osc): fix personal individual space obtain main account id error

krihy:
#### What type of PR is this?

type-bug
module-Online schema change

fix personal individual area main uid error

fix get main account uid in personal individual area, store the uid in flow variable, and get it from executionContext  

```docs

```


smallsheeeep (approved):
LGTM

PR #349, opened 2023-09-15
refactor: optimize log

MarkPotato777:
#### What type of PR is this?
type-refactor

change log level from info to debug, preventing massive log


guowl3 (approved):
LGTM

PR #361, opened 2023-09-19
fix(sql-execute): fix failed to print dbms output

### LINKED ISSUE #348: [Bug]: dbms output error, dbms_output.chararr type does not exist
### ODC version

ODC421

### OB version

Oceanbase312

### What happened?

PL debugging, after the Oceanbase312 version of the function is run in the ODC412 version and the ODC421 version, the DBMS output is inconsistent.
image

image

Don't print DBMS output

this issue can be reproduced on OceanBase 3.2.4, which doest not occur on later OceanBase.

```SQL
declare
begin
    dbms_output.put_line('Hello,world');
end;
```

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

`dbms_output.put_line` will not print anything on OceanBase 3.x when using ODC, this pr fix this bug.

odc pring dbms log by using this procedure `dbms_output.get_lines` which is added on OceanBase 4.x and not patched to OceanBase 3.x. 

this pr use `dbms_output.get_line` to print dbms log, this method will work for both OceanBase 3.x and OceanBase 4.x~

Fixes #348 

```docs

```


smallsheeeep (approved):
LGTM

PR #364, opened 2023-09-19
fix(osc): fix input sql check unsupported foreign key and different column

krihy:
#### What type of PR is this?

type-bug
module-online schema change
1. table with foreign key precheck failed in oms
2.add primary column、 drop column or rename column will failed in oms precheck

1. add validate for table with foreign key  in odc
2. compare old and new table columns, osc task will failed if they are different 

```docs

```


PeachThinking (approved):
LGTM

PR #365, opened 2023-09-19
[Enhancement]: get create database/schema sql  by schema-plugin

### LINKED ISSUE #337: [Enhancement]: make database/schema created by plugin

Extend the plugin so that the actual business logic of "new database" in the data source management office is handled by the plugin:
截屏2023-09-13 19 48 47

### Why is this needed?

The current database creation logic is hard-coded in the service layer to determine the data source type through business code, which is not flexible enough. If you need to add data sources in the future, it will need to be expanded.
```Java
DatabaseService#getCreateDatabaseSql
private String getCreateDatabaseSql(@NonNull ConnectionConfig connectionConfig, String databaseName,
            String collationName, String charsetName) {
        SqlBuilder sqlBuilder = null;
        if (connectionConfig.getDialectType().isMysql()) {
            sqlBuilder = new MySQLSqlBuilder();
            sqlBuilder.append("create database ").identifier(databaseName);
            if (StringUtils.isNotEmpty(charsetName)) {
                sqlBuilder.append(" character set ").append(charsetName);
            }
            if (StringUtils.isNotEmpty(collationName)) {
                sqlBuilder.append(" collate ").append(collationName);
            }
        } else if (connectionConfig.getDialectType().isOracle()) {
            sqlBuilder = new OracleSqlBuilder();
            sqlBuilder.append("CREATE USER ").identifier(databaseName).append(" IDENTIFIED BY ")
                    .identifier(connectionConfig.getPassword());
        }
        return Objects.isNull(sqlBuilder) ? StringUtils.EMPTY : sqlBuilder.toString();
    }
```
we should do this by plugin

PeachThinking:
#### What type of PR is this?
type-enhancement
module-create database

make `DatabaseService` get create database/schema sql  by schema-plugin
Fixes #337

```docs

```


server/plugins/schema-plugin-api/src/main/java/com/oceanbase/odc/plugin/schema/api/DatabaseExtensionPoint.java
```diff
@@ -36,4 +37,6 @@ public interface DatabaseExtensionPoint extends ExtensionPoint {
     DBDatabase getDetail(Connection connection, String dbName);
 
     List<DBDatabase> listDetails(Connection connection);
+
+    String getCreateDatabaseSql(String databaseName, Map<String, String> parameters);
```
yhilmare, line 41: this interface may be defined as follow:

```Java
DBDatabase create(Connection connection, @NonNull DBDatabase);
```

create schema by db-browser, rather than returning a sql executed by caller

yhilmare: further more, `Map<String, String> parameters` may be confusing~~

PeachThinking: > further more, `Map<String, String> parameters` may be confusing~~

how about rename to "createDBParameters"?

yhilmare (approved):
lgtm

PR #369, opened 2023-09-20
fix(security): risky URLs discovered by security scans

smallsheeeep:
Fix URL discovered by security scans


yhilmare (approved):
LGTM

PR #371, opened 2023-09-20
fix(clientMode): fail to start for lack of Service annotations

LuckyPickleZZ:
#### What type of PR is this?
type-bug 

In PR #345  , we added a service named DesktopOrganizationResourceMigrator. But it missed `@Service` annotation, leading to start failure for desktop.
Now it would be completed.

```docs

```


yhilmare (approved):
lgtm

PR #377, opened 2023-09-20
fix(mvc): api response content type converts to  xml

smallsheeeep:
#### What type of PR is this?
type-bug

After the #338 commit was merged, I find the api response is converted into XML format, causing the ODC server cannot be used by front-end.
In this PR, I fix this issue by set `application/json` as the default content type by implements the `WebMvcConfigurer`:
```java
    @Override
    public void configureContentNegotiation(ContentNegotiationConfigurer config) {
        config.defaultContentType(MediaType.APPLICATION_JSON, MediaType.ALL);
    }
```

### DIFF
--- a/server/odc-server/src/main/java/com/oceanbase/odc/server/config/MvcConfiguration.java
+++ b/server/odc-server/src/main/java/com/oceanbase/odc/server/config/MvcConfiguration.java
@@ -30,6 +30,7 @@ import org.springframework.context.annotation.Configuration;
 import org.springframework.core.Ordered;
 import org.springframework.format.FormatterRegistry;
 import org.springframework.format.datetime.standard.DateTimeFormatterRegistrar;
+import org.springframework.http.MediaType;
 import org.springframework.http.converter.HttpMessageConverter;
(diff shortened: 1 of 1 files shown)


server/odc-server/src/main/java/com/oceanbase/odc/server/config/MvcConfiguration.java
```diff
@@ -95,7 +96,7 @@ public void addCorsMappings(CorsRegistry registry) {
 
     @Override
     public void configureContentNegotiation(ContentNegotiationConfigurer config) {
-        config.favorPathExtension(false);
```
yizhouxw, line 98: why remove `config.favorPathExtension(false);` ?

smallsheeeep: Because the `favorPathExtension` config is default set to `false`, so i remove it.
By the way, it has been marked `Deprecated`.

yizhouxw (approved):
lgtm

Issue #382, opened 2023-09-21
[Bug]: parsing sensitive column even through there is no sensitive column


smallsheeeep:
### ODC version

4.2.0

### OB version

Any

### What happened?

ODC will parse sensitive column even through there is no sensitive column

### What did you expect to happen?

If there is no sensitive column, ODC should skip parsing sensitive column

### How can we reproduce it (as minimally and precisely as possible)?

Executing SQL in console

sl01388797:
pass

PR #383, opened 2023-09-21
fix: masking enabled

smallsheeeep:
#### What type of PR is this?
type-bug
module-data masking
Closing #382 

ODC will parse sensitive column even through there is no sensitive column.
This PR fix it by adding a judgement:  If there is no sensitive column, or the data-masking is disabled, then skip data masking processing.


yhilmare (approved):
lgtm

PR #390, opened 2023-09-22
fix(osc): osc support ob ce add type  ob mysql ce

krihy:
#### What type of PR is this?

module-Online-shema-change
type-bug

add oms db type OB_MYSQL_CE

create oms datasource by type OB_MYSQL_CE when target db version is ob ce

```docs

```


yizhouxw (approved):
LGTM

PR #395, opened 2023-09-25
fix(sql-execute): fix table with rowid datatype cannot be edited

### LINKED ISSUE #370: [Bug]: Failed to add data
### ODC version

odc421

### OB version

not have

### What happened?

Failed to add data

Can add data

1.Create Table
CREATE TABLE "全类型表" (
  "col1" NUMBER(*,0),
  "col2" NUMBER,
  "col3" CHAR(120),
  "col4" VARCHAR2(120),
  "col5" VARCHAR2(120),
  "col6" BLOB,
  "col7" CLOB,
  "col8" DATE,
  "col9" TIMESTAMP(6),
  "col10" TIMESTAMP(6) WITH TIME ZONE,
  "col11" TIMESTAMP(6) WITH LOCAL TIME ZONE,
  "col12" RAW(200),
  "col13" INTERVAL YEAR (2) TO MONTH,
  "col14" INTERVAL DAY (2) TO SECOND (6),
  "col15" NCHAR(120),
  "col16" NVARCHAR2(120),
  "col17" FLOAT(126),
  "col18" BINARY_FLOAT,
  "col19" BINARY_DOUBLE,
  "col20" UROWID(4000),
  "col21" UROWID(4000)
) COMPRESS FOR ARCHIVE REPLICA_NUM = 1 BLOCK_SIZE = 16384 USE_BLOOM_FILTER = FALSE TABLET_SIZE = 134217728 PCTFREE = 0;

2. Insert data within the data
6c277c36437ce12b0826c266b185ae28

_No response_

### Cloud

_No response_

PeachThinking:
#### What type of PR is this?
type-bug
module-sql execution
module-resultset
table with rowid datatype cannot be edited both in query table data and in sql-execute resultset.
table ddl:
```
CREATE TABLE "JINGTIAN"."ROWID_COL" (
  "ID1" UROWID(4000),
  "ID2" UROWID(4000)
)
```
this table's data cannot be edited both in query table data and in sql-execute resultset.
when executing sql like `' select * from ROWID_COL'`, the resultset cannot be edited.

Fixes #370

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/session/model/SqlExecuteResult.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/session/model/SqlExecuteResult.java
@@ -146,6 +146,9 @@ public class SqlExecuteResult {
         boolean editable = true;
         OdcTable resultTable = null;
         Set<OdcTable> relatedTablesOrViews = new HashSet<>();
+        Set<String> rowid_column_names = new HashSet<>();
+        rowid_column_names.add(OdcConstants.ODC_INTERNAL_ROWID);
+        rowid_column_names.add(OdcConstants.ROWID);
         if (Objects.isNull(this.resultSetMetaData)) {
             return null;
         }
@@ -176,7 +179,8 @@ public class SqlExecuteResult {
             } else {
                 columnNames.add(columnName);
             }
-            if (Types.ROWID == odcFieldMetaData.getColumnType()) {
+            if (Types.ROWID == odcFieldMetaData.getColumnType()
+                    && rowid_column_names.contains(odcFieldMetaData.getColumnName())) {
                 // not editable if there is more than one RowId
                 if (Objects.nonNull(resultTable)) {
                     editable = false;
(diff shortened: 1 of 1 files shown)


yizhouxw:
where does this issue comes from ?
seems urowid type in a table is not a commons usage, urowid often referenced in PL program block.

however, current solution may result into result edition not works while sql statement use alias for rowid type column.
e.g.  `select t.ROWID as MY_ROWID, t.* from test1 t `;

if the urowid type in table ddl, we may improve editable judgement:
- just ignore urowid type columnd defined in table

PeachThinking:
> where does this issue comes from ? seems urowid type in a table is not a commons usage, urowid often referenced in PL program block.
> 
> however, current solution may result into resultset edition not works while sql statement use alias for rowid type column. e.g. `select t.ROWID as MY_ROWID, t.* from test1 t `;
> 
> if the urowid type in table ddl confirmed as a common usage, we may improve editable judgement:
> 
> * just ignore urowid type columnd defined in table

The essential reason for this bug is that for the data types of `ROWID` and `UROWID`, the `ResultSet` returned by `jdbc` is uniformly regarded as `Types.ROWID`, so ODC cannot distinguish the data types of `ROWID` and `UROWID`.
Then this bug will be treated as a `won’t fix`？

PeachThinking:
Considering that the data type of UROWID is rarely used in actual business scenarios, and this issue is only raised by QA, the current repair solution has more advantages than disadvantages. This bug will not be fixed for the time being. We will deal with it if subsequent customers encounter this scenario.

PR #398, opened 2023-09-25
fix(data-masking): add unit test case for select sql syntax

smallsheeeep:
#### What type of PR is this?
type-refactor
module-Data masking

Add missing unit test case


yhilmare (approved):
lgtm

PR #440, opened 2023-10-08
fix(dlm): validate condition by sql explain.

guowl3:
#### What type of PR is this?
type-bug

Validate condition timeout when the amount of data is too large

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/flowtask/AbstractDlmJobPreprocessor.java
```diff
@@ -121,8 +121,9 @@ private void checkDataArchiveSql(ConnectionSession connectionSession, Map<DataAr
         // Ensure the conditions are valid when executing.
         sqlMap.forEach((key, value) -> {
             try {
-                syncJdbcExecutor.execute(value);
+                syncJdbcExecutor.execute("explain" + value);
```
yizhouxw, line 124: miss whitespace after explain ?

yizhouxw (approved):
lgtm

PR #485, opened 2023-10-11
fix(data-security): data masking failed when using uppercase column name

### LINKED ISSUE #481: [Bug]: Failed to execute SQL to view result set desensitization
### ODC version

odc422

### OB version

not have

### What happened?

image
image

Successfully desensitized the result set of executing SQL query

1. Set sensitive columns
2. Execute Query SQL

_No response_

### Cloud

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
module-data security

In MySQL mode, if the column name is uppercase, then the data masking will fail.

The sensitive column name stored in ODC metaDB is case-insensitive when executing comparison. So we should compare the column name ignoring case. If there are two columns with the same name but different case, then we will recognize both of them as sensitive columns even though only one of them is marked sensitive.

This PR fixed it by using `equalsIgnoreCase` to compare column names.

Fixes #481 

Passed self-test


yhilmare (approved):
lgtm

PR #486, opened 2023-10-11
fix(osc): execute pre and post interceptor in retry rename table

krihy:
#### What type of PR is this?

type-refactor in  online schema change
before rename and swap table we may be do some pre interceptor, eg lock table or user, this action may failed for some other reason, but swap table can retry for times, so put preRename、 renameFailed and postRenamed method put in
`do  while`.  

1.add swap table name test case for ob oracle 
2.fix ob mysql swap table test case

```docs

```


yhilmare (approved):
lgtm

PR #488, opened 2023-10-11
feat(connection): add initialization configuration capabilities for data sources

### LINKED ISSUE #93: [Feature]: add initialization configuration capabilities for data sources

Users want to init the connection when they create an ODC session. 

e.g. The default query timeout settings is too small for someone, and he/she wants to give a bigger value when he/she connect to the observer. What he/she should is as follow:
1. click a connection config link and create an ODC session.
2. type `set session ob_query_timeout=xxx`

which is too complicated. Further more, the settings doesn't work when he/she exit and re-connect.

This solution is based on some other's suggestion, and I will give a brief description:

1. provides the ability to configure the properties of the jdbc driver.
2. provides the ability to configure the variables of the db.

### Additional context

_No response_

yhilmare:
#### What type of PR is this?

type-feature

this pr add initialization configuration capabilities for data sources.

User can give a sql script that executed when jdbc connection created to init the connection. 

eg. if you want to set your sql execution timeout, you can give a sql script like this:
```SQL
set session ob_query_timeout=60000000;
set session ob_trx_timeout=60000000;
```
odc will execute them one by one when you want to create a session on this datasource.

odc connects to observer by jdbc, if you want to add some your custom jdbc parameters, you can to this by this pr. odc will append your custom jdbc parameters to the url.

this init config will work in every session creation scenario.

Fixes #93 

I create a new table named `connec_connection_attribute`, this table will be not only used by this feature, but also used in other scenario.

the `SessionCreatedInitializer` is used to execute user's custom sql init script, and `OBConsoleDataSourceFactory#getJdbcParams` is used to append users' custom jdbc parameters to odc's default parameters, odc's jdbc url parameters got the highest priority.

this init for connection will work in two scenarios:
1. session creation: if you want to create a session for a datasource, this init will work.
2. datasource test: when you test a datesource,  these init measures will also work.

```docs

```

### DIFF
--- a/server/odc-migrate/src/main/resources/migrate/common/V_4_2_2_3__add_connect_connection_attribute.sql
+++ b/server/odc-migrate/src/main/resources/migrate/common/V_4_2_2_3__add_connect_connection_attribute.sql
@@ -1,0 +1,15 @@
+---
+--- v4.2.2
+---
+CREATE TABLE IF NOT EXISTS `connect_connection_attribute` (
+    `id` bigint(20) NOT NULL AUTO_INCREMENT COMMENT 'Id for connection attribute',
+    `create_time` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT 'Record insertion time',
+  	`update_time` datetime DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP COMMENT 'Record modification time',
+    `name` varchar(1024) NOT NULL COMMENT 'Name for an attribute',
+    `organization_id` bigint(20) NOT NULL COMMENT 'Organization id',
+		`connection_id` bigint(20) NOT NULL COMMENT 'Related connection id, reference connect_connection(id)',
+    `content` mediumtext DEFAULT NULL COMMENT 'Content for key',
+    CONSTRAINT `pk_connect_connection_attribute` PRIMARY KEY (`id`),
+    KEY `idx_connect_connection_attribute` (`connection_id`, `organization_id`),
+    UNIQUE KEY `uk_connect_connection_attribute` (`name`, `connection_id`, `organization_id`)
+);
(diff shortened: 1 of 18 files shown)


server/odc-migrate/src/main/resources/migrate/common/V_4_2_2_3__add_connect_connection_attribute.sql
```diff
@@ -0,0 +1,15 @@
...
+CREATE TABLE IF NOT EXISTS `connect_connection_attribute` (
+    `id` bigint(20) NOT NULL AUTO_INCREMENT COMMENT 'Id for connection attribute',
+    `create_time` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT 'Record insertion time',
+  	`update_time` datetime DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP COMMENT 'Record modification time',
+    `name` varchar(1024) NOT NULL COMMENT 'Name for an attribute',
+    `organization_id` bigint(20) NOT NULL COMMENT 'Organization id',
+		`connection_id` bigint(20) NOT NULL COMMENT 'Related connection id, reference connect_connection(id)',
+    `content` mediumtext DEFAULT NULL COMMENT 'Content for key',
+    CONSTRAINT `pk_connect_connection_attribute` PRIMARY KEY (`id`),
+    KEY `idx_connect_connection_attribute` (`connection_id`, `organization_id`),
```
yizhouxw, line 13: index column order, organization_id before conenciton_id

server/odc-migrate/src/main/resources/migrate/common/V_4_2_2_3__add_connect_connection_attribute.sql
```diff
@@ -0,0 +1,15 @@
...
+    UNIQUE KEY `uk_connect_connection_attribute` (`name`, `connection_id`, `organization_id`)
```
yizhouxw, line 14: the uk_connect_connection_attribute should cover idx_connect_connection_attribute.
order: organization_id, connection_id, name.

yhilmare:
> Does this pr also support native mysql?

sure

yizhouxw (approved):
lgtm

PR #489, opened 2023-10-12
feat:(osc): monitor user lock status and relational sessions

krihy:
#### What type of PR is this?

This is new feature in module of online schema change

Odc users can lock assigned db user when ob version lack of table lock,  this improvement  can decrease affect on  other db user except application user 

1. `OscDBAccessor` add  method  `listUsers` for query db user accooutStatus
2.  add `DBUserMonitor` which is task for start and stop  monitor user status and can be executed separated thread
3.  add `DBUserMonitorFactory` for generate `DBUserMonitor`
4.  add `DBUserMonitorExecutor` which invoked by`SwapTableNameValve` to start or stop monitor

```docs

```


yhilmare (approved):
lgtm

PR #491, opened 2023-10-12
feat(dbbrower): add method detailUsers in DBSchemaAccessor

krihy:
#### What type of PR is this?

type-feature, module-dbbrower
DBSchemaAccessor contains  method listUsers already,  but we may know detail information about user, eg user lock status 

Add `detailUsers` method in DBSchemaAccessor to query user detail information.  current method return result is `List<DBUser>` which  add `DBUserLockType` to indicate user is locked or not.

OBMySQLSchemaAccessor  and  OBOracleSchemaAccessor implements DBSchemaAccessor  but MySQLSchemaAccessor and OracleSchemaAccessor don't,  because don't use this feature in mysql and oracle at present
 
```docs

```


libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/model/DBUserLockType.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package com.oceanbase.tools.dbbrowser.model;
+
+/**
+ * @author yaobin
+ * @date 2023-09-27
+ * @since 4.2.3
+ */
+public enum DBUserLockType {
```
yizhouxw, line 24: if locktype, then boolean type will by a simple model, e.g.  accountLocked.

however, the db-browser module was designed as an general model for RDBMS.
we may follow oracle design, `account_status` , refer https://docs.oracle.com/en/database/oracle/oracle-database/19/refrn/DBA_USERS.html#GUID-309FCCB2-2E8D-4371-9FC5-7F3B10E2A8C0 .

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/OBMySQLSchemaAccessor.java
```diff
@@ -121,6 +123,23 @@ public List<DBObjectIdentity> listTables(String schemaName, String tableNameLike
         return results;
     }
 
+    @Override
+    public List<DBUser> detailUsers(List<String> usernames) {
+        MySQLSqlBuilder sb = new MySQLSqlBuilder();
+        sb.append("SELECT user_name, is_locked FROM oceanbase.__all_user");
```
yizhouxw, line 129: avoid internal table references. use compatible dictionary view instead.

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/OBMySQLSchemaAccessor.java
```diff
@@ -121,6 +123,23 @@ public List<DBObjectIdentity> listTables(String schemaName, String tableNameLike
         return results;
     }
 
+    @Override
+    public List<DBUser> detailUsers(List<String> usernames) {
```
yizhouxw, line 127: query user detail by usernames list is not a common scenario.

what if listAllUsers ?

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/oracle/OBOracleSchemaAccessor.java
```diff
@@ -137,6 +139,23 @@ public List<DBDatabase> listDatabases() {
                 .collect(Collectors.toList());
     }
 
+    @Override
+    public List<DBUser> detailUsers(List<String> usernames) {
+        SqlBuilder sb = new OracleSqlBuilder();
+        sb.append("SELECT user_name, is_locked FROM SYS.ALL_VIRTUAL_USER_REAL_AGENT");
```
yizhouxw, line 145: avoid internal view access

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/model/DBUserLockType.java
krihy: add DBUser in osc model which contains DBAccountLockType to indicate locked or not, and  DBAccountLockType can be easy transfer from db query result, and DBAccountLockType can convenient to show in log

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/OBMySQLSchemaAccessor.java
krihy: this method `detailUsers` has placed in `OscDBAccessor` which is osc service code

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/oracle/OBOracleSchemaAccessor.java
krihy: see response above

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/OBMySQLSchemaAccessor.java
krihy: This PR will be close, query user account lock status will put in osc service

PR #492, opened 2023-10-12
feat(mysql): native mysql adapt for safety regulations

### LINKED ISSUE #2: [Feature]: ODC supports MySQL data source
**Value of feature**

Support MySQL as one of the mainstream databases, can solve the problem of MySQL users database development, increase the audience of ODC product capabilities.

**Description of feature**

Right now, ODC just support OceanBase, but if we want to touch more user, we can go further and try to support MySQL, Postgres, even Oracle. 
Here is the voice of community. 
But this is a tough decision which will determine where we go in the next few years, we had better make more preparation for it. 
https://github.com/oceanbase/oceanbase/issues/881
https://ask.oceanbase.com/t/topic/28800067

**Other instructions**

Object tree presentations, SQL Windows, result set operations, and database changes need to be adapted to the MySQL data source

**Additional context**

PeachThinking:
#### What type of PR is this?
type-feature
module-security control

native mysql adapt for safety regulations.
The following 3 rules do not support mysql:

- sql-check.prefer-local-index
- sql-check.restrict-column-name-case
- sql-check.restrict-table-name-case

See the self-test results at related documents.

Fixes #2

```docs

```

### DIFF
--- a/server/odc-migrate/src/main/resources/init-config/init/regulation-rule-applying.yaml
+++ b/server/odc-migrate/src/main/resources/init-config/init/regulation-rule-applying.yaml
@@ -6,6 +6,7 @@
   appliedDialectTypes:
       - 'OB_MYSQL'
       - 'OB_ORACLE'
+      - 'MYSQL'
   propertiesJson: '{}'
 - enabled: 1
   level: 1
@@ -14,6 +15,7 @@
   appliedDialectTypes:
       - 'OB_MYSQL'
       - 'OB_ORACLE'
+      - 'MYSQL'
   propertiesJson: '{}'
 - enabled: 1
   level: 1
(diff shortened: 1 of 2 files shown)


server/odc-migrate/src/main/resources/init-config/init/regulation-rule-applying.yaml
```diff
@@ -38,6 +42,7 @@
   appliedDialectTypes:
       - 'OB_MYSQL'
       - 'OB_ORACLE'
+      - 'MYSQL'
```
yhilmare, line 45: odp-sharding？

yhilmare (approved):
lgtm

PR #529, opened 2023-10-17
fix(osc): fix get cloud main account id throw exception when environment is not cloud

krihy:
#### What type of PR is this?

type-debug
module-online schema change

main account id will be null when odc environment is not cloud

when main account id is null return null instead of throw exception

```docs

```


yhilmare (approved):
lgtm

PR #539, opened 2023-10-17
feat(osc): odc user can assign lock db user when create osc task

krihy:
#### What type of PR is this?

type-feature
module-online schema change
assign db user(s) to lock exactly between swap table name in osc task, and this does not affects others db users

1. add api `listUsers` to query users in selected datasource
2. modify `/database/databases/{id}` add column `lockDatabaseUserRequired` which indicate current database should lock db user between swap table
3. modify `createFlowInstance` add column `lockUsers` which indicate users to locked
4. modify `LockUserInterceptor`, old way is lock all users except db administrator,  new way is lock db user from odc user input

this feature has coordinate with font end.

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
```diff
@@ -574,6 +605,25 @@ public Set<String> filterUnAuthorizedDatabaseNames(Set<String> databaseNames, @N
                 .collect(Collectors.toSet());
     }
 
+    @SkipAuthorize("internal authorized")
+    public Page<DataBaseUser> listUsers(Long dataSourceId) {
```
yizhouxw, line 609: if listAll, List may better than Page

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
```diff
@@ -631,4 +681,23 @@ private String getLockKey(@NonNull Long connectionId) {
         return "DataSource_" + connectionId;
     }
 
+    private Pair<ConnectionSession, Boolean> getConnectionSessionLockUserRequiredPair(
```
yizhouxw, line 684: too complex, avoid return Pair if not neccessary, how about return boolean instead ?

server/odc-core/src/main/java/com/oceanbase/odc/core/shared/constant/ErrorCodes.java
```diff
@@ -164,6 +164,7 @@ public enum ErrorCodes implements ErrorCode {
     OscSqlTypeInconsistent,
     NoUniqueKeyExists,
     OscNotEnabled,
+    OscNotSelectLockUser,
```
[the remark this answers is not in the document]
krihy: Okey, font-end has restriction input parameters, back-end add restriction for enhancement

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
krihy: okey,  ConnectionSession can be close  after predicate lock user or not

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
krihy: this method definition stay the same style like `/api/v2/database/databases` or other previous api

yizhouxw (approved):
lgtm

Issue #506, opened 2023-10-16
[Bug]: no traceId on OceanBase 4.2


LuckyPickleZZ:
### ODC version

4.2.0

### OB version

OB 4.2

### What happened?

No trace id was captured. Because ODC use `show trace format='json'` to get trace_id. And we haven't support OB 4.2 .

### What did you expect to happen?

Could get trace id when executing sql.

### How can we reproduce it (as minimally and precisely as possible)?

Any OB 4.2 .

sl01388797:
![Image](https://github.com/oceanbase/odc/assets/143492234/24337e4d-6c8f-4555-a718-dae463b92173)



![Image](https://github.com/oceanbase/odc/assets/143492234/3bbaea60-f9c5-4ed1-a37f-e79101919644)

pass

PR #541, opened 2023-10-17
feat: make odc adapt to OceanBase 4.2

### LINKED ISSUE #94: [Feature]: adapt to OceanBase 4.2

OceanBase 4.2 is released, and some new features is available now, ODC may not support.

adapt to OceanBase 4.2

### Additional context

_No response_

yhilmare:
#### What type of PR is this?

type-feature
feat-adapt to ob 4.2

this pr makes odc adapt to OceanBase 4.2. we do things in this pr as follow:

1. add schema in generated flashback sql.

the front end will create a new session when using recyclebin so that the observer will flashback the object to the schema where the connection connect to.

To avoid flashbacking a db object to an uncertain schema, we should add schema to the generated sql. 

Here is an example:
```SQL
flashback table `abcd` to before drop rename to `new_name`; -- before
flashback table `abcd` to before drop rename to `schema`.`new_name`; -- after
```

2. refactor the pojo object's (`TraceSpan`) structure of full-link trace and convert the field `tags`'s type from `List<Map<String, Object>>` to `List<Object>`. 

we will get the json string on OceanBase 4.1 when `show trace format='json'` is executed:

```json
[
    {
        "logs": null,
        "tags": [
           ...
            {
                "log_trace_id": "Y2F170BA2D939-000602A2DA7F897E-0-0"
            }
        ],
        ...
```

we do the same thing on OceanBase 4.2.1 and get a json string as follow:

```json
[
    {
        "logs": null,
        "tags": [
            [
                ...
                {
                    "log_trace_id": "Y2FB70BA2D939-000608DF99AE2C76-0-0"
                }
            ]
        ],
       ...
```

as you can see: the structure of the json has been changed, and that is why the full-link trace does not work on OceanBase 4.2.

I change the structure of the pojo object and find another way to avoid this change.

Fixes #94 #506 

```docs

```

### DIFF
--- a/server/odc-core/src/main/java/com/oceanbase/odc/core/sql/util/FullLinkTraceUtil.java
+++ b/server/odc-core/src/main/java/com/oceanbase/odc/core/sql/util/FullLinkTraceUtil.java
@@ -70,7 +70,7 @@ public class FullLinkTraceUtil {
                     return v;
                 });
             }
-            if (traceId == null && findLogTraceId(span.getTags()) != null) {
+            if (traceId == null) {
                 traceId = findLogTraceId(span.getTags());
             }
             span.setNode(Node.from(span.getSpanName()));
--- a/server/odc-core/src/main/java/com/oceanbase/odc/core/sql/util/FullLinkTraceUtil.java
+++ b/server/odc-core/src/main/java/com/oceanbase/odc/core/sql/util/FullLinkTraceUtil.java
@@ -18,6 +18,7 @@ package com.oceanbase.odc.core.sql.util;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


server/odc-core/src/main/java/com/oceanbase/odc/core/sql/util/FullLinkTraceUtil.java
```diff
@@ -67,8 +70,8 @@ private static SqlExecTime parseSpanList(List<TraceSpan> spanList) {
                     return v;
                 });
             }
-            if (traceId == null && parseLogTraceId(span.getTags()) != null) {
-                traceId = parseLogTraceId(span.getTags());
+            if (traceId == null && findLogTraceId(span.getTags()) != null) {
```
smallsheeeep, line 73: Why execute `findLogTraceId(span.getTags())` twice?

yhilmare: I don't know 😂, and I just refactor these codes to another way that can be executed on ob 4.2.1. @LuckyPickleZZ may answer this question

yhilmare: okay, I find another way to get traceId that execute `findLogTraceId(span.getTags())` only once

smallsheeeep (approved):
LGTM

PR #544, opened 2023-10-18
feat(dlm): data archive supports MySQL to OB.

guowl3:
#### What type of PR is this?
type-feature
module-data lifecycle management

The current data archiving task only supports OB to OB, but there are many MySQL currently used by customers. We need to support the archiving link from MySQL to OB to meet customer demands.
The content of this PR is mainly about making DLM tasks compatible with MySQL data sources.

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/dlm/DataSourceInfoBuilder.java
```diff
@@ -0,0 +1,73 @@
...
+                        dataSourceInfo.setSysPassword(EncryptUtils.encode(connectionConfig.getSysTenantPassword()));
+                    } catch (Exception e) {
+                        throw new RuntimeException(e);
+                    }
+                }
+                dataSourceInfo.setSysDatabaseName("oceanbase");
+                break;
+            }
+            default:
+                throw new UnsupportedException();
```
yizhouxw, line 69: should contains context info while throw exception

server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/job/DataArchiveJob.java
```diff
@@ -56,4 +61,49 @@ public void execute(JobExecutionContext context) {
...
+    }
+
+
+    /**
+     * Create the table in the target database before migrating the data.
+     */
+    private void createTargetTable(DlmTask dlmTask) {
+
+        if (dlmTask.getSourceDs().getDialectType() != dlmTask.getTargetDs().getDialectType()) {
+            log.info("Data sources of different types do not currently support automatic creation of target tables.");
```
yizhouxw, line 78: for mysql --> ob mysql mode tenant, we expect the dialectType same or not?

server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/job/DataArchiveJob.java
```diff
@@ -56,4 +61,49 @@ public void execute(JobExecutionContext context) {
...
+        }
+
+        DefaultConnectSessionFactory targetConnectionSessionFactory =
+                new DefaultConnectSessionFactory(dlmTask.getTargetDs());
+        ConnectionSession targetSession = targetConnectionSessionFactory.generateSession();
+        try {
+            DBSchemaAccessor targetDsAccessor = DBSchemaAccessors.create(targetSession);
+            List<String> tableNames = targetDsAccessor.showTables(dlmTask.getTargetDs().getDefaultSchema());
+            if (tableNames.contains(dlmTask.getTableName())) {
+                log.info("Target table exist.");
```
yizhouxw, line 99: refine log output, use comma separated style, contians releated context information, such as
```
 log.info("Target table exist, tableName={}", balbla);
 log.info("Begin to create target table, tableName={}", blabla);
```

server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/job/DataArchiveJob.java
guowl3: We don't expect it to be the same in different modes.

guowl3: We don't expect it to be the same in different modes.

guowl3: We don't expect it to be the same in different modes.

yizhouxw (approved):
lgmt

yizhouxw (approved):
lgtm

Issue #540, opened 2023-10-17
[Bug]: report "PACKAGE BODY 'SYS.OBODC_PL_RUN_PACKAGE' does not exist repeatedly


xiaodp:
### ODC version

版本号：4.2.0
Server: 4.2.0-20230818
Release Date: 2023-08-17

### OB version

OceanBase 构建版本号：4.2.1.0-100000182023092722

### What happened?

After creating a oracle mode tenant on OCP( 4.2.0-20230825), I add the SYS user as a datasource via obproxy into the ODC Private Space and do some queries. But the obproxy started to repeatedly report error log :
```
2023-10-17 19:08:23.664719,xxx,,,,xxx:{TENANT_NAME}:SYS,OB_ORACLE,,,COM_QUERY,DROP,failed,600,DROP PACKAGE BODY "SYS".OBODC_PL_RUN_PACKAGE,3786us,23us,0us,3701us,Y0-00007F96B806BDE0,YB420A15FC2B-0006074950278F62-0-0,,,0,10.21.252.43:2881,ORA-00600: internal error code, arguments: -5559, PACKAGE BODY 'SYS.OBODC_PL_RUN_PACKAGE' does not exist
```
I suppose this has no business with obproxy and ocp, but there exists some bug that makes ODC call ``` DROP PACKAGE BODY "SYS".OBODC_PL_RUN_PACKAGE``` all the time, in which the OBODC_PL_RUN_PACKAGE has been deprecated since ODC 4.2

ps: the tenant is accessed only by ODC 4.2 web version, no odc 4.1 or desktop version

### What did you expect to happen?

not to execute SQL "DROP PACKAGE BODY "SYS".OBODC_PL_RUN_PACKAGE" if OBODC_PL_RUN_PACKAGE doesn't exist

### How can we reproduce it (as minimally and precisely as possible)?

1. create an oracle tenant 
2. add it into odc 4.2 web version Private Space
3. do some query
4. check the obproxy error log

yhilmare:
I will remove the code which attempt to drop pkg

PR #548, opened 2023-10-18
fix(pl): remove stored procedure test calls and built-in package automatic deletion logic

### LINKED ISSUE #257: [Bug]: long time elapsed when invoke obodc_procedure_feature_test function
### ODC version

4.2.0

### OB version

Any OB version

### What happened?

ODC will actively call the obodc_procedure_feature_test function. The SQL that calls this function takes a long time to execute, causing OCP to generate an alarm.

Can odc delete obodc_procedure_feature_test funcation invocation?

<null>

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-feature

1. Remove the auto deletion logic for package `OBODC_PL_RUN_PACKAGE`.

ODC used to print `dbms_output` message through the package `OBODC_PL_RUN_PACKAGE`. We find another way using anonymous block to do the same thing on ODC 4.2.0. 

Thus, ODC will delete `OBODC_PL_RUN_PACKAGE` automatically whether it exists or not, so that it may cause a lot of error. Some users do not like this, We remove this auto deletion to avoid the error printing.

2. Remove the procedure test calls.

ODC will try to call a non-existent program to verify whether the OB supports the program. This logic will also cause many errors and even take a long time, so we delete it.

Fixes #257 #540 

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/feature/VersionDiffConfigService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/feature/VersionDiffConfigService.java
@@ -100,7 +100,8 @@ public class VersionDiffConfigService {
         config.setDbMode(getDbMode(connectionSession));
         List<VersionDiffConfig> list = this.versionDiffConfigDAO.query(config);
         String currentVersion = ConnectionSessionUtil.getVersion(connectionSession);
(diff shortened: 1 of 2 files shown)


yizhouxw (approved):
LGTM

PR #549, opened 2023-10-19
fix(unit-test): unsafe log output and unstable test case

smallsheeeep:
#### What type of PR is this?
type-bug
module-unit test

The `SystemUtilsTest` has unsafe log output, which will expose sensitive env variables.
The `ExpiredDocumentProviderTest#provide_BothExpired_ReturnAll` is unstable due to not strictly logical.
This PR fix these two issues.


yhilmare (approved):
lgtm

PR #551, opened 2023-10-19
fix(databasechange): fix task costs too much time to start up

### LINKED ISSUE #550: [Bug]: database change task costs too much time to start
### ODC version

4.2.1

### OB version

not related

### What happened?

I upload a big sql script (about 60M) when using database change task. ODC costs about 30mins to start this task (download sql script from OSS/metaDB, split sql script, etc. all these operation are executed in main thread) which causes error printing (ODC will startup a monitor thread to record task's status, and this thread will report error when task is not started up for a long time).

database change task can be started up for a short time

QAer will not care about this issue, internal improvement.

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

databasechange task takes too long time to start up which causes error printing by monitor task. I move the sql script read logic to task running phase to reduce the startup time.

Fixes #550 

```docs

```


MarkPotato777 (approved):
LGTM

Issue #70, opened 2023-08-28
[Feature]: Adapt bastion integration to new permission control model


smallsheeeep:
### Is your feature request related to a problem?

ODC 4.2.0 has made a major upgrade to the management and control model, which needs to be adapted to the official website experience station.

### Describe the solution you'd like

None for yet.

HSunboy:
ODC implemented the single data mode in personal space in dev-4.2.2

smallsheeeep:
I will close this issue and the following bastion integration issue is #672

PR #559, opened 2023-10-19
fix(integration): recover bastion integration

yizhouxw:
#### What type of PR is this?

type-bug

the ODC v4.2.0 involve much changes, broke bastion integration feature.
here we recover account integration and auto datasource configuration.

Fixes #558 To #70 
 

there exits migration error for OAuth2 ingration configuration,
which result into OAuth2 configuration not works if there exits multiple ogranization.

this pr also contains some build script fix, we can use local build by script/build_jar.sh now.

```docs

```

### DIFF
--- a/script/start-odc.sh
+++ b/script/start-odc.sh
@@ -110,8 +110,8 @@ function init_jvm_options() {
     fi
     local log_options="-Dlog4j.configurationFile=${app_log_config_file} -Dodc.log.directory=${app_log_directory}"
     local work_dir_options="-Duser.dir=${ODC_WORK_DIR:-${current_work_directory}}"
-    local plugin_options="--plugin.dir=${plugin_directory}"
-    local starter_options="--starter.dir=${starter_directory}"
+    local plugin_options="-Dplugin.dir=${plugin_directory}"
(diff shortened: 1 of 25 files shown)


script/start-odc.sh
```diff
@@ -110,15 +110,16 @@ function init_jvm_options() {
     fi
     local log_options="-Dlog4j.configurationFile=${app_log_config_file} -Dodc.log.directory=${app_log_directory}"
     local work_dir_options="-Duser.dir=${ODC_WORK_DIR:-${current_work_directory}}"
-    app_options="${log_options} ${work_dir_options}"
+    local plugin_options="-Dplugin.dir=${plugin_directory}"
```
yizhouxw, line 113: plugins initialzied before spring cloud config, so we should use app options (`-Dblabla` ) instead of app args(`--blabla` )

yhilmare:
there are two error test cases:
截屏2023-10-20 11 37 28

yizhouxw:
> there are two error test cases: 截屏2023-10-20 11 37 28

fixed now

yhilmare (approved):
lgtm

PR #566, opened 2023-10-20
fix(pl-debug): fix failed to step in a subprocedure or subfunction defined in package

### LINKED ISSUE #560: [Bug]: failed to step in a sub-procedure of a package in pl-debug
### ODC version

4.2.1

### OB version

not related

### What happened?

define a package and package body as follow:
```SQL
-- package
create or replace package test_pl_debug_pkg AS
 procedure pkg_inner_proc(p1 in integer);
 function pkg_inner_func(p1 in integer) return varchar2;
end;
-- package body
create or replace PACKAGE BODY test_pl_debug_pkg AS
function pkg_inner_func(p1 in integer) return varchar2 AS
BEGIN
	dbms_output.put_line(p1);
	RETURN 123;
END;
procedure pkg_inner_proc(p1 in integer) AS
BEGIN
	dbms_output.put_line(p1);
END;
END test_pl_debug_pkg
```

I create a procedure as follow and debug it:
```SQL
create or replace procedure test_pl_debug_proc(p1 in integer) AS
begin
    test_pl_debug_pkg.pkg_inner_proc(p1);
end;
```
I step in `test_pl_debug_pkg.pkg_inner_proc` when I run on this line:
截屏2023-10-19 20 04 06

 I can step in this sub-procedure successfully.

follow 'what happened'

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug
module-PL Debugger

this is fix two issues:

1. Some user will add some ob's grammar to PL body, such as `set autocommit='OFF'` which is not supported by oracle's parser and lead to error report when debugging.

This pr supports set_variables statement and count(*) in pivot clause to avoid syntax error.

2. Fix can not step in a sub-procedure or sub-function that defined by package.

ODC can only supports step in a standalone procedure or function in anonymous block which is a bug and we solve this by handling the two cases separately.

Fixes #560 

```docs

```

### DIFF
--- a/libs/ob-sql-parser/src/main/resources/oracle/PlSqlParser.g4
+++ b/libs/ob-sql-parser/src/main/resources/oracle/PlSqlParser.g4
@@ -3342,6 +3342,7 @@ statement
     | function_call
     | pipe_row_statement
     | procedure_call[true]
+    | set_variable
     ;
 
 swallow_to_semi
--- a/libs/ob-sql-parser/src/main/resources/oracle/PlSqlParser.g4
+++ b/libs/ob-sql-parser/src/main/resources/oracle/PlSqlParser.g4
@@ -122,6 +122,7 @@ unit_statement
     | grant_statement
 
     | procedure_call[false]
+    | set_variable
     ;
 
 // DDL -> SQL Statements for Stored PL/SQL Units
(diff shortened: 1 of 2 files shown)


yizhouxw (approved):
lgtm

PR #576, opened 2023-10-23
feat(pl-debug): add timeout settings

### LINKED ISSUE #251: [Enhancement]: add timeout control for PL debug

Allow users to configure the timeout during pl debugging, and we can solve this problem through the requirements of data source initialization configuration #93

### Why is this needed?

Users may encounter timeout problems when performing PL debugging

yhilmare:
#### What type of PR is this?

type-debug

make session init script works in pl-debug

Fixes #251 

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/ConnectionConfig.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/ConnectionConfig.java
@@ -404,6 +404,7 @@ public class ConnectionConfig
         return queryTimeoutSeconds;
     }
 
+    @Size(max = 8192, message = "Session init script is out of range [0,8192]")
     public String getSessionInitScript() {
         if (this.attributes == null) {
             return null;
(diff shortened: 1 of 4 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/ConnectionConfig.java
```diff
@@ -404,6 +404,7 @@ public int queryTimeoutSeconds() {
         return queryTimeoutSeconds;
     }
 
+    @Size(max = 8192, message = "Session init script is out of range [0,8192]")
```
krihy, line 407: why limit script length to 8192

yhilmare: 1. the script length has to be limited
2. 8192 is determined by imaging, there is no more meanings.

krihy:
lgtm

krihy (approved):

PR #583, opened 2023-10-23
fix(pl): fix wrong error message of pl parameter check

### LINKED ISSUE #581: [Bug]: error message is illegal when invalid arguments given
### ODC version

ODC422

### OB version

Oceanbase410

### What happened?

When running PL, a string is passed as the input parameter but an int is actually required, and the interface is 401.
image

execute normally

Run PL
```SQL
create procedure `testBug26143` (IN `I` int(11)) BEGIN
SELECT I * 10;
END;
```

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

the error message of pl parameter checking is illegal and define a new error code to prompt user. 

Fixes #581 

```docs

```


smallsheeeep (approved):
LGTM

PR #586, opened 2023-10-23
fix(result-export): failed to convert CSV file into Excel file

### LINKED ISSUE #457: [Bug]: Export result set in Excel format in ORACLE mode, export failed
### ODC version

odc422

### OB version

not have

### What happened?

image

Export result set successful

Export the result set in Oracle mode, select Excel format, and export

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

When export result set into Excel format, and the records contains `\t` like `xxx\txxx`, OB-DUMPER will use double quote as escape char because of ODC's setting. So the final form in the file is like `...,"xxx"      xxxx",...`.
But when convert it into .xls format, ODC use `org.apache.commons.csv` to parse CSV file. It can not parse this record, and would report IOException.
So when export into csv, we use backslash instead of quote. The above record will be translated like `..."xxx\      xxx"...`, which common-csv could recognize.

Fixes #457

```docs

```


yhilmare (approved):
lgtm

yhilmare (approved):
lgtm

PR #587, opened 2023-10-23
feat(dlm): supports json field.

guowl3:
#### What type of PR is this?
type-feature

OBServer support json field type from 3.2.4, and DLM also needs to support this field type.

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #589, opened 2023-10-23
fix(diagnose): optimize log information when explain failed 

### LINKED ISSUE #439: [Bug]: no log is printed when fail to parse explain result
### ODC version

4.2.0

### OB version

4.2.1

### What happened?

For some reason , ODC failed to parse explain results. There is a exception throwed only with error message. 
But in some cases, such as NPE, the message is just "null". We can not known what the problem was exactly.

Exception would be printed in log files.

Generally, we can not reproduce this bug. 

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

Now when explain execution plan failed, the exception would be eaten. It's hard to find root cause.
So in this PR, the exception and origin plan text will be printed.

Fixes #439


yhilmare (approved):
lgtm

PR #592, opened 2023-10-24
fix(db-browser): cannot get table charset in native mysql mode

### LINKED ISSUE #582: [Bug]: Rename table doesn't show charsetName
### ODC version

ODC 4.2.2

### OB version

mysql 5.7

### What happened?

image
Rename table doesn't show charsetName

Show table's default charsetName.

Rename the table.

_No response_

### Cloud

_No response_

PeachThinking:
#### What type of PR is this?
type-bug
module-Database object

1. Fix cannot get table charset in native mysql mode. OB mysql get table charset by `DBSchemaAccessorUtil#obtainOptionsByParse` which is not work for native mysql mode. 
we can get table by parse table ddl in both mysql and ob mysql mode. add `obtainOptionsByParser` method and use it to get table options.

2. optimize DBSchemaAccessor ut test by create `private static DBSchemaAccessor accessor`
3. fix `@Ignore("TODO: fix this test case")` ut test. I consulted kernel classmates and found that the `comment` field of the `information_schema.STATISTICS` table and the `comment` field of the `show index from` query have different sources and meanings. The `comment` field of `information_schema.STATISTICS` cannot be used to indicate the status of the index, so for `Map<String, List<DBTableIndex >> listTableIndexes(String schemaName); `method does not query whether the index is available. Actually only the `List<DBTableIndex> listTableIndexes(String schemaName, String tableName)` method needs to provide the available status of the index.

Fixes #582

```docs

```

### DIFF
--- a/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoGreaterThan5740SchemaAccessor.java
+++ b/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoGreaterThan5740SchemaAccessor.java
@@ -594,7 +594,6 @@ public class MySQLNoGreaterThan5740SchemaAccessor implements DBSchemaAccessor {
                 columnNames.add(rs.getString("COLUMN_NAME"));
                 index.setColumnNames(columnNames);
                 index.setGlobal(true);
-                handleIndexAvailability(index, rs.getString("COMMENT"));
                 fullIndexName2Index.put(tableName + indexName, index);
             } else {
                 fullIndexName2Index.get(tableName + indexName).getColumnNames()
--- a/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoGreaterThan5740SchemaAccessor.java
+++ b/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoGreaterThan5740SchemaAccessor.java
@@ -1005,7 +1005,11 @@ public class MySQLNoGreaterThan5740SchemaAccessor implements DBSchemaAccessor {
     public DBTableOptions getTableOptions(String schemaName, String tableName, @lombok.NonNull String ddl) {
         DBTableOptions dbTableOptions = new DBTableOptions();
         obtainOptionsByQuery(schemaName, tableName, dbTableOptions);
-        obtainOptionsByParser(dbTableOptions, ddl);
+        try {
+            obtainOptionsByParser(dbTableOptions, ddl);
+        } catch (Exception e) {
+            log.warn("Failed to get table options by parse table ddl, message={}", e.getMessage());
+        }
         return dbTableOptions;
     }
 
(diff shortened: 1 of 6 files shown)


libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoGreaterThan5740SchemaAccessor.java
```diff
@@ -1001,11 +1001,10 @@ public DBTableOptions getTableOptions(String schemaName, String tableName) {
     public DBTableOptions getTableOptions(String schemaName, String tableName, @lombok.NonNull String ddl) {
         DBTableOptions dbTableOptions = new DBTableOptions();
         obtainOptionsByQuery(schemaName, tableName, dbTableOptions);
-        DBSchemaAccessorUtil.obtainOptionsByParse(dbTableOptions, ddl);
```
yhilmare, line 1004: I run `show create table xxx` on MySQL 5.7 and get DDL as follow :
截屏2023-10-24 10 28 52

Why it cannot be obtained through parsing

PeachThinking: ob mysql table ddl: 
image
`CHARSET = utf8mb4`
while mysql ddl:
`CHARSET=latin1`，There are 2 spaces less than obmysql，Therefore `obtainOptionsByParse` cannot be shared

MarkPotato777: The 'parsing' here is simply by regex, which is not robust, e.x., what if the table comment contains something like 'CHARSET = utf8mb4'?
Can we use Parser instead?

yhilmare: parser may do this:
截屏2023-10-24 12 13 26

PeachThinking: add `obtainOptionsByParser` method and use it to get table options.

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoGreaterThan5740SchemaAccessor.java
```diff
@@ -1016,6 +1020,20 @@ private void obtainOptionsByQuery(String schemaName, String tableName, DBTableOp
         });
     }
 
+    private void obtainOptionsByParser(DBTableOptions dbTableOptions, String ddl) {
+        SQLParser sqlParser = new OBMySQLParser();
+        CreateTable stmt = (CreateTable) sqlParser.parse(new StringReader(ddl));
+        TableOptions options = stmt.getTableOptions();
```
yhilmare, line 1026: npe

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoGreaterThan5740SchemaAccessor.java
```diff
@@ -1001,7 +1005,7 @@ public DBTableOptions getTableOptions(String schemaName, String tableName) {
     public DBTableOptions getTableOptions(String schemaName, String tableName, @lombok.NonNull String ddl) {
         DBTableOptions dbTableOptions = new DBTableOptions();
         obtainOptionsByQuery(schemaName, tableName, dbTableOptions);
-        DBSchemaAccessorUtil.obtainOptionsByParse(dbTableOptions, ddl);
+        obtainOptionsByParser(dbTableOptions, ddl);
```
yhilmare, line 1008: catch exception

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoGreaterThan5740SchemaAccessor.java
```diff
@@ -594,7 +599,6 @@ public Map<String, List<DBTableIndex>> listTableIndexes(String schemaName) {
                 columnNames.add(rs.getString("COLUMN_NAME"));
                 index.setColumnNames(columnNames);
                 index.setGlobal(true);
-                handleIndexAvailability(index, rs.getString("COMMENT"));
```
yhilmare, line 597: why this code is deleted?

PeachThinking: I consulted kernel classmates and found that the `comment` field of the `information_schema.STATISTICS` table and the `comment` field of the `show index from` query come from different sources and have different meanings. The `comment` field of `information_schema.STATISTICS` cannot be used to indicate the status of the index, so for `Map<String, List<DBTableIndex >> listTableIndexes(String schemaName); `method does not query whether the index is available. Actually only the `List<DBTableIndex> listTableIndexes(String schemaName, String tableName) `method needs to provide the available status of the index.

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoGreaterThan5740SchemaAccessor.java
PeachThinking: fixed

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoGreaterThan5740SchemaAccessor.java
PeachThinking: fixed

yhilmare (approved):
lgtm

PR #599, opened 2023-10-24
fix(datasource): the data source list refreshes very slowly and cannot obtain the connect status while there are a huge amount of data sources

### LINKED ISSUE #483: [Bug]: The data source list loads slowly when there are a large number of data sources
### ODC version

ODC 4.2.1

### OB version

Any version of OceanBase.

### What happened?

The data source list loads slowly when there are 3000 data sources.
image

Increase the loading speed of the data source list when there are a large number of data sources.

Load 3000 data sources.

_No response_

### Cloud

_No response_

### LINKED ISSUE #484: [Bug]: Data source connection status unavailable When there are a large number of data sources
### ODC version

ODC 4.2.1

### OB version

Any version of OceanBase.

### What happened?

Data source connection status unavailable When there are a large number of data sources.
image

Load data source state when there is a large number of data sources

Load 3000 data sources and view the data source status.

_No response_

### Cloud

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug
module-datasource

While there is a huge amount of data sources (like 3k), the data source list on the left side in the SQL console refreshes very slowly, and the data source connect status would always be loading without response. 

The related API:
- `/api/v2/collaboration/projects/databases/stats`
- `/api/v2/datasource/datasources/status`

The reason for the slow stats API is that when we convert entities to DTOs, our time cost is O(n) because we will query the environment entity for each data source. To solve this issue, we just query the environments only one time, to load the results into the map, and find the specific environment entity by id. In this way, the time cost is reduced to O(1).

The reason for not being able to load the data source status is that the browser limits the URL length and when we have like 3k data sources to test status, our GET request size exceeds the limit. The way to solve this problem is to use the POST method instead and pack parameters in the request body rather than in the URL parameters.

After the optimization, in the scenario of 3k data sources, the stats API RT reduced from 15000 - 16000 ms to 250 - 300 ms.
What's more, the connection status API works now, but it costs 20s under 3k data sources which is not acceptable. The front end should have some strategy to avoid packing all the data sources into one request. @HSunboy 

Fixes #484
Fixes #483 

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/collaboration/environment/EnvironmentService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/collaboration/environment/EnvironmentService.java
@@ -18,7 +18,6 @@ package com.oceanbase.odc.service.collaboration.environment;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -35,7 +34,6 @@ import com.oceanbase.odc.core.shared.constant.ResourceType;
 import com.oceanbase.odc.core.shared.exception.NotFoundException;
 import com.oceanbase.odc.metadb.collaboration.EnvironmentEntity;
 import com.oceanbase.odc.metadb.collaboration.EnvironmentRepository;
-import com.oceanbase.odc.metadb.iam.UserEntity;
 import com.oceanbase.odc.service.collaboration.environment.model.Environment;
 import com.oceanbase.odc.service.iam.HorizontalDataPermissionValidator;
 import com.oceanbase.odc.service.iam.UserService;
@@ -143,17 +141,6 @@ public class EnvironmentService {
     }
 
     private Environment entityToModel(@NonNull EnvironmentEntity entity) {
-        Environment environment = environmentMapper.entityToModel(entity);
-        UserEntity creator = userService.nullSafeGet(entity.getCreatorId());
-        UserEntity lastModifier = userService.nullSafeGet(entity.getLastModifierId());
-        if (Objects.nonNull(environment.getCreator())) {
-            environment.getCreator().setName(creator.getName());
-            environment.getCreator().setAccountName(creator.getAccountName());
-        }
-        if (Objects.nonNull(environment.getLastModifier())) {
-            environment.getLastModifier().setName(lastModifier.getName());
-            environment.getLastModifier().setAccountName(lastModifier.getAccountName());
-        }
-        return environment;
+        return environmentMapper.entityToModel(entity);
     }
 }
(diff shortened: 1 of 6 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/collaboration/environment/EnvironmentService.java
```diff
@@ -143,17 +141,6 @@ private Environment innerDetailWithoutPermissionCheck(@NonNull Long id) {
     }
 
     private Environment entityToModel(@NonNull EnvironmentEntity entity) {
-        Environment environment = environmentMapper.entityToModel(entity);
```
MarkPotato777, line 146: Simply remove the unnecessary query as we do not need the creator/modifier information for now.

smallsheeeep:
Can you streamline the title?

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
```diff
@@ -740,6 +729,39 @@ private ConnectionEntity getEntity(@NonNull Long id) {
...
+        } else {
+            id2Environment = new HashMap<>();
+        }
+        return entities.stream().map(entity -> {
+            ConnectionConfig connection = mapper.entityToModel(entity);
+            connection.setStatus(CheckState.of(ConnectionStatus.TESTING));
+            if (withEnvironment) {
+                Environment environment = id2Environment.getOrDefault(connection.getEnvironmentId(), null);
+                if (Objects.isNull(environment)) {
+                    return connection;
```
smallsheeeep, line 750: return or throw exception?

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
```diff
@@ -740,6 +729,39 @@ private ConnectionEntity getEntity(@NonNull Long id) {
...
+            return Collections.EMPTY_LIST;
+        }
+        Map<Long, Environment> id2Environment;
+        if (withEnvironment) {
+            id2Environment = environmentService.list(organizationId).stream()
+                    .collect(Collectors.toMap(environment -> environment.getId(), environment -> environment));
+        } else {
+            id2Environment = new HashMap<>();
+        }
+        return entities.stream().map(entity -> {
```
yhilmare, line 744: Is method entityToModel still used? if not, prefer deleting it~

MarkPotato777: yes, entityToModel still works in create/update/detail, while entitiesToModels would be used in batch operation.

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
MarkPotato777: Yes, throwing an exception is a better way. Fixed.

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
yhilmare: Maybe use entitiesToModels instead of entityToModel since there is a lot of code duplication in them

yhilmare (approved):
lgtm

smallsheeeep (approved):
LGTM

Issue #605, opened 2023-10-24
[Enhancement]: npe will be thrown when failed to get ob's version


yhilmare:
### What would you like to be added?

throw a human readable exception when failed to get ob's version.

### Why is this needed?

the current implementation will eat the origin exception and gives a npe which is not human readable:
```Java
// OBMySQLInformationExtension
@Override
    public String getDBVersion(Connection connection) {
        String querySql = "show variables like 'version_comment'";
        String v = null;
        try {
            v = JdbcOperationsUtil.getJdbcOperations(connection).query(querySql, rs -> {
                if (!rs.next()) {
                    return null;
                }
                return rs.getString(2);
            });
            Verify.notNull(v, "VersionComment");

        } catch (Exception exception) {
            log.warn("Failed to get ob version", exception);
        }
        return OBUtils.parseObVersionComment(v);
    }
```

or return null which is not acceptable:

```Java
// OBUtils
public static String getObVersion(Connection connection) throws SQLException {
        try (Statement statement = connection.createStatement()) {
            try (ResultSet resultSet = statement.executeQuery("show variables like 'version_comment'")) {
                if (resultSet.next()) {
                    return OBUtils.parseObVersionComment(resultSet.getString("value"));
                }
                return null;
            }
        }

    }
```

Issue #604, opened 2023-10-24
[Bug]: failed to query table data which in interrupted by security policy


yhilmare:
### ODC version

4.2.1

### OB version

not related

### What happened?

double click a table and try to show table data, this operation is interrupted by security policy:
<img width="426" alt="截屏2023-10-24 19 49 33" src="https://github.com/oceanbase/odc/assets/22992296/a68ebe3f-4c99-4891-879d-e0d73d676032">

### What did you expect to happen?

query table data successfully

### How can we reproduce it (as minimally and precisely as possible)?

1. go to security policy and find the environment that the target datasource located:
<img width="773" alt="截屏2023-10-24 19 50 58" src="https://github.com/oceanbase/odc/assets/22992296/1bfd8534-2926-459a-a23e-50361ffe59b7">
2. find `Not Allow Select *` rule in SQL-Check tab and change it's level from 'No need to change' to 'Must change':
<img width="501" alt="截屏2023-10-24 19 52 11" src="https://github.com/oceanbase/odc/assets/22992296/2287f7af-08a2-400d-a18f-c961dfabaf9b">
3. enter the target datasource, double click a table and try to query data

sl01388797:
![Image](https://github.com/oceanbase/odc/assets/143492234/407ba3ff-0bd7-429c-b194-abda5fb3dbab)

![Image](https://github.com/oceanbase/odc/assets/143492234/736ed9c3-263e-433d-9bc9-7f0f7b01766f)
pass

PR #608, opened 2023-10-24
fix: fix failed to query data and sql rules changing is not recorded by audit event

### LINKED ISSUE #606: [Bug]: there is no audit event record about sql-check rule changing
### ODC version

4.2.1

### OB version

not releated

### What happened?

I change the sql-check settings and do not find any audit events.

audit event can record this behavior.

1. change sql-check rule's setting.
2. view audit events.

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug
type-enhancement

1. remove sql check interceptor when query table data.
2. SQL security specification modifications are included in operational audits.
3. define a special error code and throw an exception when failed to query database's version

Fixes #606 #605 #604 

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/session/interceptor/SqlCheckInterceptor.java
```diff
@@ -62,6 +62,7 @@
 @Component
 public class SqlCheckInterceptor implements SqlExecuteInterceptor {
 
+    public final static String NEED_SQL_CHECK_KEY = "NEED_SQL_CHECK_KEY";
```
yizhouxw, line 65: value expected to be 'NEED_SQL_CHECK' ?

server/odc-core/src/main/java/com/oceanbase/odc/core/shared/constant/AuditEventAction.java
```diff
@@ -303,10 +303,9 @@ public enum AuditEventAction implements Translatable {
 
     UPDATE_DATASOURCE,
 
-    CREATE_PROJECT
+    CREATE_PROJECT,
 
-
-    ;
+    CHANGE_SQL_SECURITY_RULE;
```
yizhouxw, line 308: use `UPDATE_` prefix , follow current style.

and, should we distinguish SQL Check Rule & SQL Console Rule ?

yhilmare: both sql check rule and sql console rule use the same api to update and the auditing is based on API level interception, so that we can not distinguish SQL Check Rule & SQL Console Rule

server/odc-service/src/main/java/com/oceanbase/odc/service/session/interceptor/SqlCheckInterceptor.java
yhilmare: done

server/odc-core/src/main/java/com/oceanbase/odc/core/shared/constant/ErrorCodes.java
```diff
@@ -262,6 +262,7 @@ public enum ErrorCodes implements ErrorCode {
      * DB Object
      */
     DBObjectMetadataMayNotAccurate,
+    FailedQueryDBVersion,
```
yizhouxw, line 265: QueryDBVersionFailed

server/odc-core/src/main/java/com/oceanbase/odc/core/shared/constant/AuditEventType.java
```diff
@@ -74,6 +74,8 @@ public enum AuditEventType implements Translatable {
 
     PROJECT_MANAGEMENT,
 
+    SQL_SECURITY_RULES_MANAGEMENT
```
smallsheeeep, line 77: We are more inclined to use a single term such as `SQL_SECURITY_RULE_MANAGEMENT`

server/odc-core/src/main/resources/i18n/BusinessMessages.properties
```diff
@@ -279,6 +280,7 @@ com.oceanbase.odc.AuditEventType.UNKNOWN_TASK_TYPE=Unknown task type
 com.oceanbase.odc.AuditEventType.DATABASE_MANAGEMENT=Database management
 com.oceanbase.odc.AuditEventType.DATASOURCE_MANAGEMENT=Datasource management
 com.oceanbase.odc.AuditEventType.PROJECT_MANAGEMENT=Project management
+com.oceanbase.odc.AuditEventType.SQL_SECURITY_RULES_MANAGEMENT=SQL security rule management
```
smallsheeeep, line 283: Same as above

yizhouxw (approved):
lgtm

PR #616, opened 2023-10-25
enhancement: add log printing and avoid creating connect session when syncing datasource 

### LINKED ISSUE #613: [Enhancement]: do not create a session while synchronizing databases

1. Enhance the logic of acquiring locks when synchronizing data sources to avoid locks that cannot be released due to exceptions being thrown.
```Java
// src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
    public Boolean internalSyncDataSourceSchemas(@NonNull Long dataSourceId) throws InterruptedException {
        Lock lock = jdbcLockRegistry.obtain(getLockKey(dataSourceId));
        if (!lock.tryLock(3, TimeUnit.SECONDS)) {
            throw new ConflictException(ErrorCodes.ResourceModifying, "Can not acquire jdbc lock");
        }
        // the lock will not be released if the code as follow throw an exception
        ConnectionConfig connection = connectionService.getForConnectionSkipPermissionCheck(dataSourceId);
        horizontalDataPermissionValidator.checkCurrentOrganization(connection);
        try {
            organizationService.get(connection.getOrganizationId()).ifPresent(organization -> {
                if (organization.getType() == OrganizationType.INDIVIDUAL) {
                    syncIndividualDataSources(connection);
                } else {
                    syncTeamDataSources(connection);
                }
            });
            return true;
        } catch (Exception ex) {
            log.info("sync database failed, dataSourceId={}, error message={}", dataSourceId,
                    ex.getLocalizedMessage());
            return false;
        } finally {
            lock.unlock();
        }
    }
```
2. There is no need to create a complete session when synchronizing data sources, which will cause a lot of resource consumption and excessive log printing.

### Why is this needed?

Improve code robustness

yhilmare:
#### What type of PR is this?

type-enhancement

1. avoid creating `ConnectSession` when syncing datasource and create a schema.
2. cherry-pick 9f8c81f6 from dev/4.2.x into dev/4.2.2
3. add log printing when failed to sync datasource

Fixes #613 

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionSessionHistoryService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionSessionHistoryService.java
@@ -73,7 +73,7 @@ public class ConnectionSessionHistoryService {
     @Transactional(rollbackFor = Exception.class)
     public void updateOrInsert(Long connectionId, Long userId, Date lastAccessTime) {
         connectionHistoryDAO.updateOrInsert(ConnectionHistoryEntity.of(connectionId, userId, lastAccessTime));
-        log.info("update or insert connection history successfully, connectionId={}", connectionId);
+        log.debug("update or insert connection history successfully, connectionId={}", connectionId);
     }
 
     public List<ConnectionHistoryEntity> listAll() {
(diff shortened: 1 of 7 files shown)


yizhouxw (approved):
nice enhancement!

PR #620, opened 2023-10-25
fix(data-transfer): fix wrong data objects and schema objects

LuckyPickleZZ:
#### What type of PR is this?
type-bug

It's a coding error. Data objects and schema objects would be set into wrong positions.


yhilmare:
which issue does this pr fix?

yhilmare (approved):
lgtm

PR #626, opened 2023-10-25
feat(dlm):support configuring limiter

guowl3:
#### What type of PR is this?
type-feature

The current task current limit configuration can only be modified in the metadb, which is not user-friendly. Now users are allowed to configure it on a white screen.

Fixes #

```docs

```


server/odc-core/src/main/resources/i18n/BusinessMessages.properties
```diff
@@ -54,6 +54,7 @@ com.oceanbase.odc.ResourceType.ODC_SHADOWTABLE_COMPARING_TASK=ShadowTable Compar
 com.oceanbase.odc.ResourceType.ODC_SCHEDULE=Schedule
 com.oceanbase.odc.ResourceType.ODC_SCHEDULE_TRIGGER=Schedule Trigger
 com.oceanbase.odc.ResourceType.ODC_SCHEDULE_TASK=Schedule Task
+com.oceanbase.odc.ResourceType.ODC_DLM_LIMITER_CONFIG=Limiter Config
```
yizhouxw, line 57: Rate Limiting Configuration ?

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/ScheduleController.java
```diff
@@ -91,6 +96,12 @@ public ListResponse<String> getDownloadUrl(@PathVariable Long id, @RequestBody L
         return Responses.list(scheduleService.getAsyncDownloadUrl(id, objectId));
     }
 
+    @RequestMapping(value = "/{id:[\\d]+}/jobs/dlm/limiterConfig", method = RequestMethod.PUT)
+    public SuccessResponse<DlmLimiterConfig> updateLimiterConfig(@PathVariable Long id,
```
yizhouxw, line 100: no sure the scope of rate limiting, it is schedule level or job level ?

if schedule level, url may `/schedules/{scheduleIdd}/dlmRateLimitConfiguration`

server/odc-service/src/main/java/com/oceanbase/odc/service/dlm/model/DataDeleteParameters.java
```diff
@@ -37,4 +37,6 @@ public class DataDeleteParameters implements TaskParameters {
 
     private List<DataArchiveTableConfig> tables;
 
+    private DlmLimiterConfig limiterConfig;
```
yizhouxw, line 40: limiterConfig -->  ratelimit

yizhouxw (commented):
the class DlmLimiterConfig may not clear for API model.

e.g. what is the postfix `limit` means,  batch limit or rate limit.

do we have plan to offer more configurable limit ability?

crrenttly all limit abilit belong to rate limit, it will be okay to use RateLimitConfiguration.

or, may DlmLimitConfiguration, with fields:
- max-rows-per-second
- max-size-per-second
- batch-rows

yizhouxw (approved):
lgtm

PR #629, opened 2023-10-26
feat(data-security): add data type unit into response

### LINKED ISSUE #46: [Feature]: Data masking supports columns in database view
### Describe your use case

ODC provide data masking for database tables since v4.2.0, but the database views are not taken into consideration.
In the current situation, sensitive data can be access  illegally through creating a view and select data from this view.

1. So, I propose to incorporate views into data desensitization control.

1. Allow setting columns in views as sensitive columns；
2. When extracting columns from SQL, just make views into consideration.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

smallsheeeep:
#### What type of PR is this?
type-feature
module-data security

When manually add sensitive columns, we will show column type icon as followings:
image
Different databse type and version has different column type, so we need response data type unit to front-end.

Fixes #46


yhilmare (approved):
lgtm

PR #649, opened 2023-10-27
fix(migration): rule metadata migration will be triggered every time the ODC server starts up

### LINKED ISSUE #574: [Bug]: R42017RuleMetadataMigrate repeatly execute every start up
### ODC version

4.2.0/4.2.1

### OB version

not related

### What happened?

[2023-10-19 13:25:47.500][main][,,][INFO][com.oceanbase.odc.migrate.jdbc.common.R42017RuleMetadataMigrate][90]: remove regulation rules successfully, total removed rule count=49
[2023-10-19 13:26:19.313][main][,,][INFO][com.oceanbase.odc.migrate.jdbc.common.R42017RuleMetadataMigrate][95]: add new regulation rules successfully, total added rule count=49
[2023-10-19 13:26:19.314][main][,,][INFO][com.oceanbase.odc.core.migrate.Migrates][222]: migrate REPEATABLE JDBC 4.2.0.17 init regulation rule metadata done, result=true``

R42017RuleMetadataMigrate should be executed one time  exactly,
we may use ODC version as the checksum for this Migration script.

restart odc server, check the odc.log

_No response_

### Cloud

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug
module-migration

Rule metadata will be deleted and inserted once every time the ODC server starts up, while we expect that if there is no change between records in `regulation-rule-metadata.yaml` and in the metadb, we do not update the metadata.

The issue is that we only override the `equals()` and `hashCode()` method in `MetadataEntity` but do not override them in `PropertyMetadataEntity`.

Fixes #574 

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/metadb/regulation/ruleset/MetadataEntity.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/metadb/regulation/ruleset/MetadataEntity.java
@@ -99,20 +99,24 @@ public class MetadataEntity {
             return false;
         }
         MetadataEntity that = (MetadataEntity) obj;
-        return Objects.equals(this.getId(), that.getId()) && Objects.equals(this.getBuiltIn(), that.getBuiltIn())
+        return Objects.equals(this.getBuiltIn(), that.getBuiltIn())
                 && Objects.equals(this.getName(), that.getName())
                 && Objects.equals(this.getDescription(), that.getDescription())
                 && Objects.equals(this.getType(), that.getType())
-                && ((this.getLabels() != null && that.getLabels() != null
-                        && CollectionUtils.isEqualCollection(this.getLabels(), that.getLabels()))
+                && ((CollectionUtils.isEmpty(this.getLabels()) && CollectionUtils.isEmpty(that.getLabels())) ||
+                        (this.getLabels() != null && that.getLabels() != null
+                                && CollectionUtils.isEqualCollection(this.getLabels(), that.getLabels()))
                         || (this.getLabels() == null && that.getLabels() == null))
-                && ((this.getPropertyMetadatas() != null && that.getPropertyMetadatas() != null
-                        && CollectionUtils.isEqualCollection(this.getPropertyMetadatas(), that.getPropertyMetadatas()))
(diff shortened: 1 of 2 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/metadb/regulation/ruleset/MetadataEntity.java
```diff
@@ -99,20 +99,24 @@ public boolean equals(Object obj) {
             return false;
         }
         MetadataEntity that = (MetadataEntity) obj;
-        return Objects.equals(this.getId(), that.getId()) && Objects.equals(this.getBuiltIn(), that.getBuiltIn())
```
yhilmare, line 102: why don't we use `@EqualAndHashCode`, this class looks like a normal pojo class

MarkPotato777: `MetadataEntity` contains a list of `PropertyMetadataEntity`, and what we need here is that if two lists of  `PropertyMetadataEntity` are the same list regardless of the order of the elements, then the two `MetadataEntity` entities are equal. I use CollectionUtils.isEqualCollection in the `equals` method to judge if these two `PropertyMetadataEntity` lists are equal. I doubt if `@EqualAndHashCode` could handle this scene.

yhilmare: can we use `Set< PropertyMetadataEntity>` to regardless the sequence of the `PropertyMetadataEntity`?

yhilmare (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/metadb/regulation/ruleset/MetadataEntity.java
MarkPotato777: Yes, using `Set` instead of `List` is a better choice here, but we need to change the `PropertyMetadataEntity`'s `defaultValues` and `candidates` from a list to set as well, and this modification may cause some other changes to the original code, so I prefer not to make this modification here.

PR #652, opened 2023-10-30
fix(sql-check): fix syntax error check rule can not be disabled

### LINKED ISSUE #639: [Bug]: The enable button for the syntax error check item in the security specification does not take effect
### ODC version

ODC 4.2.2

### OB version

Any version of OceanBase.

### What happened?

When the improvement level is required for approval, the enable button does not take effect for syntax error check item.
image
image

the enable button should work

1.Modify the improvement level to require approval
2.Turn off enable button
3.Execute sql statements with syntax errors

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

syntax error sql-check rule can not be disabled, which is not expected.

The root cause of it is that the syntax error rule is hard coded in odc which can not be disabled.

I create the `SyntaxErrorExists` rule to check if user's statement is illegal. This is a normal sql-check rule, so that it can be disabled.

Fixes #639 

```docs

```


smallsheeeep (approved):
LGTM

PR #653, opened 2023-10-30
fix(data-transfer): no package body 

### LINKED ISSUE #482: [Bug]: Package export has no package body
### ODC version

oceanbase4.2.1

### OB version

Ob2277

### What happened?

Package export has no package body
image

**

**

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

When getting export objects, the DBSchamaAccessor#listPackage method will only return packages without body.
So I add a method `listPackageBodies` to list package bodies.

Fixes #482 

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/datatransfer/DBObjectNameAccessor.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/datatransfer/DBObjectNameAccessor.java
@@ -161,9 +161,9 @@ public class DBObjectNameAccessor implements AutoCloseable {
         if (session.getDialectType().isMysql()) {
             return Collections.emptySet();
         }
-        return accessor.listPackages(schema).stream()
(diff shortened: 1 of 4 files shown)


yhilmare (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/datatransfer/DBObjectNameAccessor.java
```diff
@@ -161,9 +161,9 @@ public Set<String> getPackageBodyNames() {
         if (session.getDialectType().isMysql()) {
             return Collections.emptySet();
         }
-        return accessor.listPackages(schema).stream()
```
yhilmare, line 164: why change this?

LuckyPickleZZ: Because accessor#listPackages can only obtain packages and cannot obtain package bodiesBecause accessor#listPackages can only obtain packages and cannot obtain package bodies

PR #659, opened 2023-10-30
fix(database): use datasource's environment as database's environment to prevent data inconsistency 

### LINKED ISSUE #575: [Bug]: [Team space] SQL workspace group by datasource occur unexpected exception
### ODC version

4.2.0/4.2.1

### OB version

not related

### What happened?

GET http://127.0.0.1:8989/api/v2/collaboration/projects/databases/stats?currentOrganizationId=10000&containsUnassigned=false

error response
```
{
  "code": "Unexpected",
  "durationMillis": 2,
  "error": {
    "code": "Unexpected",
    "details": [
      {
        "code": "Unexpected",
        "message": "Unhandled exception, type=NullPointerException, message=null",
        "target": null
      }
    ],
    "message": "预期外异常，错误详情：Unhandled exception, type=NullPointerException, message=null"
  },
  "httpStatus": "INTERNAL_SERVER_ERROR",
  "message": "预期外异常，错误详情：Unhandled exception, type=NullPointerException, message=null",
  "successful": false,
  "timestamp": 1697804406.814000000
}
```

related log

```
[2023-10-20 20:17:33.310][http-nio-8989-exec-2][0b6b4866792947fd,,O1QVROSASQUM38L9VQ4ME][WARN][com.oceanbase.odc.server.web.controller.RestExceptionHandler][307]: Unhandled NullPointerException:
java.lang.NullPointerException: null
        at com.oceanbase.odc.service.connection.database.DatabaseService.lambda$statsConnectionConfig$3(DatabaseService.java:244) ~[odc-service-4.2.2-SNAPSHOT.jar!/:?]
        at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193) ~[?:1.8.0_382]
        at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:175) ~[?:1.8.0_382]
        at java.util.ArrayList$Itr.forEachRemaining(ArrayList.java:901) ~[?:1.8.0_382]
        at java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801) ~[?:1.8.0_382]
        at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:482) ~[?:1.8.0_382]
        at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472) ~[?:1.8.0_382]
        at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708) ~[?:1.8.0_382]
        at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234) ~[?:1.8.0_382]
        at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:566) ~[?:1.8.0_382]
        at com.oceanbase.odc.service.connection.database.DatabaseService.statsConnectionConfig(DatabaseService.java:255) ~[odc-service-4.2.2-SNAPSHOT.jar!/:?]
        at com.oceanbase.odc.service.connection.database.DatabaseService$$FastClassBySpringCGLIB$$8febfe72.invoke(<generated>) ~[odc-service-4.2.2-SNAPSHOT.jar!/:?]
```

The default value for connect_connection.environment_id is -1, while the built-in environment IDs are always >0, causing the stats code to not find a matching environment object and resulting in NPE (NullPointerException). 

The specific reason is that the migration of the environment_id in the connect_connection table occurs during the TeamOrganizationMigrator process, which is triggered when the first user of the current organization logs in. 

On the other hand, the synchronization of connect_database is triggered by a scheduled task. If syncDatabase happens before TeamOrganizationMigrator, then the value of connect_database.environment_id will be -1. 

In this case, is it possible to not maintain the environment_id in the database table, since the environment is actually maintained in the connect_connection table? Maintaining it in two tables may cause data inconsistency. To ensure data consistency, locking needs to be added to the places where connect_connection.environment_id and connect_database.environment_id fields are inserted/updated.

image

this issue also affect other feature not work, get database API result into NotFoundException
image

related log
```
[2023-10-22 12:56:30.032][http-nio-8989-exec-10][ac4660ca15bb4872,,U4XBGBD4UH9V6SLSKHD73][INFO][com.oceanbase.odc.server.web.controller.RestExceptionHandler][375]: ODC_FAILED_REQUEST, httpMethod=GET, fullURL=http://obodc-trial.alibaba.net:8989/api/v2/database/databases/94?currentOrganizationId=10000&ignoreError=true, remoteHost=30.236.2.17, clientAddress=30.236.2.17,  handler=com.oceanbase.odc.server.web.controller.v2.DataBaseController#getDatabase[1 args], exceptionType=NotFoundException, message=ODC_ENVIRONMENT not found by id=-1, rootReason= 	at com.oceanbase.odc.service.collaboration.environment.EnvironmentService.detailSkipPermissionCheck(EnvironmentService.java:102) 	at com.oceanbase.odc.service.collaboration.environment.EnvironmentService.innerDetailWithoutPermissionCheck(EnvironmentService.java:142) 	at java.util.Optional.orElseThrow(Optional.java:290) 	at com.oceanbase.odc.service.collaboration.environment.EnvironmentService.lambda$innerDetailWithoutPermissionCheck$2(EnvironmentService.java:142) com.oceanbase.odc.core.shared.exception.NotFoundException: id 为 -1 的 环境 不存在, perfLevel=P3, response=ErrorResponse(error=Error(code=NotFound, message=id 为 -1 的 环境 不存在, details=[Error.Detail(code=NotFound, target=null, message=ODC_ENVIRONMENT not found by id=-1), Error.Detail(code=NotFoundException, target=null, message=ODC_ENVIRONMENT not found by id=-1)]), code=NotFound, message=id 为 -1 的 环境 不存在)
```

no concurrent conflict for database metadata

upgrade odc from version 4.1.3 to 4.2.1， with several public connnection config already existed in version 4.1.3.

login the first account after database sync schedule.

_No response_

### Cloud

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug

Both `connect_connection` and `connect_database` maintain the `environemnt_id` which may cause data inconsistencies.
For example,  the migration of the `environment_id` in the `connect_connection` table occurs during the `TeamOrganizationMigrator` process, which is triggered when the first user of the current organization logs in. On the other hand, the synchronization of `connect_database` is triggered by a scheduled task. If `syncDatabase` happens before `TeamOrganizationMigrator`, then the value of `connect_database.environment_id` will be -1.

The solution is that we no longer rely on the `environment_id` of the `connect_database`, but set the environment of the database to the environment of the data source in the conversion layer of entity to model. Considering compatibility, we do not change any schema or migrate any data here, and the caller of the DatabaseService doesn't need to care about this change, because all changes happen in-ground.

Fixes #575 

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
@@ -235,17 +235,7 @@ public class DatabaseService {
             return Collections.emptyList();
         }
         return databases.stream().filter(database -> Objects.nonNull(database.getDataSource()))
-                .map(database -> {
-                    ConnectionConfig connection = database.getDataSource();
-                    Environment environment = database.getEnvironment();
-                    if (Objects.isNull(environment)) {
-                        log.warn("database environment is null, databaseId={}", database.getId());
-                    } else {
-                        connection.setEnvironmentStyle(environment.getStyle());
-                        connection.setEnvironmentName(environment.getName());
-                    }
-                    return connection;
-                })
+                .map(Database::getDataSource)
                 .collect(Collectors.collectingAndThen(
                         Collectors.toCollection(() -> new TreeSet<>(Comparator.comparing(ConnectionConfig::getId))),
                         ArrayList::new));
@@ -318,14 +308,6 @@ public class DatabaseService {
     public Set<Database> listDatabaseByNames(@NotEmpty Collection<String> names) {
         return databaseRepository.findByNameIn(names).stream().map(databaseMapper::entityToModel)
                 .collect(Collectors.toSet());
-    }
-
-    @SkipAuthorize("internal usage")
-    @Transactional(rollbackFor = Exception.class)
-    public void updateEnvironmentByDataSourceId(@NonNull Long dataSourceId, @NonNull Long environmentId) {
-        List<DatabaseEntity> databases = databaseRepository.findByConnectionId(dataSourceId);
-        databases.forEach(database -> database.setEnvironmentId(environmentId));
-        databaseRepository.saveAll(databases);
     }
 
     @Transactional(rollbackFor = Exception.class)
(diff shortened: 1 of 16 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
```diff
@@ -620,7 +599,7 @@ private Database entityToModel(DatabaseEntity entity) {
             model.setProject(projectService.detail(entity.getProjectId()));
         }
         model.setDataSource(connectionService.getForConnectionSkipPermissionCheck(entity.getConnectionId()));
-        model.setEnvironment(environmentService.detailSkipPermissionCheck(entity.getEnvironmentId()));
+        model.setEnvironment(environmentService.detailSkipPermissionCheck(model.getDataSource().getEnvironmentId()));
```
yizhouxw, line 602: does. environmentService.detailSkipPermissionCheck requires MetaDB IO , seems performance risk while entityToModel ?

MarkPotato777, line 602: It looks like there's no way we can save this IO cost. entityToModel is only used in the DatabaseService#detail and DatabaseService#create, which I think is acceptable and actually, in the real scene, the create database and the update database APIs' RT look like normal.
On the other hand, in the situation of batch operation, we use another method entitiesToModels and this method loads other POJOs in one time rather than in iteration one by one, so there is no additional time cost as well.

yizhouxw (approved):
lgtm

PR #661, opened 2023-10-30
feat(sql-splitter): support SqlCommentProcessor to split sql by stream

LuckyPickleZZ:
#### What type of PR is this?
type-feature

The `SqlCommentProcessor` needs to read SQL text into memory at once and then split it. When the SQL file is large, OOM exceptions may occur. 
This PR supports reading and splitting SQL through a `InputStream`.

none, just an optimization

The UT has passed.

```docs

```

### DIFF
--- a/server/odc-core/src/main/java/com/oceanbase/odc/core/sql/split/SqlCommentProcessor.java
+++ b/server/odc-core/src/main/java/com/oceanbase/odc/core/sql/split/SqlCommentProcessor.java
@@ -96,8 +96,11 @@ public class SqlCommentProcessor {
 
     public SqlCommentProcessor() {}
 
-    public SqlIterator iterator(InputStream in) {
-        return new SqlIterator(in);
+    public static SqlLine getSqlLines(InputStream in, DialectType dialectType,
+            boolean preserveFormat,
+            boolean preserveSingleComments,
+            boolean preserveMultiComments) {
+        return new SqlLine(in, dialectType, preserveFormat, preserveSingleComments, preserveMultiComments);
     }
 
     public static List<String> removeSqlComments(String originalSql,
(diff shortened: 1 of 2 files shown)


server/odc-core/src/main/java/com/oceanbase/odc/core/sql/split/SqlCommentProcessor.java
```diff
@@ -599,4 +609,71 @@ public int getValue() {
         }
     }
 
+    public class SqlIterator implements Iterator<String>, AutoCloseable {
```
yhilmare, line 612: don't do this, make `SqlCommentProcessor` as an util.

the internal status of `SqlCommentProcessor` is much too complex that should not care about it.

from my point of view, `addLineMySQL` and `addLineOracle` is enough

yhilmare, line 612: further more, the name of this class may be `SqlLine`? user will get a line from every iteration.

LuckyPickleZZ: done. Used a static method `getSqlLines` to return inner static class `SqlLine`. The object of `SqlLine` will create a new `SqlCommentProcessor` for usage.

LuckyPickleZZ: done. Renamed to `SqlLine`

server/odc-core/src/main/java/com/oceanbase/odc/core/sql/split/SqlCommentProcessor.java
```diff
@@ -599,4 +612,77 @@ public int getValue() {
         }
     }
 
+    public static class SqlLine implements Iterator<String>, AutoCloseable {
```
yizhouxw, line 615: may `SqlStatementIterator`

yizhouxw (approved):
lgtm

PR #668, opened 2023-10-31
fix(dlm): submit task got condition not supported error while condition contains subquery

### LINKED ISSUE #623: [Bug]: Subquery does not support archiving
### ODC version

ODC422

### OB version

OCEANBASE410

### What happened?

Subquery does not support archiving
image

support

`LENGTH(col11) < 10
AND col11 NOT IN (
    SELECT col11
    FROM nopart_pri_normal
);
`

_No response_

### Cloud

_No response_

guowl3:
#### What type of PR is this?
type-bug

Set default schema before create connection session.

Fixes #623

```docs

```


yizhouxw:
for bugfix PR, the title describe the problem, not  root cause, neither  solution.

e.g.
fix(dlm): submit task got condition not supported error while condition contains subquery

yhilmare (approved):
lgtm

PR #677, opened 2023-10-31
fix(pl-debug): fix the problem of hang when debugging on arm architecture ob

### LINKED ISSUE #341: [Bug]: odc will hang when debugging PL objects on the arm architecture observer.
### ODC version

any odc

### OB version

OceanBase 4.2.0.0 (r100000062023080417-9307fa0001861c22d5b34adaa72a056c493883c1) (Built Aug  4 2023 18:10:08)

### What happened?

start debugging a procedure which is very simple:
```SQL
create or replace procedure test_proc(p in varchar2)
as
begin
    dbms_output.put_line('Hello,world');
end;
```
odc will hang up when you click "start debug"

odc will not hang up

follow "What happened"

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

odc will hang up on arm architecture ob when debugging. this problem is caused by observer, but odc can do somethings to prevent it. 

you can get more infos from the related issue.

Fixes #341 

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/pldebug/session/DebuggeeSession.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/pldebug/session/DebuggeeSession.java
@@ -85,10 +85,10 @@ public class DebuggeeSession extends AbstractDebugSession {
                     debugId = resultSet.getString(1);
                 }
             }
-            // 打开调试开关
-            stmt.execute("call dbms_debug.debug_on();");
             // 打开pl的日志输出
             enableDbmsOutput(stmt);
+            // 打开调试开关
+            stmt.execute("call dbms_debug.debug_on();");
         } catch (SQLSyntaxErrorException e) {
             if (Objects.equals(e.getErrorCode(), ERROR_CODE)
                     && StringUtils.contains(e.getMessage(), PL_DEBUGGING_ERROR_CODE)) {


yhilmare:
odc will hang on `call dbms_output.enable(100000);` which will not happened on x86 architecture ob and most arm architecture ob.

the root cause of it may be the bad way of packaging for arm architecture OceanBase. what we can do is that execute `call dbms_output.enable(100000);` firstly and then execute `call dbms_debug.debug_on();`.

smallsheeeep (approved):
LGTM

server/odc-service/src/main/java/com/oceanbase/odc/service/pldebug/session/DebuggeeSession.java
```diff
@@ -85,10 +85,10 @@ public DebuggeeSession(ConnectionSession connectionSession, ThreadPoolExecutor d
                     debugId = resultSet.getString(1);
                 }
             }
-            // 打开调试开关
-            stmt.execute("call dbms_debug.debug_on();");
             // 打开pl的日志输出
             enableDbmsOutput(stmt);
+            // 打开调试开关
```
smallsheeeep, line 90: Why the problem will be solved by adjusting the execution sequence?

yhilmare: if you execute `dbms_debug.debug_on` which means that you will hang up in the next sql you executed. ob-server will consider that the next sql you executed is the target that you want to debug.

so that I swap them to avoid the situation I mentioned above.

you may be interested in the reason why does these codes work if any sql executed after `dbms_debug.debug_on` will hang up, I just want to say I don't know either😂

PR #685, opened 2023-11-01
fix(result-set-export): get wrong filename for result-set export task on cloud

LuckyPickleZZ:
#### What type of PR is this?
type-bug

There is a code bug which will lead to get wrong filename for result-set export task. So we use user-configured filename to upload result instead of origin filename.

none


yhilmare (approved):
lgtm

PR #691, opened 2023-11-01
fix(datasource): make 'socketTimeout' and 'connectTimeout' settings work for backend datasource

yhilmare:
#### What type of PR is this?

type-bug

odc's backend datasource will be set a default socket timeout which defined in system config. this socket timeout can not be changed even we set jdbc url parameter.

There is a bug for Druid that it can only set socketTimeout and connectTimeout by jdbc url parameter for MySQL datasource.  That is to say that we can not set these two properties for OceanBase datasource.

We set these two properties to a default value(60s), by the way, these to properties can not be set to zero, the druid will give them a default value(10s) if we do that. 

Next, we parse these two properties from the jdbc url, if we find any one of them, we will reset it.

Fixes #

```docs

```


LuckyPickleZZ (approved):
lgtm

PR #699, opened 2023-11-01
fix(sql-parser): failed to report syntax error if the input's any prefix is grammatical

### LINKED ISSUE #693: [Bug]: SQL statements executed within obmysql and native mysql cannot be checked for syntax errors
### ODC version

odc422

### OB version

*

### What happened?

image
image
image

Before executing SQL, SQL checks for errors and jumps to database changes

1. Turn on the switch for syntax errors
2. Entering incorrect SQL statements in the SQL window

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

the parser will not report any syntax error if the input's any prefix is grammatical. 

eg.
```SQL
select * from test asdasdasd asdasdas asdasdasd asd;
```

the sql above has syntax error obviously but the parser will consider that this is a sql without any syntax error cause that the prefix of this sql `select * from test` is grammatical.

the root cause of this issue is that the parser does not parse a sql from the beginning to the end. The parser will return when it get a grammatical statement.

you can get more infos from the issue #693 

Fixes #693 

```docs

```

### DIFF
--- a/libs/db-browser/pom.xml
+++ b/libs/db-browser/pom.xml
@@ -83,7 +83,7 @@
         <commons-collections4.version>4.4</commons-collections4.version>
         <log4j.version>2.18.0</log4j.version>
         <fastjson.version>1.2.83</fastjson.version>
-        <ob-sql-parser.version>1.1.2</ob-sql-parser.version>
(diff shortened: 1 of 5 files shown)


smallsheeeep (approved):
LGTM

PR #707, opened 2023-11-03
fix(dlm):wrong order status when task is rollback

### LINKED ISSUE #704: [Bug]: When executing a DLM task periodically, the status changes to Completed, resulting in no disable button.
### ODC version

ODC422

### OB version

independent

### What happened?

When executing a DLM task periodically, the status changes to Completed, resulting in no disable button.
image

normally

Create periodic execution tasks

_No response_

### Cloud

_No response_

guowl3:
#### What type of PR is this?
type-bug

When a sub-task of a periodic work order is rolled back, the status of the work order will not be affected.

Fixes #704

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/quartz/OdcJobListener.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/quartz/OdcJobListener.java
@@ -17,6 +17,7 @@ package com.oceanbase.odc.service.quartz;
 
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 
 import org.quartz.JobExecutionContext;
 import org.quartz.JobExecutionException;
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/quartz/OdcJobListener.java
```diff
@@ -129,6 +130,12 @@ public void jobExecutionVetoed(JobExecutionContext context) {
     public void jobWasExecuted(JobExecutionContext context, JobExecutionException jobException) {
         JobKey key = context.getJobDetail().getKey();
         List<? extends Trigger> jobTriggers;
+        Optional<ScheduleEntity> scheduleEntityOptional =
+                scheduleRepository.findById(ScheduleTaskUtils.getScheduleId(context));
+        if (!scheduleEntityOptional.isPresent()
```
yizhouxw, line 135: a little hard to understand..

may split into 2 steps

if(!scheduleEntityOptional.isPresent() {
    // if unespected, log.warn here 
    return;
}

isRollbackTask=!key.getGroup().equals(scheduleEntityOptional.get().getJobType().name())
if (isRollbackTask) {
 return;
}

guowl3: simplified the method and the key of the rollback task  not used to find the trigger now.

yizhouxw (approved):
lgtm

PR #722, opened 2023-11-06
fix(web): response header content-type would be application/xml while using RestTemplate

MarkPotato777:
#### What type of PR is this?

type-bug
module-web

The [PR](https://github.com/oceanbase/odc/pull/338) involves a problem that the response header `content-type` would be `application/xml` while using RestTemplate. The root cause is that after importing the dependency `jackson-dataformat-xml`, Spring will automatically register a `MappingJackson2XmlHttpMessageConverter` whose order is higher than `MappingJackson2HttpMessageConverter`. If the request header Accept is like `'text/plain, */*'`, Spring will firstly use `MappingJackson2XmlHttpMessageConverter` to convert the response to an xml format which would cause the issue.

This PR swap the order of `MappingJackson2HttpMessageConverter` and `MappingJackson2XmlHttpMessageConverter` to make `MappingJackson2XmlHttpMessageConverter` work before `MappingJackson2XmlHttpMessageConverter`.

Fixes #

```docs

```

### DIFF
--- a/server/odc-server/src/main/java/com/oceanbase/odc/server/config/MvcConfiguration.java
+++ b/server/odc-server/src/main/java/com/oceanbase/odc/server/config/MvcConfiguration.java
@@ -33,6 +33,7 @@ import org.springframework.format.datetime.standard.DateTimeFormatterRegistrar;
 import org.springframework.http.MediaType;
 import org.springframework.http.converter.HttpMessageConverter;
 import org.springframework.http.converter.json.MappingJackson2HttpMessageConverter;
+import org.springframework.http.converter.xml.MappingJackson2XmlHttpMessageConverter;
 import org.springframework.web.context.request.RequestContextListener;
 import org.springframework.web.filter.RequestContextFilter;
 import org.springframework.web.filter.ShallowEtagHeaderFilter;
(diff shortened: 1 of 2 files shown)


server/odc-server/src/main/java/com/oceanbase/odc/server/config/MvcConfiguration.java
```diff
@@ -115,15 +116,28 @@ public void addFormatters(FormatterRegistry registry) {
...
                 jacksonConverter.setObjectMapper(objectMapper);
                 log.info("update objectMapper for jackson http message converter");
-                return;
             }
         }
-        throw new RuntimeException("no jackson http message converter found!");
+        if (jacksonConverterIndex == -1) {
+            throw new RuntimeException("no jackson http message converter found!");
+        }
+        if (xmlConverterIndex != -1) {
```
smallsheeeep, line 136: May use  `if (xmlConverterIndex != -1 && xmlConverterIndex < jacksonConverterIndex)` in case that the xmlConverter is already after jacksonConverter?

smallsheeeep (commented):
Good job!

server/odc-server/src/main/java/com/oceanbase/odc/server/config/MvcConfiguration.java
MarkPotato777: nice suggestion! Updated.

smallsheeeep (approved):
LGTM

PR #726, opened 2023-11-07
feat(osc): reactor api get database about lock user required

krihy:
#### What type of PR is this?

type-feature
module-Online schema change
show database selected should lock db user when create osc task

Fixes #
old api is `/databases/{id:[\d]+}` is common api in font web, add a  alone api `/lockDatabaseUserRequired/{databaseId:[\d]+}` to query database lock user info,  cooperation succeed with front-end

```docs

```


yhilmare (approved):
lgtm

PR #735, opened 2023-11-08
Fix(dlm): data deletion failed after archiving the data

### LINKED ISSUE #734: [Bug]: DLM task source cleanup does not take effect
### ODC version

ODC422

### OB version

Oceanbase312

### What happened?

DLM task source cleanup does not take effect
image
image

can cleanhup

create DLM task 
and Check the cleanhup

_No response_

### Cloud

_No response_

guowl3:
#### What type of PR is this?
type-bug

Regenerate the ID for the rollback task and delete task to avoid task failure due to duplicate IDs

Fixes #734

```docs

```


yizhouxw (approved):
lgtm

PR #748, opened 2023-11-09
fix(sql-check): avoid reporting syntax error when sql is executed successfully

### LINKED ISSUE #746: [Bug]: sql-check report a syntax error when a sql is executed successfully
### ODC version

4.2.2

### OB version

not replated

### What happened?

1. switch on the 'syntax error' rule
截屏2023-11-09 11 37 24

2. type and execute the sql as follow:
```SQL
drop table if exists t1;
create table t1 (c1 text) ;
insert into t1 select  (randstr(1+zipf(1, 43250, random()), random())) from table(generator(1000));
```
3. as you can see, the database execute these sqls successfully but the sql-check gives a syntax error:
0F35EBB1-2ECC-40A5-B749-5986BBF75501

if a sql is executed by observer without any syntax error, the sql-check should not report syntax error.

follow 'what happened'

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

out parser can not behaves same with the observer's parser completely so that a sql may be executed completely but get a syntax error from sql-check in odc which will confuse user.

this pr fix this issue.

I have tested 2.2.77, 3.2.4 and 4.2.1 for both MySQL and Oracle mode and I asked kernel R&D classmates for help, and what I can confirm is that

```shell
mysql324: ErrorCode = 1064, SQLState = 42000
mysql2277: ErrorCode = 1064, SQLState = 42000 
mysql421: ErrorCode = 1064, SQLState = 42000
mysql57: ErrorCode = 1064, SQLState = 42000
oracle2277: ErrorCode = 900, SQLState = 42000
oracle324: ErrorCode = 900, SQLState = 42000 
oracle421: ErrorCode = 900, SQLState = 42000
```

Fixes #746 

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/session/factory/OBConsoleDataSourceFactory.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/session/factory/OBConsoleDataSourceFactory.java
@@ -194,11 +194,11 @@ public class OBConsoleDataSourceFactory implements CloneableDataSourceFactory {
             dataSource.setAutoCommit(autoCommit);
         }
         if (this.initConnection) {
-            dataSource.addInitializer(new SessionCreatedInitializer(connectionConfig));
             List<ConnectionInitializer> initializers = connectionExtensionPoint.getConnectionInitializers();
             if (!CollectionUtils.isEmpty(initializers)) {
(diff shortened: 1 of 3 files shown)


yizhouxw (approved):
lgtm

PR #752, opened 2023-11-09
enhancement(full-link-trace): optimize  display effect and functions of Full-Link-Trace

### LINKED ISSUE #322: [Feature]: The display effect and functions of Full-Link-Trace need to be optimized

The current display effect of the FullLinkTrace function is not very suitable for us to locate the problem. Moreover, it lacks search, sorting and other functions, making it inconvenient to use.
Plus, when observer's version is lower than 4.1, FullLinkTrace is not available. It may give some hints.

FullLinkTrace requires functions such as search and sorting. 
When this function is not available, prompt the reason, such as OB version lower than 4.1.

### Additional context

_No response_

LuckyPickleZZ:
#### What type of PR is this?

type-feature

1. In this PR, we added `withFullLinkTrace` and `traceEmptyReason` into api to indicate if this query has trace info . 
2. Added a interface `ThirdPartyTraceConverter` to convert from oceanbase trace span into json format map that can be parsed by a 3rd party tool, like Jaeger/Zipkin... 

Fixes #322

```docs

```


server/odc-core/src/main/java/com/oceanbase/odc/core/shared/constant/ErrorCodes.java
```diff
@@ -225,6 +225,8 @@ public enum ErrorCodes implements ErrorCode {
     ObGetExecuteDetailFailed,
     ObGetPlanExplainFailed,
     ObGetPlanExplainEmpty,
+    ObGetFullLinkTraceFailed,
+    FullLinkTraceNotAvailable,
```
yizhouxw, line 229: ObFullLinkTraceNotSupported 

not available seems service not working, 'not supported' will be an expected behavior.

server/odc-core/src/main/java/com/oceanbase/odc/core/sql/execute/model/SqlExecTime.java
```diff
@@ -33,5 +33,7 @@ public class SqlExecTime {
 
     // 链路信息
     private TraceSpan traceSpan;
+    private boolean withFullLinkTrace;
+    private String traceEmptyReason;
```
yizhouxw, line 37: traceEmptyReason  or emptyTraceReaon, which will be better?

server/odc-core/src/main/resources/i18n/ErrorMessages.properties
```diff
@@ -126,6 +126,8 @@ com.oceanbase.odc.ErrorCodes.ObCreatePlRunPackageFailed=Create PL run package fa
 com.oceanbase.odc.ErrorCodes.ObGetExecuteDetailFailed=Get execute info failed, details: {0}
 com.oceanbase.odc.ErrorCodes.ObGetPlanExplainFailed=Get plan explain failed, may invalid sql or DDL, details: {0}
 com.oceanbase.odc.ErrorCodes.ObGetPlanExplainEmpty=Cannot get plain explain, may invalid sql or DDL, or some unsupported command, details: {0}
+com.oceanbase.odc.ErrorCodes.ObGetFullLinkTraceFailed=Get full link trace failed, please check if ob_enable_show_trace is enabled
```
yizhouxw, line 129: please check if ob_enable_show_trace is enabled

-->

please check if session variable 'ob_enable_show_trace' is 'ON'

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v1/SqlDiagnoseController.java
```diff
@@ -75,4 +79,12 @@ public SuccessResponse<TraceSpan> getFullLinkTrace(@PathVariable String sid, @Re
         return Responses
                 .success(diagnoseService.getFullLinkTrace(sessionService.nullSafeGet(SidUtils.getSessionId(sid)), sql));
     }
+
+    @ApiOperation(value = "getFullLinkTraceJson")
+    @RequestMapping(value = "/getFullLinkTraceJson/{sid}", method = RequestMethod.POST)
```
yizhouxw, line 84: what is the difference between getFullLinkTrace and getFullLinkTraceJson ?
the  previous one return json format also.

server/odc-service/src/main/java/com/oceanbase/odc/service/diagnose/fulllinktrace/JaegerConverter.java
```diff
@@ -0,0 +1,150 @@
...
+import org.apache.commons.collections4.CollectionUtils;
+
+import com.google.gson.GsonBuilder;
+import com.oceanbase.odc.core.shared.model.TraceSpan;
+
+import lombok.Builder;
+import lombok.Getter;
+
+public class JaegerConverter implements ThirdPartyTraceConverter {
+    private static final String PROCESS_ID_PATTERN = "%s-%s-%s";
```
yizhouxw, line 40: PROCESS_ID_FORMAT

server/odc-service/src/main/java/com/oceanbase/odc/service/diagnose/fulllinktrace/JaegerConverter.java
```diff
@@ -0,0 +1,150 @@
...
+    private long toTimestampWithMicros(String timeStr) {
+        Instant instant = LocalDateTime.parse(timeStr, TraceSpan.TIMESTAMP_FORMATTER)
+                .atZone(ZoneId.systemDefault()).toInstant();
+        return instant.getEpochSecond() * 1000000 + instant.getNano() / 1000;
+    }
+
+    @Getter
+    @Builder
+    private static class Span {
+        private String traceID;
```
yizhouxw, line 104: here 'traceID', 'spanID', 'processID' use 'ID' postfix, other module use 'Id' instead.

what is the reason of diferent style ?

server/odc-service/src/main/java/com/oceanbase/odc/service/diagnose/fulllinktrace/JaegerConverter.java
```diff
@@ -0,0 +1,150 @@
...
+        List<Span> spans = new ArrayList<>();
+        Map<String, Process> processes = new HashMap<>();
+        dataMap.put("spans", spans);
+        dataMap.put("processes", processes);
+
+        dfs(spans, processes, span, null);
+
+        Map<String, Object> root = new HashMap<>();
+        root.put("data", Collections.singletonList(dataMap));
+        return new GsonBuilder().setPrettyPrinting().create().toJson(root);
```
yizhouxw, line 55: ODC use Jackson, avoid Gson if not neccesary

server/odc-service/src/main/java/com/oceanbase/odc/service/diagnose/fulllinktrace/JaegerConverter.java
LuckyPickleZZ: because Jaeger can only recognize this format rather than `Id` 😂

server/odc-service/src/main/java/com/oceanbase/odc/service/diagnose/fulllinktrace/JaegerConverter.java
LuckyPickleZZ: The reason for using Gson here is to beautify. Gson can output tree-shaped JSON files, while jackson can only be tiled.
Do I need to change this?

server/odc-service/src/main/java/com/oceanbase/odc/service/diagnose/fulllinktrace/JaegerConverter.java
LuckyPickleZZ: done .

server/odc-core/src/main/java/com/oceanbase/odc/core/shared/constant/ErrorCodes.java
LuckyPickleZZ: Very reasonable. Done.

server/odc-core/src/main/java/com/oceanbase/odc/core/sql/execute/model/SqlExecTime.java
LuckyPickleZZ: I prefer traceEmptyReason. The other seems like a boolean variable that indicates the 'trace reason' is empty.

server/odc-core/src/main/resources/i18n/ErrorMessages.properties
LuckyPickleZZ: yes, updated this

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v1/SqlDiagnoseController.java
LuckyPickleZZ: This api is a new interface to query json content, only for download json file. It is not the same api as the previous one

server/odc-service/src/main/java/com/oceanbase/odc/service/diagnose/fulllinktrace/JaegerConverter.java
LuckyPickleZZ: done, used Jackson instead of Gson

server/odc-core/src/main/java/com/oceanbase/odc/core/shared/constant/ErrorCodes.java
```diff
@@ -225,6 +225,8 @@ public enum ErrorCodes implements ErrorCode {
     ObGetExecuteDetailFailed,
     ObGetPlanExplainFailed,
     ObGetPlanExplainEmpty,
+    ObGetFullLinkTraceFailed,
+    FullLinkTraceNotSupported,
```
yizhouxw, line 229: miss `Ob` prefix

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v1/SqlDiagnoseController.java
yizhouxw: however we cannot distinguish from the API endpoint. 

if for download, here we may return a file id then front-end call file download API
- step1, getFulLinkTraceDownloadLink, return download link
- step, call download link, which is a cloud storage url in cloud env

if for query jager compatible, may share the same API wieh getFullLinkTraceJson, add an optional queryparameter `jagerCompatible=true`, then download logic is controlled by front-end

LuckyPickleZZ: changed this api into `/getFullLinkTraceDownloadUrl` which would return a download link ~

server/odc-core/src/main/java/com/oceanbase/odc/core/shared/constant/ErrorCodes.java
LuckyPickleZZ: fixed.

yizhouxw (approved):
lgtm

PR #755, opened 2023-11-09
feat(project): add two built-in project roles

### LINKED ISSUE #360: [Feature]: support more project roles

For now, ODC only supports 3 different project roles: owner, DBA, and developer. However, in actual scenarios, there may be more roles in a project, such as security administrators, data analysts, etc., and some tickets need more roles' review.

Support custom project roles

### Additional context

_No response_

MarkPotato777:
#### What type of PR is this?
type-feature
module-collaboration

Current project roles cannot really meet the requirement, so I added two built-in project roles: security administrator and participant. The security administrator could work as an approval node and manage the sensitive columns; the participant could only work as an approval node.

Additionally,  this PR allows the project owner to list **all** tickets of this project; for other project roles, they can the tickets created by them, pending their approval, and they have already approved of the project.

Fixes #360 

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/permissionapply/project/ApplyProjectParameter.java
```diff
@@ -69,10 +68,6 @@ public static class ApplyResourceRole implements Serializable {
          * Resource role id, refer to {@link ResourceRole#getId()}
          */
         private Long id;
-        /**
-         * Resource role name, refer to {@link ResourceRole#getRoleName()}
-         */
-        private ResourceRoleName name;
```
smallsheeeep, line 75: Please do not delete this field. The front-end need it to display.
BTW: here exists a bug, i will fix it soon...

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/FlowInstanceController.java
```diff
@@ -105,6 +105,7 @@ public PaginatedResponse<FlowInstanceDetailResp> listFlowInstances(
             @RequestParam(required = false, name = "endTime") Date endTime,
             @RequestParam(name = "createdByCurrentUser") Boolean createdByCurrentUser,
             @RequestParam(name = "approveByCurrentUser") Boolean approveByCurrentUser,
+            @RequestParam(required = false, name = "containsAll", defaultValue = "false") Boolean containsAll,
```
yizhouxw, line 108: if createdByCurrentUser/approveByCurrentUser not set, will it be contains All?

all query parameters can mapping to a filter, if not parameters, means there is not filter ?

MarkPotato777: Currently, the parameter `createdByCurrentUser/approveByCurrentUser` is required, so they will always be set. If `containsAll = true`, then the `containsAll`'s priority is higher than `createdByCurrentUser/approveByCurrentUser`'s; if `containsAll = false`, then the query depends on `createdByCurrentUser/approveByCurrentUser`.
But it seems a bit tricky...

yhilmare (commented):
lgtm

yhilmare (approved):
lgtm

PR #756, opened 2023-11-09
build: simplify arm docker build

yizhouxw:
#### What type of PR is this?

type-build

simplify arm docker build
1. same Dockerfile for both amd64 and arm64 arch
2. default argument for rpm build and docker build

now we can build arm docker locally with 2 steps
- step1 script/build_rpm.sh, with no argument ,  will set release number automatically
- step2 cd distribution/docker && ./build.sh build && ./build.sh save, with no argument , will calculate image tag automatically from odc version and cpu arch

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #760, opened 2023-11-10
fix(apply-project): Resource role name in DTO is modified but be trusted

### LINKED ISSUE #759: [Bug]: Resource role name in DTO should not be trusted
### ODC version

4.2.3

### OB version

null

### What happened?

Resource role name in DTO is trusted by ODC

Resource role name in DTO should not be trusted

Start a apply project task;
Catch the request and modify the resource role name field, and then submit this request.

_No response_

### Cloud

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
module-apply project

Resource role name in DTO could by modified but trusted by ODC. This PR fix it.
Fixes #759

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/permissionapply/project/ApplyProjectPreprocessor.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/permissionapply/project/ApplyProjectPreprocessor.java
@@ -66,6 +66,10 @@ public class ApplyProjectPreprocessor implements Preprocessor {
         List<ResourceRoleEntity> resourceRoleEntities = checkResourceRoleExist(
                 parameter.getResourceRoles().stream().map(ApplyResourceRole::getId).collect(Collectors.toList()));
         parameter.setUserId(authenticationFacade.currentUserId());
+        parameter.getProject().setName(projectEntity.getName());
+        for (int index = 0; index < parameter.getResourceRoles().size(); index++) {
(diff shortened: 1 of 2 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/permissionapply/project/ApplyProjectPreprocessor.java
```diff
@@ -66,6 +66,10 @@ public void process(CreateFlowInstanceReq req) {
         List<ResourceRoleEntity> resourceRoleEntities = checkResourceRoleExist(
                 parameter.getResourceRoles().stream().map(ApplyResourceRole::getId).collect(Collectors.toList()));
         parameter.setUserId(authenticationFacade.currentUserId());
+        parameter.getProject().setName(projectEntity.getName());
+        for (int index = 0; index < parameter.getResourceRoles().size(); index++) {
+            parameter.getResourceRoles().get(index).setName(resourceRoleEntities.get(index).getRoleName());
```
MarkPotato777, line 71: Why do you need the project role name?

smallsheeeep: The front-end should display it in task details. If we return it, the front-end will no need extra query.

MarkPotato777 (dismissed):
LGTM

yhilmare (approved):
lgtm

PR #764, opened 2023-11-10
fix(sql-execute): fix the issue where the number type is displayed using scientific notation when the value is 0.0000000

### LINKED ISSUE #762: [Enhancement]: MySQL租户，将显示为科学计数法的小数正常显示成小数

ODC Server: 4.2.1-20231009

CREATE TABLE tn (c1 decimal(20,7) ) ;
insert into c1 values (0.0000000);

期望正常显示为小数
image

### Why is this needed?

开发测试阶段使用方便

yhilmare:
#### What type of PR is this?

type-bug

the number content will be displayed by using scientific notation when the value is 0.0000000 which confused the user. this pr fix this problem.

I have tested for three number datatype: `number`, `decimal` and `fixed` and I have also tested these datatypes with zerofill settings. Both two scenarios are normal.

Fixes #762 

```docs

```


smallsheeeep (approved):
LGTM

Issue #210, opened 2023-09-06
[Bug]: CAS redirect URL is http while the actual is https


smallsheeeep:
### ODC version

4.2.0

### OB version

Regardless of OB version

### What happened?

CAS redirect URL is http while the actual is https:
<img width="1502" alt="image" src="https://github.com/oceanbase/odc/assets/84772536/946ce38d-0089-4d14-aee4-637b432d1a4c">

### What did you expect to happen?

The CAS redirect URL should be https

### How can we reproduce it (as minimally and precisely as possible)?

<null>

ungreat:
Typically, this is caused by the use of load balancing.The host used by ODC's redirectUrl is determined by the request to the ODC server.The load balancer listens on the HTTPS port and redirect the request to the ODC using HTTP，may result in a different protocol for the redirect url.
In this case,  you can specify odc site url by changing the record of the `config_system_configuration` table, which key is `odc.site.url`.Note that this value cannot contain localhost, otherwise it will not take effect.

ungreat:
Typically, this is caused by the use of load balancing.The host used by ODC's redirectUrl is determined by the request to the ODC server.The load balancer listens on the HTTPS port and redirect the request to the ODC using HTTP，may result in a different protocol for the redirect url.
In this case,  you can specify odc site url by changing the record of the `config_system_configuration` table, which key is `odc.site.url`.Note that this value cannot contain localhost, otherwise it will not take effect.

Huangxiao-mas:
passed

PR #766, opened 2023-11-10
fix(SSO): test login protocal not match

ungreat:
#### What type of PR is this?
fix test login redirect url schema may not match. see https://github.com/oceanbase/odc/issues/210

ODC not use {baseUrl} to rewrite redirect Url, instead of , use the host obtained by the front-end.

Fixes #https://github.com/oceanbase/odc/issues/210

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/integration/model/Oauth2Parameter.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/integration/model/Oauth2Parameter.java
@@ -18,6 +18,7 @@ package com.oceanbase.odc.service.integration.model;
 import static com.oceanbase.odc.core.shared.constant.OdcConstants.ODC_BACK_URL_PARAM;
 import static com.oceanbase.odc.service.integration.model.SSOIntegrationConfig.parseOrganizationId;
 
+import java.net.URL;
 import java.util.Set;
 
 import org.springframework.boot.autoconfigure.security.oauth2.client.OAuth2ClientProperties.Provider;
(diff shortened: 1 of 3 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/integration/model/Oauth2Parameter.java
```diff
@@ -100,6 +99,11 @@ private Provider toProvider(String issueUrl) {
         return provider;
     }
 
+    @SneakyThrows
+    private String getRedirectHost(String redirectUrl) {
+        URL url = new URL(redirectUrl);
+        return url.getProtocol() + "://" + url.getHost() + (url.getPort() > 0 ? ":" + url.getPort() : "");
```
yizhouxw, line 105: if port is 80 for http protocal, or port is 443 for https protocal, skip  port part in url.

yizhouxw: may refer spring util `UrlUtils.buildFullRequestUrl`

ungreat: `UrlUtils.buildFullRequestUrl` is more applicable to Request objects, which returns the default port of the protocol when you get port. Here use java.net.URL to parse url string, if there is no port, the value of port will be -1, which will be lead error.

yizhouxw (approved):
lgtm

PR #781, opened 2023-11-13
fix(result-set): only allow to edit result set when there is pk / uk / rowid

### LINKED ISSUE #782: [Feature]: optimize result set update logic

When the query result set does not contain the primary key or UK, ODC will splice all units into where conditions. 
This may lead to uncontrollable factors in the updated results (although there are reminders on the page), and this uncertainty should be avoided.

In this case, editing of the result set should not be allowed.

### Additional context

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-feature

When the query result set does not contain the primary key or UK, ODC will splice all units into where conditions. This is restricted in this PR and no editing is allowed in this case.

Fixes #782


yhilmare (commented):
@yizhouxw this pr will not allow user modify data if a table doest not contain uk/pk

server/odc-core/src/main/resources/i18n/ErrorMessages_zh_CN.properties
```diff
@@ -115,6 +115,7 @@ com.oceanbase.odc.ErrorCodes.ObInvalidCreateSqlCondition=无法创建 SQL，条
 com.oceanbase.odc.ErrorCodes.ObInvalidPartType=无效的分区类型 ''{0}''
 com.oceanbase.odc.ErrorCodes.ObPartTypeNotSupported=不支持 ''{0}'' 分区类型，目前只支持 ''{1}''
 com.oceanbase.odc.ErrorCodes.ObBasicTypeColumnRequired=缺少基本类型列
+com.oceanbase.odc.ErrorCodes.ObUpdateKeyRequired=缺少主键或唯一约束，无法生成更新条件
```
yizhouxw, line 118: let's add recommended actions to user, not only the reason why not supported, such as:
当年结果集不支持修改或删除数据，因为结果集的列不包含任意主键/唯一性约束，基于当前结果集包含的列生成的更新条件可能会导致其他数据行受影响，如需要修改数据，请在 SQL 窗口执行语句或者发起数据库变更任务

yhilmare (approved):
lgtm

Issue #180, opened 2023-09-06
[Bug]: failed to purge a specific object in recyclebin


yhilmare:
### ODC version

4.2.0

### OB version

OceanBase 4.x

### What happened?

I want to purge a specific table object in recyclebin, but there is an exception thrown:
![image (1)](https://github.com/oceanbase/odc/assets/22992296/6ca1276e-caa9-4840-ae91-a4b2a1da880d)

I can do this on OceanBase 3.x without any errors.

### What did you expect to happen?

I can do this without any errors

### How can we reproduce it (as minimally and precisely as possible)?

purge a specific table object on OceanBase 4.x

yhilmare:
this issue is caused by odc 4.2.0, the root cause of it is that OceanBase can not allow user purge an object located in A schema by. a connection that connected to B schema

yhilmare:
<img width="1114" alt="截屏2023-10-31 16 51 41" src="https://github.com/oceanbase/odc/assets/22992296/2f6daa3a-ccf4-4b2f-a74b-b46b36452153">

as you can see, we use recyclebin in datasource management panel, the connection we use will connect to any other schema that can not be controlled by user but the object in recyclebin that you want to purge is located in a specific schema. that is why we will get an error when we attempt to purge it.

yhilmare:
we will switch schema when we attempt to purge an object

HSunboy:
Fixed In Commit [597ec21486460ffebed6156c85168a47b650f6fb](https://github.com/oceanbase/odc-client/commit/597ec21486460ffebed6156c85168a47b650f6fb)
            ------
            晓康
            Mon Nov 13 14:21:20 2023 +0800

PR #783, opened 2023-11-13
fix(recyclebin): fix can not delete recyclebin objects

### LINKED ISSUE #705: [Bug]: failed to operate recyclebin in a datasource that isn't related to any projects
### ODC version

4.2.0

### OB version

not related

### What happened?

I create a datasource that isn't related to any projects. I switch to recyclebin tab and attempt to flash back an object.

截屏2023-11-02 15 20 17

I get an error prompt when I click 'execute' button:

截屏2023-11-02 15 21 45

flashback this object normally

follow 'what happened'

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

there are two bug related to recyclebin

1. failed to flashback db object that related to a database/schema which is not assigned to any project. The root cause of this bug is that the front end will call `asyncExecute` api to execute the flashback sql which may be intercepted by security policy.
2. failed to delete recyclebin objects in OceanBase 4.1.x. the root cause of this is that OceanBase 4.1.x doesn't allow user delete an object located in A schema and the schema connect to is B.

To fix these issues, I create 3 new apis to replace the old one:
1. I executed the flashback sql directly so that we can avoid the interception of the security policy
2. I switch schema to the db object located in.

Fixes #705 #180 

```docs

```

### DIFF
--- a/server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v1/DBRecyclebinController.java
+++ b/server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v1/DBRecyclebinController.java
@@ -26,7 +26,6 @@ import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RestController;
 
 import com.oceanbase.odc.core.session.ConnectionSession;
-import com.oceanbase.odc.service.common.model.ResourceSql;
 import com.oceanbase.odc.service.common.response.OdcResult;
 import com.oceanbase.odc.service.common.util.SidUtils;
 import com.oceanbase.odc.service.db.DBRecyclebinService;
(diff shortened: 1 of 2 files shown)


server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v1/DBRecyclebinController.java
```diff
@@ -59,26 +58,25 @@ public OdcResult<List<DBRecycleObject>> list(@PathVariable String sid) {
...
     }
 
-    @ApiOperation(value = "getDeleteSql", notes = "获取清除回收站指定对象的sql，sid示例：sid:1000-1:d:db1")
-    @RequestMapping(value = "/getDeleteSql/{sid:.*}", method = RequestMethod.PATCH)
-    public OdcResult<ResourceSql> getDeleteSql(@PathVariable String sid,
-            @RequestBody List<DBRecycleObject> resource) {
-        ConnectionSession session = sessionService.nullSafeGet(SidUtils.getSessionId(sid));
-        return OdcResult.ok(ResourceSql.ofSql(recyclebinService.getPurgeSql(session, resource)));
+    @ApiOperation(value = "purgeObject", notes = "purge a specific db object")
+    @RequestMapping(value = "/purge/{sid:.*}", method = RequestMethod.PATCH)
```
yizhouxw, line 62: if the API was totally redesigned, how about keep the v1 API , and use v2 API for new design?

we should keep API compatibility as  concept of changing.

yhilmare: done, make v2 api and return SuccessResponse instead of OdcResult

yizhouxw (approved):
lgtm

PR #795, opened 2023-11-13
feat(auth): add system config for max attempt times and account lock time

### LINKED ISSUE #245: [Enhancement]: optimize the behavior of odc in the scenario of login failure

1. Increase the system configuration item to control the time interval of retrying in the case of login failure. A shorter time can be configured according to customer scenarios, such as 30 seconds. Even 0 seconds is unlimited.
2. When the login is restricted, the error message needs to be more clear, inform the specific reason, and how long before you can try again.

### Why is this needed?

these two enhancements can improve user's experience

yhilmare:
#### What type of PR is this?

type-feature

some users consider that the max retry times for login failure and account lock time can not be configured is not so good. This pr is to fix this issue.

1. I add two system config `odc.iam.auth.alipay.max-failed-login-attempt-times` and `odc.iam.auth.alipay.failed-login-lock-timeout-seconds` to control max retry times and timeout for locking account. If these two item is set to zero which means no limit.
2. I change the localized message for some error code cause that the current message contains max retry times info and these info will not exist when the correspond config is set to zero
3. here is the behavior of the odc:

||`odc.iam.auth.alipay.max-failed-login-attempt-times <= 0`|`odc.iam.auth.alipay.max-failed-login-attempt-times > 0`|
|:----:|:----:|:----:|
|`odc.iam.auth.alipay.failed-login-lock-timeout-seconds <= 0`|no limit|account will be locked when attach the max retry times and will be unlocked in next time|
|`odc.iam.auth.alipay.failed-login-lock-timeout-seconds > 0`|no limit|as usual|

Fixes #245 

```docs

```


server/odc-core/src/main/resources/i18n/ErrorMessages.properties
```diff
@@ -16,7 +16,7 @@ com.oceanbase.odc.ErrorCodes.UnauthorizedSessionAccess=Unauthorized session acce
 com.oceanbase.odc.ErrorCodes.PermissionChanged=Permission has been changed, please exit and try again
 com.oceanbase.odc.ErrorCodes.UnauthorizedDataAccess=Unauthorized data access
 com.oceanbase.odc.ErrorCodes.OverLimit=Cannot process due {0} exceed limit, maximum was {1}
-com.oceanbase.odc.ErrorCodes.AttemptLoginOverLimit=Cannot process due {0} exceed limit, maximum was {1}, please try {2} second(s) later
+com.oceanbase.odc.ErrorCodes.AttemptLoginOverLimit=Cannot process due {0} exceed limit, maximum was {1}, please try {2} second(s) later, or contact the administrator to modify the system configuration
```
yizhouxw, line 19: here we may not change error message, or we are encouraging user to disturb the administrator

server/odc-server/src/main/resources/data.sql
```diff
@@ -696,3 +696,9 @@ INSERT INTO config_system_configuration ( `key`, `value`, `description` ) VALUES
 --
 INSERT INTO config_system_configuration(`key`, `value`, `description`) VALUES('odc.integration.url-white-list',
  '', 'Only whitelisted URLs are allowed when configuring the integration API to prevent SSRF security vulnerabilities. The whitelist is empty by default, allowing access to all URLs.') ON DUPLICATE KEY UPDATE `id`=`id`;
+
+---
+--- v4.2.3
+---
+INSERT INTO config_system_configuration ( `key`, `value`, `description` ) VALUES( 'odc.iam.auth.alipay.max-failed-login-attempt-times', '5', '登录失败情况下最大重试次数，小于等于 0 意味着次数无限制' ) ON DUPLICATE KEY UPDATE `id` = `id`;
```
yizhouxw, line 703: description 增加 “，重启生效”

server/odc-core/src/main/resources/i18n/ErrorMessages.properties
yhilmare: done

server/odc-server/src/main/resources/data.sql
yhilmare: done

yizhouxw (approved):
lgtm

PR #814, opened 2023-11-14
fix(schema-plugin):  cannot get table detail in odp sharding mysql mode when lower_case_table_names = 1 or 2

### LINKED ISSUE #793: [Bug]: Error in viewing table details in ODP sharding mysql mode
### ODC version

4.2.3

### OB version

odp sharding mysql when lower_case_table_names = 1 or 2

### What happened?

Click "View Table Details" and the front-end calls the getTable interface twice. When the interface is called for the first time, the table name passed by the front-end is the table name passed by the back-end. The returned table information of table name is lowercase which is different from `list tables` result, and this lead to the interface is called secondly time to check table
image

When the interface is called the second time, the table name passed by the front-end is lowercase. , causing the backend to throw an exception "the table with tableName as tab2 does not exist".
image

View table details normally

1. Create an uppercase table
2. Click "View Table Details"

_No response_

### Cloud

_No response_

PeachThinking:
#### What type of PR is this?
type-bug
module-Database object

fix cannot get table detail in odp sharding mysql when lower_case_table_names = 1 or 2.
In ODP sharding mysql mode, when variable lower_case_table_names = 1 or 2, the table name will be stored in uppercase letters. There is no need to automatic lowercase conversion of table name when database is Case Sensitive.

Fixes #793

```docs

```


yhilmare (approved):
lgtm

PR #834, opened 2023-11-14
enhancement(sql-check): add cache for sql-check

### LINKED ISSUE #666: [Enhancement]: add cache for long duration phase

 Add cache for long duration phase, such as sql-check rule loading, sql console rule loading.
a

### Why is this needed?

The time consuming will be too large to accept if we don't do that

yhilmare:
#### What type of PR is this?

type-enhancement

sql-check will load all rules from metadb for every query which costs too much time and I reuse the ruleset cache to accelerate it.

Fixes #666 

```docs

```


yizhouxw (approved):
lgtm

PR #835, opened 2023-11-14
enhancement(schema-plugin): speed up getting index and improve code robustness when getting table ddl

### LINKED ISSUE #571: [Enhancement]: several enahncement on table extension and db-browser

1. full text replacement may lead to some problems, we need another way to add schema message [OBOracleTableExtension](https://github.com/oceanbase/odc/blob/8dffddf93c97b11f19dd58c0553f1c116494fd6b/server/plugins/schema-plugin-ob-oracle/src/main/java/com/oceanbase/odc/plugin/schema/oboracle/OBOracleTableExtension.java)
```Java
// OBOracleTableExtension#getTableDDL
JdbcOperationsUtil.getJdbcOperations(connection).query(getTableDDlSql, t -> {
          // Create table ddl like this: CREATE [GLOBAL TEMPORARY|SHARDED|DUPLICATED] TABLE T...
          String ddl = t.getString(1);
          if (Objects.nonNull(ddl)) {
              // fix: Replace " TABLE " to " TABLE schemaName."
              ddlRef.set(StringUtils.replace(ddl, " TABLE ",
                      " TABLE " + StringUtils.quoteOracleIdentifier(schemaName) + ".", 1));
          }
      });
```

2. Use `DBMS_METADATA` to get DDL of a view. It is not so good to generate DDL by ourself [OBOracleSchemaAccessor](https://github.com/oceanbase/odc/blob/8dffddf93c97b11f19dd58c0553f1c116494fd6b/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/oracle/OBOracleSchemaAccessor.java):

```Java
// OBOracleSchemaAccessor#getView
boolean updatable = fillOracleUpdatableInfo(view);
        String ddl = String.format("CREATE VIEW %s AS %s",
                StringUtils.quoteOracleIdentifier(viewName), partDdl);
        if (!updatable && !ddl.toUpperCase().contains("WITH READ ONLY")) {
            view.setDdl(ddl + " WITH READ ONLY");
        } else {
            view.setDdl(ddl);
        }
```

Tips: If we do this above, we can avoid using internal virtual table to get `VIEW_IS_UPDATABLE`.

3. Use `show create function` and `show create procedure` to get PL object's DDL for MySQL mode.

### Why is this needed?

Increase program robustness

PeachThinking:
#### What type of PR is this?
type-enhancement
module-Database object

1. Use `sqlParser` to determine whether there is a SchemaName in the DDL of the table. If not, splice the SchemaName in DDL.
2. Optimize `OBOracleGetDBTableByParser#listIndexes` and `OBOracleTableExtension#getTableDDL` repeatedly execute the query SQL to obtain the DDL of the index

Fixes #571

```docs

```

### DIFF
--- a/server/plugins/schema-plugin-ob-oracle/src/main/java/com/oceanbase/odc/plugin/schema/oboracle/OBOracleTableExtension.java
+++ b/server/plugins/schema-plugin-ob-oracle/src/main/java/com/oceanbase/odc/plugin/schema/oboracle/OBOracleTableExtension.java
@@ -20,11 +20,10 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.concurrent.atomic.AtomicReference;
 
+import org.apache.commons.lang3.Validate;
 import org.pf4j.Extension;
 
-import com.oceanbase.odc.common.util.JdbcOperationsUtil;
 import com.oceanbase.odc.plugin.schema.obmysql.OBMySQLTableExtension;
 import com.oceanbase.odc.plugin.schema.oboracle.parser.OBOracleGetDBTableByParser;
 import com.oceanbase.odc.plugin.schema.oboracle.utils.DBAccessorUtil;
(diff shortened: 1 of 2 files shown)


yhilmare (approved):
lgtm

PR #837, opened 2023-11-14
fix(security): add white list for security scanning and modify mysql pl parser's g4

yhilmare:
#### What type of PR is this?

type-security
type-bug

add with list for security scanning.

Fixes #

```docs

```


smallsheeeep (approved):
LGTM

PR #839, opened 2023-11-14
fix(name): change resource name length limit from 64 to 128

smallsheeeep:
#### What type of PR is this?
type-bug

In cloud mode, the user name may longer than 64 chars. This PR change resource name length limit from 64 to 128.


yhilmare (dismissed):
lgtm

yizhouxw (approved):

PR #843, opened 2023-11-15
fix(data-security): regex column recognization rule may suffer ReDos

### LINKED ISSUE #842: [Bug]: Sensitive column recognization rules using Regex may suffer ReDos attack
### ODC version

4.2.2

### OB version

null

### What happened?

ODC may suffer from ReDos attack.

Defending ReDos attack

I create a table and comment a column with aaaa...(a*2000)b. 
And then I create a column recognization rules with Regex mode and the regex expression is `(a|a?)+` or `(a|aa)+`.
When I create a scanning task, the task may cause ODC server crush down. 

_No response_

### Cloud

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
module-data security

ODC server may suffer ReDos attack. This PR fix it.

Fixes #842 

Self-test passed

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1047,6 +1047,7 @@
                 <groupId>org.awaitility</groupId>
                 <artifactId>awaitility</artifactId>
                 <version>${awaitility.version}</version>
+                <scope>test</scope>
             </dependency>
             <!-- 以上测试依赖 -->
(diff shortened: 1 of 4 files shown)


yhilmare (approved):
lgtm

PR #844, opened 2023-11-15
feat(session): make connect session auto-reconnect when session is deleted or disabled

### LINKED ISSUE #105: [Bug]: open the connection after a period of time error "ID for XXX session does not exist"
### ODC version

ODC 4.2.0

### OB version

Any version of OceanBase

### What happened?

Open the connection and restart the ODC. If you perform operations within the connection, an error message is displayed indicating that the session whose id is xxx does not exist

Can be automatically reconnected

Open the connection for a period of time, do not operate, open the connection window again can be repeated

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-feature

The front end will give an error prompt when connect session is deleted or disabled, this pr make odc auto-reconnect when session is not available.

This pr does these things below:

1. Remove useless code.
    1. several connection session's attributes are not used in odc 4.2.x anymore such as `permitted actions`, `visible scope`, `account type` and so on and I delete them.
    2. remove `ConnectionAccountType` from connection session module cause that this is a deprecated concept in odc 4.2.x.
    3. delete some unused class, suck as `ConnectionPermissionCheckResp`, `SwitchSchemaReq` and so on.

2. Avoid switching schema on every sql execution. The `java.sql.connection` that the `DataSourceFactory` create will connect to the schema that the username standards for and odc will set `ConnectionSessionUtil#setCurrentSchema` to guarantee that the actual schema is right by switching schema on every sql execution which will slow down the execution.

we can not change the target schema that the `java.sql.connection` connect to by changing `ConnectionConfig#defaultSchema` and this field behaves as follow:

|setting value|giving value(OracleMode)|giving value(MySQL Mode)|
|:--:|:--:|:--:|
|null|`ConnectionConfig#username`|`information_schema`|
|not null|`ConnectionConfig#username`|setting value|

As you can see, we can not determine the schema that we want to connect by setting `defaultSchema` on oracle mode which makes non-scene so I change the `defaultSchema` field's logic as follow:

|setting value|giving value(OracleMode)|giving value(MySQL Mode)|
|:--:|:--:|:--:|
|null|as usual|as usual|
|not null|setting value|as usual|

Thus, we can connect to the schema what we want by setting `defaultSchema` and avoid switching schema on every sql execution. I set an attribute `CONNECT_SCHEMA` in `ConnectionSession` to record the schema we connect to originally and we will switch the schema if the current schema is not equal to `CONNECT_SCHEMA`.

I have tested for this but I can not say that there will be fine and I notice that the `defaultSchema` is used for some other scenarios, please check me @krihy @guowl3 

3. Make connection session auto-reconnect when the session is not available.
    1. I modify the session id's generation logic to record the necessary information such as `datasource_id` and `database_id`, you can get more information by viewing `DefaultConnectSessionIdGenerator`, this class show us how to encode a session id and how to decode session id to get these information. Further more, we use 10 bytes short uuid to identify the connection session.
    2. add a new method `ConnectionSessionService#nullSafeGet` get or create a connection session. I try to get the connection session and I will get the necessary information from the session id and re-create a session by these information when I failed to get the connection session which the session id identify. By the way, the new created session will get the same session id with the old one.
    3. most of the scenarios will enjoy this connection session auto reconnection logic except for these:

|Scenario|Reason|
|:--:|:--:|
|Get PL Dbms output|we will get illegal dbms output info if we use re-connected session and confuse the user|
|Get Full Link Trace Msg|Full link message is stored in connection session's cache|
|Kill Query|Can not kill a older query using a new created connection session|
|Get SQL execution result|user will get an empty result if he try to get result from a new session which may confuse the user|

Here is some tips that you may want to know:

To avoid duplicated session id, I use short uuid to replace original auto increased digital id. 

The short uuid is 10 bytes long which is much shorter than standard uuid which may lead to conflicts. I did experiments for 20 times to show the possibility of the conflict, here is the result:

|No.|Max Times Encounter Conflict|
|:--:|:--:|
|1|1149045|
|2|792428|
|3|189651|
|4|983708|
|5|1254116|
|6|824129|
|7|470938|
|8|811145|
|9|135047|
|10|159263|
|11|849475|
|12|1508997|
|13|202297|
|14|1020124|
|15|1261784|
|16|816476|
|17|472039|
|18|280303|
|19|200573|
|20|1011938|

Thus. 10 bytes long is suitable.

Fixes #105 

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/session/factory/DefaultConnectSessionIdGenerator.java
```diff
@@ -0,0 +1,57 @@
...
+import com.oceanbase.odc.service.connection.model.CreateSessionReq;
+
+import lombok.NonNull;
+import lombok.Setter;
+
+@Setter
+public class DefaultConnectSessionIdGenerator implements ConnectionSessionIdGenerator<CreateSessionReq> {
+
+    private Long databaseId;
+    private String fixRealId;
```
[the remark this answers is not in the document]
krihy: LGTM  
cc @guowl3

yhilmare:
there are still two tips I want to discuss:
1. Is it necessary to maintain the connection session's timeout? (current timeout is 8 hrs, should we make it shorter?)
2. Should we stop the first sql executing while the connection session reconnected?

LuckyPickleZZ:
> 2. Should we stop the first sql executing while the connection session reconnected?

If we can prompt the user with some information on the interaction, such as that the current session has been reset. Then we may not need to stop the first SQL

yhilmare:
this issue will not affect DLM @guowl3

yhilmare:
> > 2. Should we stop the first sql executing while the connection session reconnected?
> 
> If we can prompt the user with some information on the interaction, such as that the current session has been reset. Then we may not need to stop the first SQL

We'd better stop the first sql execution since reconnection to prevent the bad influence. We may consider a situation as follow:

> some one is executing sql in a transaction. 
1. The connection session is closed suddenly but the user does not know currently. 
2. He runs sql  without knowing of closing of the connection session.
3. ODC will create a new connection session.
4. if we don't stop the execution, this sql may lead to some bad influence cause the sql is running in a new transaction actually but the user expect that the sql is running in the former transaction.

smallsheeeep:
> 3\. most of the scenarios will enjoy this connection session auto reconnection logic except for these:

So, shall we do something  to solve these problems?

yhilmare:
> > 3. most of the scenarios will enjoy this connection session auto reconnection logic except for these:
> 
> So, shall we do something to solve these problems?

not yet, we can just prompt user~

server/odc-core/src/main/java/com/oceanbase/odc/core/session/InMemoryConnectionSessionRepository.java
```diff
@@ -43,6 +43,9 @@ public class InMemorySessionRepository implements ConnectionSessionRepository {
     @Override
     public String store(@NonNull ConnectionSession session) {
         Validate.notNull(session.getId(), "Session.Id can not be null");
+        if (this.sessionId2SessionMap.containsKey(session.getId())) {
```
smallsheeeep, line 46: So what will happen if generate a duplicated sessionId?

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionStatusManager.java
```diff
@@ -199,21 +197,13 @@ private class CheckTask implements Callable<ConnectionTestResult> {
         final CheckState checkState;
         final User user;
 
-        CheckTask(ConnectionConfig connection, CheckState checkState, ConnectionAccountType accountType) {
-            this.testConnectionReq = TestConnectionReq.fromConnection(connection, accountType);
+        CheckTask(ConnectionConfig connection, CheckState checkState) {
+            this.testConnectionReq = TestConnectionReq.fromConnection(connection, ConnectionAccountType.MAIN);
```
smallsheeeep, line 201: `ConnectionAccountType.MAIN` should be deleted?

server/odc-core/src/main/java/com/oceanbase/odc/core/session/InMemoryConnectionSessionRepository.java
yhilmare: as you can see, an exception will be thrown and the user will be prompted that failed to enter a database. he/she need to re-entry the database. 

> Will the session be destroied?

no, ODC will do nothing.

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionStatusManager.java
yhilmare: no, this code is also used at `DataTransferService#testSysTenantAccount`

smallsheeeep (approved):
LGTM

server/plugins/connect-plugin-api/src/main/java/com/oceanbase/odc/plugin/connect/api/SessionExtensionPoint.java
```diff
@@ -20,18 +20,17 @@
...
+
 /**
  * @author yaobin
  * @date 2023-04-14
  * @since 4.2.0
  */
-public interface SessionExtensionPoint extends ExtensionPoint {
-
-    void killQuery(Connection connection, String connectionId);
+public interface SessionExtensionPoint extends ExtensionPoint, SessionOperations {
```
PeachThinking, line 30: Why abstract a `SessionOperations` interface?

yhilmare: we need to use `SessionExtensionPoint` at `odc-core` and we define `SessionExtensionPoint` at `connect-plugin-api`. 

so, we need to make `odc-core` rely on `connect-plugin-api` if we want to use `SessionExtensionPoint` at `odc-core` but `connect-plugin-api` reply on `odc-core` actually. That is to say that we can not make `odc-core` rely on `connect-plugin-api` to avoid circular dependency.

I create the `SessionOperations` and move some methods that located in `SessionExtensionPoint` originally and used by `odc-core` to `odc-core` to fix this problem.

server/odc-service/src/main/java/com/oceanbase/odc/service/session/initializer/SwitchSchemaInitializer.java
```diff
@@ -48,11 +50,13 @@ public void init(Connection connection) throws SQLException {
         if (this.connectionSession.isExpired()) {
             return;
         }
-        String schema = ConnectionSessionUtil.getCurrentSchema(this.connectionSession);
-        if (connection == null || schema == null) {
+        String currentSchema = ConnectionSessionUtil.getCurrentSchema(this.connectionSession);
```
LuckyPickleZZ, line 53: Can this logic work properly in the following scenario?
1. I created a session with connectSchema A;
2. I switched it into currentSchema B;
3. The session was expired, then reconnected.
I wonder how to know which currentSchema is in this case.

yhilmare: odc does not support switch schema currently.

Unfortunately, we can not switch to the schema that he/she switch to before if we supports schema in the future and the connect session is reconnected

smallsheeeep (dismissed):

smallsheeeep (approved):

PR #845, opened 2023-11-15
fix(data-editing): optimize error message when the length of field exceeds the maximum limit

### LINKED ISSUE #401: [Bug]: Optimize the error message, create a JSON type table, enter a long JSON content, and get the error "The number must be between 0 and 204800"
### ODC version

ODC4.2.1

### OB version

Oceanbase 4.1.0

### What happened?

Create a JSON type table, input a very long JSON content, and report ‘the number must be between 0 and 204800’
image

Create a JSON type table, input a very long JSON content, and report ‘the number must be between 0 and 204800’

1. Create table, type json
2. Enter the super long json and click to confirm the modification.

_No response_

### Cloud

_No response_

PeachThinking:
#### What type of PR is this?
type-bug
module-Data viewing and editing

optimize error message when the length of data inserted into a json field exceeds the maximum limit.
before:
image
after:
image

Fixes #401

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/dml/model/DataModifyUnit.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/dml/model/DataModifyUnit.java
@@ -41,7 +41,8 @@ public class DataModifyUnit {
     private String columnName;
     private String columnType;
     private String oldData;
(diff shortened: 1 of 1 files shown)


yhilmare (approved):
lgtm

PR #846, opened 2023-11-15
fix(db-browser): add "SYS" prefix for oracle dictionary views

### LINKED ISSUE #602: [Bug]: sql statements for querying tables name need to be concatenated with sys tenants
### ODC version

ODC 4.2.2

### OB version

Any version of OceanBase.

### What happened?

The SQL of querying table name is:
SELECT TABLE_NAME FROM ALL_TABLES WHERE OWNER='XSTORACLE' ORDER BY TABLE_NAME ASC;
If the database has a table named ALL_TABLES，an error will be reported.

The SQL of querying table name is:
SELECT TABLE_NAME FROM SYS.ALL_TABLES WHERE OWNER='XSTORACLE' ORDER BY TABLE_NAME ASC;

If the database has a table named ALL_TABLES，an error will be reported.

_No response_

### Cloud

_No response_

PeachThinking:
#### What type of PR is this?
type-bug
module-Database object

When `OracleSchemaAccessor` queries Oracle dictionary views, some SQL do not add the `SYS` prefix. As a result, if there is a table or view with the same name as the view under the current user, the query under the current user will be given priority.

Fixes #602

```docs

```


libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/oracle/OBOracleSchemaAccessor.java
```diff
@@ -111,7 +113,7 @@ public DBDatabase getDatabase(String schemaName) {
     @Override
     public List<DBDatabase> listDatabases() {
         List<DBDatabase> databases = new ArrayList<>();
-        String sql = "SELECT USERNAME from ALL_USERS;";
+        String sql = "SELECT USERNAME from SYS.ALL_USERS;";
```
yhilmare, line 116: why don't we use `dataDictTableNames.USERS()`?

yhilmare (commented):
if you change the db-browser, you should upgrade the db-browser's version cause we have to publish it to the maven central repo to guarantee that we can build it successfully when we attempt to release it

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/oracle/OBOracleSchemaAccessor.java
PeachThinking: Because this is a String type sql, there is no need to use OracleSqlBuilder, so I just added the `SYS` prefix. It is better to use dataDictTableNames uniformly. I will modify it.

PeachThinking:
> if you change the db-browser, you should upgrade the db-browser's version cause we have to publish it to the maven central repo to guarantee that we can build it successfully when we attempt to release it

done

yhilmare (approved):
lgtm

PR #850, opened 2023-11-16
docs: update developer guide

yhilmare:
#### What type of PR is this?

there is a discussion #336 about how to startup odc and we find that our developer guide leaks installation of plugin and relied modules which may lead to this issue.

so that I update the developer guide and add these contents.

Fixes #

```docs

```


docs/en-US/DEVELOPER_GUIDE.md
```diff
@@ -231,6 +231,22 @@ When plaintext configuration is identified, EncryptableConfiguration will automa
 
 ## 3.1 jar build and run
 
+### 3.1.1 Dependent component installation
+
+ODC relies on 2 self-developed components, which you can see in `lib/ob-sql-parser` and `lib/db-browser` respectively. In the official release, if these two components are modified, we will upload them to the maven central warehouse for reference in advance, but during the development process, we will reference them through local installation. You need to manually install these two components locally before building, otherwise the build may not be successful. You can complete the installation of dependent components through the following shell script:
+
```
yizhouxw, line 237: we can also use `script/build_libs.sh`

yhilmare: done

yizhouxw (approved):
lgtm

PR #856, opened 2023-11-20
fix(connect-plugin): fix mysql jdbc security vulnerability

PeachThinking:
#### What type of PR is this?

security bug
module-Adapt to mysql

fix 422 security bug: 
Mysql JDBC arbitrary file reading vulnerability exists in the test connection interface: [api/v2/connect/test?currentOrganizationId=1&ignoreError=true](http://11.162.217.57:9199/api/v2/connect/test?currentOrganizationId=1&ignoreError=true)

fix method :https://yuque.antfin.com/fcloudsec/hqu249/pyk140xgyt03uzxz

```docs

```


MarkPotato777 (changes requested):
PR title advice: `fix(security): MySQL JDBC arbitrary file reading vulnerability`
- it's not necessary to mention 'fix' again in the title given the PR starts with `fix(xxx):`
- this is related to a security issue, so use security as the module name instead of the business module

PeachThinking:
> PR title advice: `fix(security): MySQL JDBC arbitrary file reading vulnerability`
> 
> * it's not necessary to mention 'fix' again in the title given the PR starts with `fix(xxx):`
> * this is related to a security issue, so use security as the module name instead of the business module

👌

yhilmare (approved):
lgtm

MarkPotato777 (commented):
ASD

PR #857, opened 2023-11-20
fix(dlm):data delete failure in periodic task and remove sys tenant verification

guowl3:
#### What type of PR is this?
type-bug

The data delete job will work as a periodic task, and users cannot configure the sys tenant user.

Fixes #

```docs

```


yizhouxw (approved):
lgtm

PR #858, opened 2023-11-20
feat(sql-execution): let sql be only parsed once during execution

yhilmare:
#### What type of PR is this?

type-feature

we will parse sql multi times during its execution that costs too much time. I parse sql only once during its execution.

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/DefaultSqlChecker.java
```diff
@@ -91,34 +70,4 @@ protected List<CheckViolation> doCheck(Statement statement, SqlCheckContext cont
                 .flatMap(rule -> rule.check(statement, context).stream()).collect(Collectors.toList());
     }
 
-    /**
-     * 目前 ob 的 parser 在解析 pl 和 sql 是分开的，这导致 sql check 时会存在一个问题：如果用户送检的是一个 pl 的 ddl，那么此时的 sql parser
-     * 就会报错语法错误。这里这个方法就是为了处理这种情况：当 parser 报语法错误的时候再用 pl 的 parser 解析一次，如果 pl 的 parser 也报错才认为有语法错误。
-     */
-    private void tryParsePl(String sql) {
```
LuckyPickleZZ, line 98: This logic here seems to be adapted to PL. So where will it be implemented after being deleted here?

yhilmare: as you can see:

```Java
// server/odc-core/src/main/java/com/oceanbase/odc/core/sql/parser/BaseOBAstFactory.java
 @Override
    public AbstractSyntaxTree buildAst(@NonNull String statement) throws SyntaxErrorException {
        long start = System.currentTimeMillis();
        SyntaxErrorException thrown;
        BaseSQLParser<? extends Lexer, ? extends Parser> sqlParser = getSqlParser();
        try {
            return doBuildAst(sqlParser.buildAst(new StringReader(statement)), sqlParser);
        } catch (SyntaxErrorException e) {
            thrown = e;
        }
        this.timeoutMillis = this.timeoutMillis - (System.currentTimeMillis() - start);
        if (this.timeoutMillis <= 0) {
            throw new ParseCancellationException("Timeout, abort!");
        }
        BaseSQLParser<? extends Lexer, ? extends Parser> plParser = getPlParser();
        try {
            return doBuildAst(plParser.buildAst(new StringReader(statement)), plParser);
        } catch (SyntaxErrorException e) {
            throw thrown;
        }
    }
```

we parse the input as the sql firstly and we will parse it as pl secondly if we get a syntax error exception at first step.

yizhouxw (approved):
lgtm

PR #860, opened 2023-11-20
fix(migrate): failed to start up when there is no users or organizations

yhilmare:
#### What type of PR is this?

type-bug

this pr fix that odc failed to start up when there is no users or organizations

Fixes #

```docs

```


MarkPotato777 (approved):
LGTM

PR #861, opened 2023-11-20
feat(obclient): upgrade obclient to 2.2.4

### LINKED ISSUE #324: [Feature]: Upgrade obclient tool

ODC is using obclient v1.2.8, which is out of date.
In order to custom disabled commands, it must be upgraded.

Upgrade it to a stable and usable version.

### Additional context

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-feature

In this PR, we will upgrade obclient's version from 1.2.8 to 2.2.4.
Obclient allows disable commands to be specified by `--ob-disable_commands` param. I have specified that the following command will be disabled.
`connect,conn,edit,nopager,notee,pager,print,prompt,rehash,system,tee,resetconnection`
User could custom these commands in config_system_configuration table.

Plus, the new version of obclient requires manual installation of the `libncurses.so.5` library. Therefore, I placed it in the lib directory and would copy it to the `/usr/lib64` directory after decompression.

Fixes #324


server/odc-server/pom.xml
```diff
@@ -290,7 +290,8 @@
                     </preinstallScriptlet>
                     <postinstallScriptlet>
                         <script>tar -xvzf ${rpm.rootdir}/bin/obclient.tar.gz -C ${rpm.rootdir} &amp;&amp;
-                            rm -f ${rpm.rootdir}/bin/obclient.tar.gz
+                            rm -f ${rpm.rootdir}/bin/obclient.tar.gz &amp;&amp;
+                            cp ${rpm.rootdir}/lib/libncurses.so.5 /usr/lib64/libncurses.so.5
```
yizhouxw, line 294: why cp , commonly use `ln -s instead` ?

LuckyPickleZZ: Sure, used symbolic link to replace cp.

yhilmare (approved):
lgtm

PR #864, opened 2023-11-21
fix(database-change): OOM may occur when executing database change task with large SQL files

### LINKED ISSUE #409: [Bug]: when the amount of data executed during database change is too large, OOM occurs
### ODC version

4.2.0

### OB version

none

### What happened?

When  the amount of data executed during database change is too large, OOM occurs

OOM should be avoided in any case.

Raise a database change task with a lot sql files.

_No response_

### Cloud

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
module-database change

In current, When user init a database change task and upload SQL script files, the single file size limit is <=256MB, and user can upload multiple files with size 256MB. ODC will read SQL files to a single Java String object. So if the SQL files is very large, then OOM may happened.
The most directly solution is decrease the upload SQL file's size. We will make it as 32 MB at largest in a task.

Fixes #409


yhilmare (approved):
lgtm

PR #872, opened 2023-11-21
fix(security): upgrade the version of some modules to avoid security problems

yhilmare:
#### What type of PR is this?

type-bug

upgrade version of spring-security to 5.7.10
upgrade version of hutool to 5.8.23
upgrade version of netty to 4.1.94.Final
upgrade version of pf4j to 3.10.0

by the way, what does netty use? @yizhouxw 

Fixes #

```docs

```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -44,7 +44,7 @@
 
         <!-- dependency version -->
         <springframework.version>5.3.26</springframework.version>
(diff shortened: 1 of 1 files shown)


smallsheeeep (approved):
lgtm

PR #875, opened 2023-11-21
fix(config): modify bad system configuration

smallsheeeep:
This PR (#864 ) is wrong, and this PR fix it.

### DIFF
--- a/server/odc-migrate/src/main/resources/migrate/common/V_4_2_3_5__modify_upload_file_config.sql
+++ b/server/odc-migrate/src/main/resources/migrate/common/V_4_2_3_5__modify_upload_file_config.sql
@@ -9,17 +9,11 @@ SET
 WHERE
   `key` = 'odc.flow.async.max-upload-file-count';
 
-DELETE FROM
+UPDATE
   `config_system_configuration`
+SET
+  `key` = 'odc.flow.async.max-upload-file-total-size',
+  `value` = '100',
+  `description` = '异步任务最大上传文件数量，默认 100 个'
 WHERE
   `key` = 'odc.flow.async.max-upload-file-total-size}';
-
-INSERT INTO
-  config_system_configuration (`key`, `value`, `description`)
-VALUES(
-    'odc.flow.async.max-upload-file-total-size',
-    '#{32*1024*1024}',
-    '异步任务最大上传文件总大小，单位为字节，默认 32 MB'
-  ) ON DUPLICATE KEY
-UPDATE
-  `value` = '#{32*1024*1024}';
--- a/server/odc-migrate/src/main/resources/migrate/common/V_4_2_3_5__modify_upload_file_config.sql
+++ b/server/odc-migrate/src/main/resources/migrate/common/V_4_2_3_5__modify_upload_file_config.sql
@@ -9,11 +9,17 @@ SET
 WHERE
   `key` = 'odc.flow.async.max-upload-file-count';
 
-UPDATE
+DELETE FROM
   `config_system_configuration`
-SET
-  `key` = 'odc.flow.async.max-upload-file-total-size',
-  `value` = '100',
-  `description` = '异步任务最大上传文件数量，默认 100 个'
 WHERE
   `key` = 'odc.flow.async.max-upload-file-total-size}';
+
+INSERT INTO
+  config_system_configuration (`key`, `value`, `description`)
+VALUES(
+    'odc.flow.async.max-upload-file-total-size',
+    '#{32*1024*1024}',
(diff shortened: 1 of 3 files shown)


server/odc-migrate/src/main/resources/migrate/common/V_4_2_3_5__modify_upload_file_config.sql
```diff
@@ -0,0 +1,19 @@
...
+  `key` = 'odc.flow.async.max-upload-file-count';
+
+UPDATE
+  `config_system_configuration`
+SET
+  `key` = 'odc.flow.async.max-upload-file-total-size',
+  `value` = '100',
+  `description` = '异步任务最大上传文件数量，默认 100 个'
+WHERE
+  `key` = 'odc.flow.async.max-upload-file-total-size}';
```
MarkPotato777, line 19: Perhaps you need to **correct** the key name to `odc.flow.async.max-upload-file-total-size`, otherwise your business code can not load this configuration successfully.

MarkPotato777 (changes requested):

server/odc-migrate/src/main/resources/migrate/common/V_4_2_3_5__modify_upload_file_config.sql
smallsheeeep: yes, the code is depracated

server/odc-server/src/main/resources/data.sql
```diff
@@ -704,4 +704,6 @@ INSERT INTO config_system_configuration ( `key`, `value`, `description` ) VALUES
 INSERT INTO config_system_configuration ( `key`, `value`, `description` ) VALUES( 'odc.iam.auth.alipay.failed-login-lock-timeout-seconds', '600', '账户被锁定时长，默认 600 秒，如果该值小于等于 0 意味着不锁定，改变此参数后需要重启生效' ) ON DUPLICATE KEY UPDATE `id` = `id`;
 
 INSERT INTO config_system_configuration ( `key`, `value`, `description` ) VALUES( 'odc.server.obclient.command-black-list',
- 'connect,conn,edit,nopager,notee,pager,print,prompt,rehash,system,tee,resetconnection', 'Prohibited commands that can be executed by obclient' ) ON DUPLICATE KEY UPDATE `id` = `id`;
\ No newline at end of file
+ 'connect,conn,edit,nopager,notee,pager,print,prompt,rehash,system,tee,resetconnection', 'Prohibited commands that can be executed by obclient' ) ON DUPLICATE KEY UPDATE `id` = `id`;
+
+INSERT INTO config_system_configuration ( `key`, `value`, `description` ) VALUES( 'odc.flow.async.max-upload-file-total-size', '#{32*1024*1024}', '异步任务最大上传文件总大小，单位为字节，默认 32 MB' ) ON DUPLICATE KEY UPDATE `id` = `id`;
```
yizhouxw, line 709: 1. duplicated INSERT stmt in `data.sql`, the key `odc.flow.async.max-upload-file-total-size` has been set to 256MB already, so here the INSERT stmt affect nothing due `ON DUPLICATE KEY UPDATE id=id`.

2. we should not change the default setting for upload files and upload size.  How about add new configuration  items for SQL check, such as `odc.sqlcheck.limit.max-statement-size` , `odc.sqlcheck.limit.max-statements` .

smallsheeeep: 1. The old config key `odc.flow.async.max-upload-file-total-size}` is wrong because the suffix `}` 😂   So I insert into a new correct key `odc.flow.async.max-upload-file-total-size`, and the old config key won't be deleted.
2. We are going to limit the file size for database change task, sql-check is only a stage inner this task. So its better to keep current config key name.

server/odc-migrate/src/main/resources/migrate/common/V_4_2_3_5__modify_upload_file_config.sql
```diff
@@ -0,0 +1,10 @@
+--
+-- Modify upload file config
+--
+UPDATE
+  `config_system_configuration`
+SET
+  `value` = '100',
+  `description` = '异步任务最大上传文件数量，默认 100 个'
```
MarkPotato777, line 8: Use English instead?

smallsheeeep: fine

smallsheeeep:
This PR only corrects dirty data. The problems of database change timeout and OOM caused by oversized SQL files will be fixed later.

MarkPotato777 (approved):
LGTM

yizhouxw (approved):
lgtm

PR #881, opened 2023-11-22
fix(sql-execution): avoid adding rowid when dblink exists

### LINKED ISSUE #214: [Bug]: SQL window executes select sql with dblink without rewriting rowid
### ODC version

4.2.0

### OB version

all mode

### What happened?

image

rowid to be rewrite

_No response_

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

avoid adding rowid when dblink exists

Fixes #214 

```docs

```


LuckyPickleZZ (approved):
LGTM

PR #887, opened 2023-11-22
fix(snippet): snippet body's size is too long to insert into metadb

### LINKED ISSUE #855: [Bug]: Failed to create code snippet when the number of characters in the code snippet is close to 100000
### ODC version

odc423

### OB version

*

### What happened?

02decc44572681105e1c6466e412ae31
5e8122aef84281d0d5e5dcdbab169e11

Created successfully

Create a code snippet with a character count close to 100000

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

the data type of snippet body's corresponding metadb column is text which is limited to 65535. An exception will be thrown by metadb when you attempt to insert a too long data.

This pr add a constraint to the snippet body~

Fixes #855

```docs

```


libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/parser/SqlParser.java
```diff
@@ -27,6 +27,7 @@
 import com.oceanbase.tools.dbbrowser.parser.result.ParseSqlResult;
 import com.oceanbase.tools.sqlparser.obmysql.OBLexer;
 import com.oceanbase.tools.sqlparser.obmysql.OBParser;
+import com.oceanbase.tools.sqlparser.util.TimeoutTokenStream;
```
smallsheeeep, line 30: same as above

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/parser/PLParser.java
```diff
@@ -33,6 +33,7 @@
 import com.oceanbase.tools.sqlparser.oracle.PlSqlLexer;
 import com.oceanbase.tools.sqlparser.oracle.PlSqlParser;
 import com.oceanbase.tools.sqlparser.util.CaseChangingCharStream;
+import com.oceanbase.tools.sqlparser.util.TimeoutTokenStream;
```
smallsheeeep, line 36: import but no usage?

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/parser/SqlParser.java
yhilmare: TimeoutTokenStream is defined in db-browser originally and I move it to the ob-sql-parser, to that the import is changed~

smallsheeeep (approved):
lgtm

PR #888, opened 2023-11-22
fix(sql-execute): move internal rowid to after last select item when rewriting sql

### LINKED ISSUE #742: [Bug]: Execute sql with the same semantics, and the result sets display differently
### ODC version

ODC 4.2.2

### OB version

Any versions of OceanBase.

### What happened?

Execute two semantically identical statements in an sql window but display different result sets.
SQL1:
select a,length(a)as len from TEST_ORDER order by 2 desc;
SQL2:
select a,length(a)as len from TEST_ORDER order by length(a) desc;
The ODC result sets are shown as：
image
image
obclient：
image

ODC result set display is consistent with obclient.

Follow what happened

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

When executing a sql like `select a, b from xxx order by 2`, the result will be sorted by the `b` column .
But in ODC, it would be rewrite into `select ROWID AS "__ODC_INTERNAL_ROWID__" , a, b from xxx order by 2`, which will get a result sorted by the `a` column.

In this PR, we move rowid part to after last select item position. So the sql above would be rewrite as `select a, b, ROWID AS "__ODC_INTERNAL_ROWID__"   from xxx order by 2`. The result will be corrected.

Fixes #742

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/SqlRewriteUtil.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/SqlRewriteUtil.java
@@ -45,7 +45,7 @@ import lombok.extern.slf4j.Slf4j;
 public class SqlRewriteUtil {
 
     private static final String SELECT_ODC_INTERNAL_ROWID_STMT =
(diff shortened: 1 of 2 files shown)


yhilmare (approved):
lgtm

PR #891, opened 2023-11-22
fix(obclient): fix unzip exceptions and symbolic link failed

LuckyPickleZZ:
#### What type of PR is this?
type- bug

This PR is to fix #861 :
1. The `ln -s` command is wrong.
2. The `obclient_xx.tar.gz` was compressed on mac, so there would be some exceptions when de-compressing. Now I repackaged these on a linux server instead of mac.

I have tested it.
image


yhilmare:
@tonghan-gsh try this on arm architecture machine. This doesn't seem to work：
截屏2023-11-22 20 22 29

further more, the `curses` is installed correctly:
A319051F-DF9C-4662-9011-D91D66686E9E

LuckyPickleZZ:
> we may install the library of `curses` based on arm machine

Sure, I have re packaged this on an arm machine. Now we can successfully connect to OBServer using the command line.
image

yhilmare (approved):
lgtm

PR #904, opened 2023-11-23
fix(PL): PL params of MySQL mode are not escaped

### LINKED ISSUE #174: [Bug]: PL parameter not escaped
### ODC version

4.2.0

### OB version

ob-mysql, any version.

### What happened?

As shown in the following figure, the single quotation mark in the input parameter is not escaped

image

Single quote would be escaped.

Execute 
```
create function `print_message` (`message` varchar(255)) returns varchar(255) BEGIN
  declare result  varchar(255);
  SELECT message into result;
  return result;
END;
```
Call it in ODC, then input any param with single quote.

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

When executing PL in MySQL mode, the input params are just wrapped but not escaped. This leads to syntax errors.

Fixes #174


yhilmare (approved):
lgtm

Issue #900, opened 2023-11-23
[Bug]: When exporting data files , the data is not exported and the status shows Completed


Huangxiao-mas:
### ODC version

ODC422bp

### OB version

Oceanbase3240

### What happened?

When exporting data files, the data is not exported and the status shows Completed
![image](https://github.com/oceanbase/odc/assets/59468957/d1725901-8423-444b-8094-13051d765b4a)
![image](https://github.com/oceanbase/odc/assets/59468957/9311a094-0e97-48f4-915d-31ae263b9f0c)

### What did you expect to happen?

execute normally

### How can we reproduce it (as minimally and precisely as possible)?

chat privately

LuckyPickleZZ:
We have set the `PageSize` of ob-dumper to `Integer.MAX_VALUE`, so the task would not be splitted anymore.
But there is another case, when we eporting a table with several partitions, each partition will still be splitted into a standalone task. This issue is still exists.

PR #906, opened 2023-11-23
fix(data-transfer): set page size to avoid ob-dumper splitting files

### LINKED ISSUE #266: [Bug]: multiple data sql files will be generated when exporting data
### ODC version

4.2.0

### OB version

2.2.77-bp12

### What happened?

When exporting data, multiple data sql files will be generated.

When exporting data, only one SQL file is exported.

image

image

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

This PR is to set `DumpParameter.pageSize` into a big value. OB-DUMPER will split transfer task by this param as line counts. 
For example, if we export a table with 10,000 lines by specifing 1000 as  value of `pageSize`  , we will get 10 files after transferring. There files would be named like `xxx.0.0.csv`, `xxx.1.0.csv`, `xxx.2.0.csv`...

So we set this param to `Interger.MAX_VALUE` to avoid ob-dumper splitting fies.

Fixes #266 #900 

FYI, this PR can only avoid issues like above. If you export a table with several partitions, it will still be splitted into several files.. Each file represents one of its partitions.


yhilmare (approved):
lgtm

PR #913, opened 2023-11-24
feat(task-framwork): task executor basic implementation

smallsheeeep:
#### What type of PR is this?
module-task framwork

Basic implementation for task executor
- `Task` interface define the method for all task implementatios, and `SampleTask` is a sample for test, in whcih a list of SQLs will be executed one by one.
- `TaskExecutor` interface define the method for all executor implementations, and `ThreadPoolTaskExecutor` is a sample for test, in which a new task will be executed in a thread pool.
- `JobContextProvider` interface define the method for all `JobContext` providers, and `K8sJobContextProvider` is a implementation in which the job context will be extracted from negotiated system env variables.
- `TaskApplication` is the bootstrap class, it initializes `JobContextProvider` and encapsulate a task, and then using `TaskExecutor` to execut it.

This PR only provides basic implementation and can continue to promote testing. The log, process-based executor and other details will be finished later.

### DIFF
--- a/server/integration-test/src/test/java/com/oceanbase/odc/service/flow/FlowResponseMapperFactoryTest.java
+++ b/server/integration-test/src/test/java/com/oceanbase/odc/service/flow/FlowResponseMapperFactoryTest.java
@@ -135,7 +135,7 @@ public class FlowResponseMapperFactoryTest extends ServiceTestEnv {
         FlowInstanceMapper mapper = FlowInstanceDetailResp.mapper()
                 .withApprovable(id -> false)
                 .withGetConnectionById(id -> TestRandom.nextObject(ConnectionEntity.class))
-                .withGetTaskByFlowInstanceId(id -> Collections.singleton(getValidRandomTaskEntity()))
+                .withGetTaskByFlowInstanceId(id -> Collections.singleton(createTaskEntity()))
                 .withGetUserById(id -> TestRandom.nextObject(UserEntity.class))
                 .withGetExecutionStrategyByFlowInstanceId(
                         id -> Collections.singletonList(TestRandom.nextObject(FlowTaskExecutionStrategy.class)))
@@ -146,14 +146,6 @@ public class FlowResponseMapperFactoryTest extends ServiceTestEnv {
         FlowInstanceDetailResp detailResp = mapper.map(instanceEntity);
 
         Assert.assertEquals(instanceEntity.getId(), detailResp.getId());
-    }
-
-    private TaskEntity getValidRandomTaskEntity() {
-        TaskEntity taskEntity = TestRandom.nextObject(TaskEntity.class);
-        while (taskEntity.getTaskType() == TaskType.PRE_CHECK) {
-            taskEntity = TestRandom.nextObject(TaskEntity.class);
-        }
-        return taskEntity;
     }
 
     @Test
@@ -277,7 +269,9 @@ public class FlowResponseMapperFactoryTest extends ServiceTestEnv {
     }
 
     private TaskEntity createTaskEntity() {
-        return TestRandom.nextObject(TaskEntity.class);
+        TaskEntity entity = TestRandom.nextObject(TaskEntity.class);
+        entity.setTaskType(TaskType.ASYNC);
+        return entity;
     }
 
     private FlowInstanceEntity createFlowInstanceEntity() {
--- a/server/integration-test/src/test/java/com/oceanbase/odc/service/flow/FlowResponseMapperFactoryTest.java
+++ b/server/integration-test/src/test/java/com/oceanbase/odc/service/flow/FlowResponseMapperFactoryTest.java
@@ -135,7 +135,7 @@ public class FlowResponseMapperFactoryTest extends ServiceTestEnv {
         FlowInstanceMapper mapper = FlowInstanceDetailResp.mapper()
                 .withApprovable(id -> false)
                 .withGetConnectionById(id -> TestRandom.nextObject(ConnectionEntity.class))
-                .withGetTaskByFlowInstanceId(id -> Collections.singleton(TestRandom.nextObject(TaskEntity.class)))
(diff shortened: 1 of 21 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/sampletask/SampleTaskResult.java
```diff
@@ -0,0 +1,44 @@
...
+import com.oceanbase.odc.core.flow.model.FlowTaskResult;
+
+import lombok.Data;
+
+/**
+ * @author gaoda.xy
+ * @date 2023/11/22 20:01
+ */
+@Data
+public class SampleTaskResult implements FlowTaskResult {
```
krihy, line 28: avoid use Flow Class

server/odc-server/src/main/java/com/oceanbase/odc/server/OdcServer.java
```diff
@@ -70,6 +72,10 @@ public OdcServer(@Qualifier("metadbMigrate") AbstractMetaDBMigrate metadbMigrate
      * @param args
      */
     public static void main(String[] args) {
+        if (Objects.equals(SystemUtils.getEnvOrProperty("ODC_SERVER_MODE"), "TASK_EXECUTOR")) {
```
krihy: ODC_SERVER_MODE and TASK_EXECUTOR  put in JobEvnConstants

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/sampletask/SampleTaskResult.java
smallsheeeep: You are right. But we no need to create a new class to represent task executing result. `FlowTaskResult` will be change to `TaskResult` in future.

server/odc-server/src/main/java/com/oceanbase/odc/server/OdcServer.java
smallsheeeep: Okay, Done!

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/util/SystemEnvUtil.java
```diff
@@ -0,0 +1,37 @@
...
+
+import java.util.Objects;
+
+import com.oceanbase.odc.common.util.SystemUtils;
+
+/**
+ * @author gaoda.xy
+ * @date 2023/11/24 14:05
+ */
+public class SystemEnvUtil {
```
krihy, line 27: SystemEnvUtil is not necessary, use SystemUtils directly

smallsheeeep: I use customized SystemEnvUtil for null-safe get env.

server/integration-test/src/test/java/com/oceanbase/odc/service/flow/FlowResponseMapperFactoryTest.java
```diff
@@ -148,6 +148,14 @@ public void generateFlowInstanceDetailResp_entityInput_returnDesp() {
         Assert.assertEquals(instanceEntity.getId(), detailResp.getId());
     }
 
+    private TaskEntity getValidRandomTaskEntity() {
+        TaskEntity taskEntity = TestRandom.nextObject(TaskEntity.class);
```
yizhouxw, line 152: why not set to a specifica taskType,
we should avoid involve random object if the random value will affect test result

server/odc-core/src/main/resources/i18n/BusinessMessages_zh_CN.properties
```diff
@@ -662,6 +662,7 @@ com.oceanbase.odc.TaskType.GENERATE_ROLLBACK=备份回滚方案生成
 com.oceanbase.odc.TaskType.PRE_CHECK=预检查
 com.oceanbase.odc.TaskType.EXPORT_RESULT_SET=导出结果集
 com.oceanbase.odc.TaskType.APPLY_PROJECT_PERMISSION=申请项目权限
+com.oceanbase.odc.TaskType.SAMPLE=测试样例
```
yizhouxw, line 665: 样例任务

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/Task.java
```diff
@@ -0,0 +1,76 @@
...
+ * 
+ * @author gaoda.xy
+ * @date 2023/11/22 15:46
+ */
+public interface Task {
+
+    /**
+     * Start current task. This method will be called by {@link TaskExecutor} for fire a task
+     */
+    void start();
```
yizhouxw, line 34: what is the behaivior of `start()` method, block until complete or return straightly ?

server/integration-test/src/test/java/com/oceanbase/odc/service/flow/FlowResponseMapperFactoryTest.java
smallsheeeep: Here is a bug left by history: If the task type is PRE_CHECK, then exception will be thrown in FlowResponseMapperFactory. You are right. Now I fix it by set the task type to ASYNC

server/odc-core/src/main/resources/i18n/BusinessMessages_zh_CN.properties
smallsheeeep: get!

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/Task.java
smallsheeeep: The `start()` method itself in `Task` implementation is blocked until complete. But we can execute this method in a `Executor` (such as `executor.submit(task::start)`) and get empty return straightly.

PR #914, opened 2023-11-24
fix(connect-plugin): move JdbcUrlParser to connect plugin

yhilmare:
#### What type of PR is this?

type-bug

`DefaultJdbcUrlParser` is only used for OceanBase and it will throw an exception when we attempt to use it on MySQL datasource.

I add an interface method to connection-plugin. I use mariadb's impl as the return value of this method.

Fixes #

```docs

```


yizhouxw (approved):
lgtm

yizhouxw (approved):

PR #916, opened 2023-11-27
feat(taskframework): add job scheduler to taskframework

krihy:
#### What type of PR is this?

type-feature
module-taskframework
1.integrate scheduler with task-framework, we can put a job describe by `JobDefinition` to `JobScheduler` to schedule,
and job will be scheduled by trigger config

Fixes #

```docs

```

### DIFF
--- a/script/start-odc.sh
+++ b/script/start-odc.sh
@@ -66,11 +66,13 @@ function check_env_value_set() {
 function init_parameters() {
     echo "init parameters start"
 
-    # init parameters require environment variable
-    check_env_value_set DATABASE_HOST "${DATABASE_HOST}"
-    check_env_value_set DATABASE_PORT "${DATABASE_PORT}"
-    check_env_value_set DATABASE_USERNAME "${DATABASE_USERNAME}"
-    check_env_value_set DATABASE_NAME "${DATABASE_NAME}"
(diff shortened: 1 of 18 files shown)


script/start-odc.sh
```diff
@@ -66,11 +66,13 @@ function check_env_value_set() {
 function init_parameters() {
     echo "init parameters start"
 
-    # init parameters require environment variable
-    check_env_value_set DATABASE_HOST "${DATABASE_HOST}"
-    check_env_value_set DATABASE_PORT "${DATABASE_PORT}"
-    check_env_value_set DATABASE_USERNAME "${DATABASE_USERNAME}"
-    check_env_value_set DATABASE_NAME "${DATABASE_NAME}"
+    if [ -z "${ODC_BOOT_MODE}"  ] || [ "${ODC_BOOT_MODE}" != 'TASK_EXECUTOR' ]; then
```
yhilmare, line 69: I think we should set these env event though we run odc as executor mode cause we may rely on metadb

krihy: yes , ODC_BOOT_MODE will be passed by  k8s pod environment, check_env_value_set  will be skipped when odc start by ODC_BOOT_MODE

krihy: revert this change

smallsheeeep (approved):
LGTM

PR #928, opened 2023-11-27
feat(taskframework): add meta db env and rename method with JobIdentity

krihy:
#### What type of PR is this?

module-taskframework
type-feature

1.rename all method with taskId to JobIdentity for adapter different task source.
2.add environment for odc task executor starter
3.rename DeployModelEnum to TaskRunMode and remove MASTER-SLAVE

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/K8sJobCaller.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/K8sJobCaller.java
@@ -16,8 +16,14 @@
 
 package com.oceanbase.odc.service.task.caller;
 
+import java.util.Map;
+
 import com.oceanbase.odc.common.json.JsonUtils;
+import com.oceanbase.odc.common.util.SystemUtils;
+import com.oceanbase.odc.service.task.constants.JobConstants;
 import com.oceanbase.odc.service.task.constants.JobEnvConstants;
+import com.oceanbase.odc.service.task.enums.TaskRunModeEnum;
+import com.oceanbase.odc.service.task.schedule.JobIdentity;
 
 /**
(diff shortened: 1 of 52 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/K8sJobCaller.java
```diff
@@ -36,17 +42,26 @@ public K8sJobCaller(K8sJobClient client, PodConfig podConfig) {
...
-        String jobName = JobUtils.generateJobName(context.getTaskId());
+        String jobName = JobUtils.generateJobName(context.getJobIdentity());
         PodParam podParam = new PodParam();
-        podParam.getEnvironments().put(JobEnvConstants.TASK_PARAMETER, JsonUtils.toJson(context));
+        Map<String, String> envs = podParam.getEnvironments();
+        envs.put(JobEnvConstants.BOOT_MODE, JobConstants.ODC_BOOT_MODE_EXECUTOR);
+        envs.put(JobEnvConstants.TASK_RUN_MODE, TaskRunModeEnum.K8S.name());
+        envs.put(JobEnvConstants.TASK_PARAMETER, JsonUtils.toJson(context));
+
+        envs.put("DATABASE_HOST", SystemUtils.getEnvOrProperty("ODC_DATABASE_HOST"));
```
smallsheeeep, line 52: Why set database connect info to ENV? Executor no need these envs?

krihy: because dml need meta db store some task middle data

smallsheeeep (approved):
LGTM

PR #968, opened 2023-12-01
fix(database): block built-in databases when auto-sync databases to project

### LINKED ISSUE #955: [Bug]: Built-in databases are auto-sychronized when the datasouce is bound to a project
### ODC version

4.2.0

### OB version

any

### What happened?

If I bind a datasource to a project, then all databases in the datasource are auto-sychronized to project including built-in databases like `test`, `mysql`, `oceanbase`, `information_schema`, `SYS`.

Block built-in databases when auto-sychronizing databases to project：
1. In MySQL mode, block following databases:
- `test`
- `mysql`
- `oceanbase`
- `information_schema`

3. In Oracle mode, block following schemas:
- `SYS`

Bind datasource to a project and then look the database list in project.

_No response_

### Cloud

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
module-database management

When a datasource is bound to a project, then these built-in database such as `oceanbase`, `mysql`, `information_schema`, `test`, `SYS` are inculded when auto-sync databases to project. But these database are usually not used by DEVELOPER. 
In this PR, I make these databases defaultly not auto-sync. And user themselves can config it by alter metaDB.

Fixes #955 

Self-test passed

```docs

```


server/odc-server/src/main/resources/data.sql
```diff
@@ -706,3 +706,6 @@ INSERT INTO config_system_configuration ( `key`, `value`, `description` ) VALUES
 
 INSERT INTO config_system_configuration ( `key`, `value`, `description` ) VALUES( 'odc.server.obclient.command-black-list',
  'connect,conn,edit,nopager,notee,pager,print,prompt,rehash,system,tee,resetconnection', 'Prohibited commands that can be executed by obclient' ) ON DUPLICATE KEY UPDATE `id` = `id`;
+
+INSERT INTO config_system_configuration ( `key`, `value`, `description` ) VALUES( 'odc.connect.database.sync.block-internal-database', 'true',
```
yhilmare, line 710: why use this word 'block' ?

yhilmare:
lgtm, @MarkPotato777 may see this

server/odc-server/src/main/resources/data.sql
smallsheeeep: Means "blocking" the built-in database during synchronization.

MarkPotato777 (approved):
LGTM

PR #970, opened 2023-12-01
feat(osc): lock user is not required when create osc task on ob

krihy:
#### What type of PR is this?

type-feature
module-online schema change
in some data  consistent is not strict ，user can select lock user and kill session  when ob mysql mode

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #987, opened 2023-12-04
fix(data-security): test masking algorithm may lead to security issue

### LINKED ISSUE #986: [Bug]: Trusting user-input segments to test masking algorithm can lead to security problems
### ODC version

4.2.3

### OB version

any

### What happened?

When testing the desensitization effectiveness of the desensitization algorithm, the segmentation configuration of the desensitization algorithm can be modified by modifying the HTTP request body. This may lead to security issues.

Instead of trusting the desensitization algorithm configuration information entered by the user, the desensitization configuration is queried by the ODC back-end service via the ID.

Testing desensitization effects, tampering with request body.

_No response_

### Cloud

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
module-data security

Do not trust user-input `MaskingAlgorithm` when test masking algorithm.

Fixes #986 

```docs

```


yhilmare (approved):
lgtm

PR #994, opened 2023-12-04
feat(sql-execute): unable to obtain locale info in subthread

### LINKED ISSUE #931: [Bug]: The reason for the unavailability of the full link trace is different from the design draft
### ODC version

odc423

### OB version

ob410

### What happened?

image
image

image

Close OB_ Enable_ Show_ The prompt after the trace parameter is different from the design draft

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

In `OdcStatementCallBack`, we use `ErrorCode#getLocalizedMessage` to set `traceEmptyReason`. However, since this operation is performed in a sub thread, `LocaleContextHolder` cannot obtain the correct local value.
So I add a `ServletConfiguration` to to enable inheritance for custom child threads in `LocaleContext`.

I've tested this.

Fixes #931


yhilmare (approved):
👏 great job, lgtm

Issue #975, opened 2023-12-04
[Bug]: 4.2.2版本 日期类型 时分秒都是0


guwenrui:
### ODC version

4.2.2

### OB version

![image](https://github.com/oceanbase/odc/assets/74892585/40dc7dbf-7168-4126-a2c4-af21e9f0fdd0)
![image](https://github.com/oceanbase/odc/assets/74892585/afe01584-c916-466f-98b2-4402446f688e)

### What happened?

4.2.2版本 日期类型 时分秒都是0

### What did you expect to happen?

期望能显示时分秒 如to_char一样
![image](https://github.com/oceanbase/odc/assets/74892585/69bb5d6b-8d65-4acd-a8ff-317de6cb4fa6)

### How can we reproduce it (as minimally and precisely as possible)?

set global nls_date_format='YYYY-MM-DD HH24:MI:SS'

SELECT TO_CHAR(SYSDATE,'YYYY-MM-DD HH24:MI:SS') from dual;

SELECT TO_DATE(SYSDATE,'YYYY-MM-DD HH24:MI:SS') from dual;

sl01388797:
![Image](https://github.com/oceanbase/odc/assets/143492234/bf1a7d27-ee81-4896-a847-d04700b18e25)

PASS

Issue #952, opened 2023-11-29
[Bug]: failed to read hours, minutes and seconds for date type


yhilmare:
### ODC version

4.2.2bp1

### OB version

not related

### What happened?

1. create a table contains date type:
```SQL
create table time_tbl(
    id date
);
insert into time_tbl values(to_date('2023-07-22 01:01:01', 'YYYY-MM-DD HH24:MI:SS'));
```
2. query data:
```SQL
alter session set nls_date_format='YYYY-MM-DD HH24:MI:SS';
select * from time_tbl;
```

3. there is not hours, minutes and seconds info:
<img width="561" alt="截屏2023-11-29 16 01 31" src="https://github.com/oceanbase/odc/assets/22992296/4bf0920d-878e-4337-9389-f81356c71f5d">

### What did you expect to happen?

hours, minutes and seconds info should be displayed.

### How can we reproduce it (as minimally and precisely as possible)?

follow 'what happened'

guwenrui:
I also encountered the same problem. May I ask if there is any solution

LuckyPickleZZ:
> I also encountered the same problem. May I ask if there is any solution

Hello, if you are using ODC V4.2.2,  you can bypass this by setting jdbc param `compatibleOjdbcVersion=8` when creating datasource.

![image](https://github.com/oceanbase/odc/assets/36879124/cb675eff-f509-4fdb-8e14-fd07bc186a69)

Huangxiao-mas:
PASS

PR #1008, opened 2023-12-05
fix(variables): variable updating may lead to sql injection 

### LINKED ISSUE #937: [Bug]: variable management module will lead to sql injection
### ODC version

4.2.2

### OB version

not related

### What happened?

1. begin modify session variable or global variable
截屏2023-11-28 16 26 41
2. click confirm and the front end will call `/api/v1/variables/execute/`, this api is not safe.

call a specific api which will not lead to sql injection

follow 'what happened'

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

there is a api named `/api/v1/variables/execute` which can execute any sql without any security check.

this pr remove this api and add a new api `/api/v2/variables/update` to update session/global variable.

Fixes #937 #975 #952 

```docs

```


smallsheeeep (approved):
LGTM

PR #1014, opened 2023-12-05
fix(sql-rules): cannot add/update any sql rule default values

### LINKED ISSUE #120: [Bug]: can not add any new sql check rules

MarkPotato777:
#### What type of PR is this?

type-bug
module-sql-rule

ODC cannot add/update any SQL rule(sql-check rule and sql-console rule) default values. The root cause is the bad design of database tables and models: we don't distinguish from user rule values and ODC's builtin default rule values -- we record them in the same table. So when ODC wants to change the default value or add a new rule because we don't know if the rule values data has been changed by users or not.

The solution is that we create a new table `regulation_default_rule_applying` to record the default rules created by ODC. And it will not have an `organization_id` column because all organizations will share this. And the business logic has some changes given the new table:

- when we list a specific organization's rules, we merge the value of `regulation_default_rule_applying` and `regulation_rule_applying`: if regulation_rule_applying has a record then return directly, otherwise return the record of `regulation_default_rule_applying`
- when we update a specific rule, we just add a new record in `regulation_rule_applying` and that's all

Fixes #120 

```docs

```

### DIFF
--- a/server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/R4237DefaultRuleApplyingMigrate.java
+++ b/server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/R4237DefaultRuleApplyingMigrate.java
@@ -20,12 +20,15 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
+import javax.persistence.Column;
+import javax.persistence.Convert;
 import javax.sql.DataSource;
 
 import org.apache.commons.collections4.CollectionUtils;
 import org.springframework.transaction.annotation.Transactional;
 
 import com.fasterxml.jackson.core.type.TypeReference;
+import com.oceanbase.odc.common.jpa.JsonListConverter;
 import com.oceanbase.odc.common.util.YamlUtils;
 import com.oceanbase.odc.core.migrate.JdbcMigratable;
 import com.oceanbase.odc.core.migrate.Migratable;
@@ -35,6 +38,11 @@ import com.oceanbase.odc.metadb.regulation.ruleset.MetadataEntity;
 import com.oceanbase.odc.metadb.regulation.ruleset.RuleMetadataRepository;
 import com.oceanbase.odc.service.common.util.SpringContextUtil;
 
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
 
 /**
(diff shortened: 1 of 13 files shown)


server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/R4237DefaultRuleApplyingMigrate.java
```diff
@@ -0,0 +1,155 @@
...
+ * @Date: 2023/12/5 11:06
+ * @Description: []
+ */
+@Slf4j
+@Migratable(version = "4.2.3.7", description = "init default regulation rule applying", repeatable = true,
+        ignoreChecksum = true)
+public class R4237DefaultRuleApplyingMigrate implements JdbcMigratable {
+    private static final String MIGRATE_CONFIG_FILE = "init-config/init/regulation-rule-applying.yaml";
+
+    private DefaultRuleApplyingRepository defaultRuleApplyingRepository;
```
yhilmare, line 60: there is no need making `DefaultRuleApplyingRepository` and `RuleMetadataRepository` as the field value, local value will be fine

server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/R4237DefaultRuleApplyingMigrate.java
```diff
@@ -0,0 +1,155 @@
...
+        this.defaultRuleApplyingRepository = SpringContextUtil.getBean(DefaultRuleApplyingRepository.class);
+        this.ruleMetadataRepository = SpringContextUtil.getBean(RuleMetadataRepository.class);
+        Map<String, MetadataEntity> metadataName2Metadata =
+                ruleMetadataRepository.findAll().stream().collect(Collectors.toMap(MetadataEntity::getName, e -> e));
+        List<InnerDefaultRuleApplying> expected =
+                YamlUtils.fromYaml(MIGRATE_CONFIG_FILE, new TypeReference<List<InnerDefaultRuleApplying>>() {});
+        Map<String, List<DefaultRuleApplyingEntity>> actualRulesetName2RuleApplyings = defaultRuleApplyingRepository
+                .findAll().stream().collect(Collectors.groupingBy(DefaultRuleApplyingEntity::getRulesetName));
+        List<DefaultRuleApplyingEntity> toAdd = new ArrayList<>();
+        for (InnerDefaultRuleApplying expectedApplying : expected) {
```
yhilmare, line 75: expected will be null if the YamlUtils encounter an exception

server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/R4237DefaultRuleApplyingMigrate.java
```diff
@@ -0,0 +1,155 @@
...
+@Migratable(version = "4.2.3.7", description = "init default regulation rule applying", repeatable = true,
+        ignoreChecksum = true)
+public class R4237DefaultRuleApplyingMigrate implements JdbcMigratable {
+    private static final String MIGRATE_CONFIG_FILE = "init-config/init/regulation-rule-applying.yaml";
+
+    private DefaultRuleApplyingRepository defaultRuleApplyingRepository;
+    private RuleMetadataRepository ruleMetadataRepository;
+
+    @Override
+    @Transactional(rollbackFor = Exception.class)
```
yhilmare, line 64: Does this annotation work?

server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/R4237DefaultRuleApplyingMigrate.java
```diff
@@ -0,0 +1,155 @@
...
+                throw new UnexpectedException("rule metadata not found, ruleName: " + expectedApplying.getRuleName());
+            }
+
+            List<DefaultRuleApplyingEntity> actualRuleApplyings =
+                    actualRulesetName2RuleApplyings.get(rulesetName);
+            if (CollectionUtils.isEmpty(actualRuleApplyings)) {
+                toAdd.add(generateNewEntity(expectedApplying, rulesetName, metadataEntity.getId()));
+            } else {
+                Optional<DefaultRuleApplyingEntity> existed = actualRuleApplyings.stream()
+                        .filter(r -> metadataEntity.getId() == r.getRuleMetadataId()).findFirst();
```
yhilmare, line 88: `equals` will be better

server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/R4237DefaultRuleApplyingMigrate.java
```diff
@@ -0,0 +1,155 @@
...
+                if (existed.isPresent()) {
+                    DefaultRuleApplyingEntity actualRuleApplying = existed.get();
+                    if (!isApplyingEquals(expectedApplying, actualRuleApplying)) {
+                        actualRuleApplying.setEnabled(expectedApplying.getEnabled());
+                        actualRuleApplying.setLevel(expectedApplying.getLevel());
+                        actualRuleApplying.setAppliedDialectTypes(expectedApplying.getAppliedDialectTypes());
+                        actualRuleApplying.setPropertiesJson(expectedApplying.getPropertiesJson());
+                        toAdd.add(actualRuleApplying);
```
yhilmare, line 96: how to prevent unique constraint of table 'regulation_default_rule_applying'?

MarkPotato777: ` SimpleRepository#saveAll` could be used in both `insert` and `update` scenarios. If the `@Id` existed in the table, then it will update the record; otherwise it insert a new record.

server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/R4237DefaultRuleApplyingMigrate.java
MarkPotato777: done~

server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/R4237DefaultRuleApplyingMigrate.java
MarkPotato777: I think it will work, why do you think it won't?

server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/R4237DefaultRuleApplyingMigrate.java
MarkPotato777: Okay, verification added.

yhilmare (approved):
lgtm

Issue #683, opened 2023-11-01
[Bug]: failed to reclaim system administrator privileges


yhilmare:
### ODC version

4.2.1

### OB version

not related

### What happened?

1. I login odc as system_admin and assign the role 'system_admin' to a normal user such as 'abcd@email.com'.
<img width="1286" alt="截屏2023-11-01 10 01 31" src="https://github.com/oceanbase/odc/assets/22992296/5ce176c7-d1b1-47aa-976f-2847f525c219">

2. I logout and login odc using 'abcd@email.com' and I create a datasource and a project(create a project is necessary).
<img width="1265" alt="截屏2023-11-01 10 02 17" src="https://github.com/oceanbase/odc/assets/22992296/eaab8866-9c35-4ba7-82f0-5e55859429e9">
<img width="639" alt="截屏2023-11-01 10 03 01" src="https://github.com/oceanbase/odc/assets/22992296/994b9cb9-27fc-4547-bab8-0b1c1460b0cf">

3. I logout and login as the role 'system_admin' and remove the role 'system_admin' from user 'abcd@email.com'.
<img width="643" alt="截屏2023-11-01 10 05 39" src="https://github.com/oceanbase/odc/assets/22992296/6c2e028b-d794-4875-857a-10e5e50f4e1b">

4. I logout and login as 'abcd@email.com', as you can see this account does not have any roles:
<img width="441" alt="截屏2023-11-01 10 07 09" src="https://github.com/oceanbase/odc/assets/22992296/7c96a4a7-8d5e-4edf-bf4b-d11cc6d82e16">

therefore, I expect that this account can not create any datasources but I really can create a datasource:
<img width="1432" alt="截屏2023-11-01 10 15 11" src="https://github.com/oceanbase/odc/assets/22992296/d9a5ed87-c2cb-40a1-a172-192f3fc02fca">

### What did you expect to happen?

the account 'abcd@email.com' does have any roles, so that he/she can not create datasource.

by the way, this account can not create a project cause that he/she does not have any roles, the front end hide the create button which is not so pretty:

<img width="1263" alt="截屏2023-11-01 10 18 40" src="https://github.com/oceanbase/odc/assets/22992296/bec84822-de56-47c7-99b9-67d83fff1fb7">

### How can we reproduce it (as minimally and precisely as possible)?

follow 'what happened'

PR #1019, opened 2023-12-06
fix(permission): user can create datasource without any project and role

### LINKED ISSUE #1013: [Bug]: The user does not belong to any role but has permission to create data sources
### ODC version

ODC 4.2.3

### OB version

-

### What happened?

user：tonghan_1204
The user does not belong to any role but has permission to create data sources
image
Permissions are directly bound to users, not roles.

Users' permissions are controlled by roles.

1.Grant user tonghan_1204 project tonghan administrator rights
2.Create a role that has permission to create data sources
3.Grant this role to the user
4.Revoke this role from the user

_No response_

### Cloud

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
module-user permission

In ODC 4.2.3, if a user join a project with OWNER or DBA role, then ODC  will grant `create` permission of `CONNECTION:*`  to this user. But when the user leave the project, these permission is not reclaimed. So user can create datasource without any project and role. 
In this PR, I fix it by delete the internal grant permission operation.

Fixes #1013 #683 

Self-test passed

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/iam/UserPermissionService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/iam/UserPermissionService.java
@@ -150,6 +150,7 @@ public class UserPermissionService {
                 .collect(Collectors.toList());
     }
 
+    @Deprecated
     @Transactional(rollbackFor = Exception.class)
     @SkipAuthorize("inside method permission check")
     public List<UserPermissionResp> batchUpdateForConnection(@NonNull BatchUpdateUserPermissionsReq req) {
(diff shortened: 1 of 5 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/iam/UserPermissionService.java
```diff
@@ -150,6 +150,7 @@ public List<UserPermissionResp> list(@NonNull String resourceIdentifier) {
                 .collect(Collectors.toList());
     }
 
+    @Deprecated
```
yhilmare, line 153: shall we delete these deprecated methods?

smallsheeeep: Yes, I confirmed it with font-end and deleted these codes.

yhilmare (approved):
great job 😊 ,lgtm

PR #1021, opened 2023-12-06
fix(data-security): create sensitive columns failed due to scanning duplicated columns

### LINKED ISSUE #1005: [Bug]: Scan failed to create sensitive column
### ODC version

odc423

### OB version

*

### What happened?

image

Created successfully

1. Enter the Shilin project
2. Click to scan and add sensitive columns
3. Database selection oracle421, select all databases and recognition rules
4. Click submit

_No response_

### Cloud

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
module-data security

ODC recognize and store sensitive columns by `databaseId`, `tableName` and `columnName`.  And, the `varchar` in meta DB is case-insensitive. So if there are two columns with case-sensitive same `tableName` + `columnName`, then we cannot add both of them as sensitive column. In ODC 4.2.3, ODC will recognize both two columns when scan and this PR fix it.

Fixes #1005 

Self-test passed.

```docs

```


yhilmare (approved):
lgtm

PR #1023, opened 2023-12-06
reactor(taskframework): refactor JobContext remove business column and all task data put in jobData

krihy:
#### What type of PR is this?

type-feature
module-task framework
because job definition need to reschedule if schedule  failed , so all jobs add in task-framework should be restored in db,
after refactor code ,  data in JobDefinition  can mapping to dbms easily

Fixes #

```docs

```


smallsheeeep (approved):
LGTM

PR #1032, opened 2023-12-06
fix(data-transfer): remove permission check when testing connection

### LINKED ISSUE #1030: [Bug]: Ordinary accounts do not have the permission to modify the data source to initiate an export task. When the sys account clicks the test connection, a permission error is reported.
### ODC version

ODC423

### OB version

independent

### What happened?

Ordinary accounts do not have the permission to modify the data source to initiate an export task. When the sys account clicks the test connection, a permission error is reported.
image

normally

chat privately

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

When create a datatransfer task, we could input sys account in task configuration. However, if we use a user without editing data source permissions to perform this operation, an access denied exception will be reported.

In fact , testing connection will not  involve any security issues, so I removed this permission check.

Fixes #1030


LuckyPickleZZ:
close for security issues

PR #1039, opened 2023-12-06
fix(integration): cannot deal with array when parsing json or xml response

### LINKED ISSUE #1006: [Bug]: If the XML contains multiple identical node fields, the external approval configuration cannot find the corresponding fields.
### ODC version

ODC423

### OB version

independent

### What happened?

If the XML contains multiple identical node fields, the external approval configuration cannot find the corresponding fields.
image
image

Tip optimization

chat privately

_No response_

### Cloud

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
module-integration

In ODC 4.2.3, if the response of external system is:
```json
{
    "company": {
        "employee": [
            {
                "id": "1",
                "name": "John Doe",
                "department": "Engineering",
                "skills": {
                    "skill": [
                        "Java",
                        "Python",
                        "SQL"
                    ]
                }
            },
            {
                "id": "2",
                "name": "Jane Smith",
                "department": "Marketing",
                "skills": {
                    "skill": [
                        "Marketing Strategy",
                        "Communication",
                        "Analytics"
                    ]
                }
            }
        ]
    }
}
```
Then we cannot deal with the array, because I have not use `ExpressionParser` in a right way and only support use SpEL like `[company][employee]`. In this PR, I fix it. And you can use SpEL like `company.employee[0].skills.skill[1]` which is easier for common user.

By the way, I refine the `xmlToJson` and `jsonToXml` methods by using `org.json`, and the dependency `jackson-dataformat-xml` is removed. Bucause the XML file root tag will be ignored by `XmlMapper` inner `jackson-dataformat-xml` which is not a right performance.

Fixes #1006 

```docs

```

### DIFF
--- a/server/odc-common/src/main/java/com/oceanbase/odc/common/json/JsonUtils.java
+++ b/server/odc-common/src/main/java/com/oceanbase/odc/common/json/JsonUtils.java
@@ -15,22 +15,22 @@
  */
 package com.oceanbase.odc.common.json;
 
-import java.io.IOException;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+
+import org.json.JSONObject;
+import org.json.XML;
 
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.DeserializationFeature;
-import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.PropertyNamingStrategies;
 import com.fasterxml.jackson.databind.SerializationFeature;
 import com.fasterxml.jackson.databind.type.CollectionType;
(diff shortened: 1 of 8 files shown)


server/odc-common/src/main/java/com/oceanbase/odc/common/json/JsonUtils.java
```diff
@@ -164,12 +164,25 @@ public static String xmlToJson(String xml) {
...
         try {
-            XmlMapper xmlMapper = new XmlMapper();
-            JsonNode jsonNode = xmlMapper.readTree(xml.getBytes());
-            ObjectMapper objectMapper = new ObjectMapper();
-            return objectMapper.writeValueAsString(jsonNode);
-        } catch (IOException e) {
-            log.error("failed to convert xml to json string, reason:{}", e.getMessage());
+            return XML.toJSONObject(xml).toString();
+        } catch (Exception e) {
+            log.error("failed to convert json to xml string, reason:{}", e.getMessage());
```
yhilmare, line 169: log.warn will be better, you may change this in your next pr

PR #1052, opened 2023-12-07
fix(dlm):task cannot be executed due to insufficient connections

### LINKED ISSUE #1033: [Bug]: Data archiving, the task is in a waiting state after half execution
### ODC version

ODC423

### OB version

independent

### What happened?

Data archiving, the task is in a waiting state after half execution
image
image

normally

chat privately

_No response_

### Cloud

_No response_

guowl3:
#### What type of PR is this?
type-bug

The current task connection pool size is determined by the number of task threads, but in scenarios where the number of threads is configured to be small, there may be a problem of insufficient connections causing the task to freeze.

Fixes #1033

```docs

```


yhilmare (approved):
lgtm

PR #1053, opened 2023-12-07
fix(partition-plan): create partition plan task failed in obmysql 1479

### LINKED ISSUE #969: [Bug]: Partition plan execution failed
### ODC version

4.2.2

### OB version

1.4.7.9

### What happened?

Partition plan execution failed

Partition plan creates partitions normally

Create a partition plan job and review task execution records.

_No response_

### Cloud

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug

The partition plan uses SchemaAccessor to obtain range partition info. But in obmysql 1479, the query SQL is wrong because `oceanbase.__all_part` doesn't have `list_val` column(higher version does).
This PR corrects the partition info query SQL by deleting the non-existing column. It only affects the partition plan given that it uses Parser to obtain partition info in other scenery. And partition plan doesn't care about the list partition info so deleting this column will be fine.

Fixes #969 

```docs

```


guowl3 (approved):
LGTM

PR #1056, opened 2023-12-07
feat(data-transfer): support log throughput of datatransfer

LuckyPickleZZ:
#### What type of PR is this?
type-feature

This PR mainly does these things:
1. I added a `ThroughputReporter` to log throughput of MySQLDataTransferJob. When a new object is started, the job will publish a `ObjectStartEvent `. This event is going to be caught in the reporter, which will calculate the amount of archived data and incremental data, in order to print the transfer speed.
2. If the `maxDumpSize` is set, we will check the directory size every second. If the directory size exceeds max size, the job will be shutdown.
3. If user specified 'skippedDataType', we will filter the columns by this field.
4. Fix some bugs.

Fixes #

```docs

```

### DIFF
--- a/server/plugins/task-plugin-mysql/src/main/java/com/oceanbase/odc/plugin/task/mysql/datatransfer/MySQLDataTransferJob.java
+++ b/server/plugins/task-plugin-mysql/src/main/java/com/oceanbase/odc/plugin/task/mysql/datatransfer/MySQLDataTransferJob.java
@@ -284,7 +284,7 @@ public class MySQLDataTransferJob implements DataTransferJob {
     }
 
     private void initSchedules(ScheduledExecutorService executor) {
-        if (baseConfig.isCompressed()) {
+        if (baseConfig.isCompressed() || baseConfig.getTransferType() == DataTransferType.EXPORT) {
             ThroughputReporter reporter = new ThroughputReporter();
             publisher.addEventListener(reporter);
             executor.scheduleAtFixedRate(reporter, 5, 5, TimeUnit.SECONDS);


server/plugins/task-plugin-mysql/src/main/java/com/oceanbase/odc/plugin/task/mysql/datatransfer/MySQLDataTransferJob.java
```diff
@@ -268,15 +283,37 @@ private synchronized static void unzipDataXToWorkingDir(File workingDir) throws
...
+            executor.scheduleAtFixedRate(() -> {
+                File dataDir = new File(workingDir, "data");
+                long size;
+                if (dataDir.exists()
+                        && (size = FileUtils.sizeOfDirectory(dataDir)) >= baseConfig.getMaxDumpSizeBytes()) {
+                    LOGGER.info("Exported size {} exceeds {}, transfer will stop.", size,
+                            baseConfig.getMaxDumpSizeBytes());
+                    schemaJobs.forEach(AbstractJob::cancel);
+                    dataJobs.forEach(AbstractJob::cancel);
+                    executor.shutdown();
```
yhilmare, line 302: do not shutdown executor in its thread~

yhilmare (commented):
add a read me doc to guide how to build datax

server/plugins/task-plugin-mysql/src/main/java/com/oceanbase/odc/plugin/task/mysql/datatransfer/MySQLDataTransferJob.java
LuckyPickleZZ: done, fixed this

yhilmare (approved):
lgtm

PR #1061, opened 2023-12-07
fix(project): could delete users who are currently joining projects

### LINKED ISSUE #922: [Bug]: Deleting a user requires verifying whether the user is bound to the item
### ODC version

ODC 4.2.0

### OB version

-

### What happened?

If a user is both the administrator and dba of the project, after I delete this user, I can no longer operate the database in the project.

If you want to delete the user, you should verify whether it is bound to the item.

1.create a user u1;
2.create a item p1,set u1 as administrator and dba;
3.Add a database to the project p1;
4.delete user u1;

_No response_

### Cloud

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug
module-project

When deleting users, we do not verify if they are project members which may cause the problem that some project doesn't have any OWNER after deletion, then nobody could assign a new OWNER to this project because only OWNER could assign project roles.
This PR adds a pre-delete-user hook to detect if the deleting user still joins some project, if yes then reject the operation.

Fixes #922 

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/config/HookConfiguration.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/config/HookConfiguration.java
@@ -31,7 +31,7 @@ import com.oceanbase.odc.core.shared.constant.AuditEventAction;
 import com.oceanbase.odc.core.shared.constant.ErrorCodes;
 import com.oceanbase.odc.core.shared.constant.ResourceRoleName;
 import com.oceanbase.odc.core.shared.constant.ResourceType;
-import com.oceanbase.odc.core.shared.exception.UnsupportedException;
+import com.oceanbase.odc.core.shared.exception.BadRequestException;
 import com.oceanbase.odc.metadb.collaboration.ProjectEntity;
 import com.oceanbase.odc.metadb.collaboration.ProjectRepository;
 import com.oceanbase.odc.service.collaboration.project.ProjectService;
@@ -100,7 +100,7 @@ public class HookConfiguration {
             String errorMessage = String.format(
                     "Approval flow config id=%s cannot be %s because it has been referenced to following risk level: {%s}",
                     flowConfigId, AuditEventAction.DELETE_FLOW_CONFIG, names);
-            throw new UnsupportedException(ErrorCodes.CannotOperateDueReference,
+            throw new BadRequestException(ErrorCodes.CannotOperateDueReference,
                     new Object[] {AuditEventAction.DELETE_FLOW_CONFIG.getLocalizedMessage(),
                             ResourceType.ODC_APPROVAL_FLOW_CONFIG.getLocalizedMessage(), "name", names,
                             ResourceType.ODC_RISK_LEVEL.getLocalizedMessage()},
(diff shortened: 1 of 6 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/config/HookConfiguration.java
```diff
@@ -91,4 +108,20 @@ private void approvalFlowConfigUsageCheck(Long flowConfigId) {
         }
     }
 
+    private void projectReferenceCheck(Long userId, Long organizationId) {
```
smallsheeeep, line 111: As long as the current user is added to any of the project, they cannot be deleted. Would this be too difficult to delete a user.

server/odc-service/src/main/java/com/oceanbase/odc/config/HookConfiguration.java
```diff
@@ -91,4 +108,20 @@ private void approvalFlowConfigUsageCheck(Long flowConfigId) {
...
+        Map<Long, Set<ResourceRoleName>> projectId2ResourceRoleNames =
+                resourceRoleService.getProjectId2ResourceRoleNames(userId);
+        if (projectId2ResourceRoleNames.size() == 0) {
+            return;
+        }
+        Map<Long, ProjectEntity> id2Project =
+                projectRepository.findAllByOrganizationId(organizationId).stream().filter(p -> p.getArchived() == false)
+                        .collect(Collectors.toMap(ProjectEntity::getId, p -> p));
+        throw new UnsupportedException(
```
smallsheeeep, line 120: Error messages should ideally be internationalized, see `ErrorCodes.CannotOperateDueReference`

smallsheeeep (changes requested):

server/odc-service/src/main/java/com/oceanbase/odc/config/HookConfiguration.java
MarkPotato777: I changed the condition, now only if the user is the owner or dba then it cannot be deleted.

server/odc-service/src/main/java/com/oceanbase/odc/config/HookConfiguration.java
MarkPotato777: okay, done~

server/odc-service/src/main/java/com/oceanbase/odc/config/HookConfiguration.java
```diff
@@ -91,4 +108,31 @@ private void approvalFlowConfigUsageCheck(Long flowConfigId) {
...
+                projectRepository.findAllByOrganizationId(organizationId).stream().filter(p -> p.getArchived() == false)
+                        .collect(Collectors.toMap(ProjectEntity::getId, p -> p));
+        String names = projectId2ResourceRoleNames.keySet().stream().map(id -> id2Project.get(id).getName())
+                .collect(Collectors.joining(", "));
+
+        String errorMessage = String.format(
+                "User id=%s cannot be deleted because it has been referenced to following project: {%s}", userId,
+                names);
+
+        throw new UnsupportedException(ErrorCodes.CannotOperateDueReference,
```
yhilmare, line 130: BadRequestException will be better？
UnsupportedException means that a feature is not supported yet

MarkPotato777: done

smallsheeeep (approved):

PR #1071, opened 2023-12-08
fix(project): transaction timeout when transfer too many databases or add too many users into projects

### LINKED ISSUE #151: [Bug]: transaction timeout error while bind more than 300 databases into project
### ODC version

4.2.0

### OB version

not related

### What happened?

transaction timeout error while bind more than 300 databases into project.

the MetaDB unit config was 1C4G, may not sufficient.

bind databases to project success

bind more than 300 databases into project

there also exists slow sql for table `SPRING_SESSION`, update operation

### Cloud

_No response_

MarkPotato777:
#### What type of PR is this?
type-bug
module-project

When adding too many(like over 300) users into the projects or transferring too many databases into one project at one time, it will fail as the transaction times out. The root cause is that we use Spring Data JPA `saveAll()` which has efficiency issues when operating a huge amount of records.
The solution is that we use JdbcTemplate to batch update/insert to reduce network IO.

Fixes #151 

```docs

```


yizhouxw (approved):
lgtm

yizhouxw (dismissed):
lgtm

yhilmare (approved):
lgtm

PR #1072, opened 2023-12-08
fix(osc): lock ob mysql user failed when host with ip limited

krihy:
#### What type of PR is this?

module-onlineSchemeChange
type-feature

1.fix lock ob mysql account when host with ip segment , eg:
```
MySQL [(none)]> select user,host from mysql.user where user = 'test1';
+---------+-------+
| user    | host  |
+---------+-------+
| test1 | 11.% |
+---------+-------+
1 row in set (0.05 sec)
```    
lock user sql will like :
```
alter user  test1@'11.%' account lock;
``` 
2. query user lock status and filter user should is unlocked
3. fix osc unit
4. add sysTenantUser and sysTenantPassword in `TestDBType` for local unit test

Fixes #

```docs

```


libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoGreaterThan5740SchemaAccessor.java
```diff
@@ -164,10 +164,10 @@ public void switchDatabase(String schemaName) {
     @Override
     public List<DBObjectIdentity> listUsers() {
         MySQLSqlBuilder sb = new MySQLSqlBuilder();
-        sb.append("SELECT DISTINCT user FROM mysql.user");
+        sb.append("SELECT user,host FROM mysql.user");
```
yizhouxw, line 167: what is the behavior if there exist multile records for same user ?

yizhouxw (approved):
lgtm

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoGreaterThan5740SchemaAccessor.java
krihy: 1. add DISTINCT for remove duplicate record if db user and host is same
2. web will show multi db user record with different host, and odc user can select any of them to be locked

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoGreaterThan5740SchemaAccessor.java
```diff
@@ -164,10 +164,10 @@ public void switchDatabase(String schemaName) {
     @Override
     public List<DBObjectIdentity> listUsers() {
         MySQLSqlBuilder sb = new MySQLSqlBuilder();
-        sb.append("SELECT DISTINCT user FROM mysql.user");
+        sb.append("SELECT DISTINCT user,host FROM mysql.user");
         return jdbcOperations.query(sb.toString(), (rs, rowNum) -> {
             DBObjectIdentity dbUser = new DBObjectIdentity();
-            dbUser.setName(rs.getString(1));
+            dbUser.setName(rs.getString(1) + "@'"+ rs.getString(2)+"'");
```
yhilmare, line 170: we may add a field to hold host rather than append host message to the user's name

krihy: we can regard 'user'%'host' as a whole username, like : `CREATE USER 'username'@'hostname' IDENTIFIED BY 'password'`;

yhilmare: it is different, what do do is to change the behavior and that is forbidden unless it is necessary~

krihy: done, instand of query user from OscDBAccessor

yhilmare (approved):
lgtm

PR #1073, opened 2023-12-08
fix(dlm):displays incomplete information after editing

guowl3:
#### What type of PR is this?
type-bug

Fixed the problem of incomplete information display after task editing.

Fixes #

```docs

```


krihy (approved):
lgtm

PR #1078, opened 2023-12-08
fix(pl): no sys_refcursor shown in return type select panel when creating function

### LINKED ISSUE #240: [Bug]: Create a function return type add a sys_refcursor type drop-down options
### ODC version

ODC 4.2.0

### OB version

Any version of OceanBase

### What happened?

Function now supports sys_refcursor return type
image

Added sys_refcursor type option

create function and return type is sys_refcursor

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

add `sys_refcursor` data type for oracle mode

Fixes #240 

```docs

```


LuckyPickleZZ (approved):
LGTM

PR #1079, opened 2023-12-08
fix(sql-rule): the sql type 'desc' does not work in the allow-sql-types rule

### LINKED ISSUE #775: [Bug]: SQL console specifications have not taken effect after modifying
### ODC version

4.2.1

### OB version

null

### What happened?

image
image

.

.

.

### Cloud

.

MarkPotato777:
#### What type of PR is this?

type-bug

The parser doesn't distinguish `desc/describe` and `explain` types given that they are synonyms. However, the sql console rule 'allow-sql-types' involves both `desc` and `explain`. The result is that the `desc` will not work forever because the parser considers `desc` as `explain`.

This PR distinguish `desc/describe` with `explain`.

Fixes #775 

```docs

```


yhilmare (approved):
lgtm

PR #1089, opened 2023-12-08
fix(ticket): list all tickets returns empty in individual space

### LINKED ISSUE #893: [Bug]: failed to get orders in private zone
### ODC version

4.2.3

### OB version

not related

### What happened?

I can not see any orders in private zone:
截屏2023-11-22 19 59 40

but I can get orders on condition of 'created by me':
 
截屏2023-11-22 20 00 50

I can get orders.

follow what happened

_No response_

### Cloud

_No response_

MarkPotato777:
#### What type of PR is this?
type-bug

Listing all tickets returned empty in the individual space, because when listing current user project roles, we do not query by organization_id, then in individual space we obtain the user's team space's resource roles which cause the logic error.

When listing resource roles, add organization_id into query condition.

Fixes #893 

```docs

```


yhilmare (approved):
lgtm

PR #1095, opened 2023-12-11
feat(data-masking): prohibit data-masking for native MySQL datasource

### LINKED ISSUE #1054: [Feature]: Prohibit setting desensitization columns for MySQL data sources.

Since the support for data desensitization for MySQL data sources is currently incomplete, this capability will be disabled  for security reasons.

Since the support for data desensitization for MySQL data sources is currently incomplete, this capability will be disabled first for security reasons.

### Additional context

_No response_

smallsheeeep:
#### What type of PR is this?
type-feature
module-data security

In ODC 4.2.3, we support exporting data for native MySQL but data masking is not supported yet. So, we should prohibit the use of data-masking for native MySQL datasource by prohibitting user to create sensitive columns. This PR just do these.

Close #1054 

```docs

```


yhilmare (approved):
lgtm

PR #1096, opened 2023-12-11
fix(obclient): do not create os user when it already exists

### LINKED ISSUE #972: [Bug]: only create os user when it doesn't exist
### ODC version

4.2.2

### OB version

none

### What happened?

When ODC opening a obclient window, it would  call `useradd` command to create a os user .This is for security issues and for resource isolation. 
However , this user would not be deleted. So the logs of the operating system will have some records of user creation failures.

This command should be called only when the user does not exist.

Open the command line multiple times and then check the logs.

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

When ODC opening a obclient window, it would call useradd command to create a os user .This is for security issues and for resource isolation.
However , this user would not be deleted. So the logs of the operating system will have some records of user creation failures.

So we will query whether this user already exists by grep file `/etc/passwd`. All users have read access to this file. If it exists, we won't create repeatedly.

Fixes #972


yhilmare (approved):
lgtm

PR #1100, opened 2023-12-11
feat(taskframework):modify job definition add jobClass and remove trigger and sourceId

krihy:
#### What type of PR is this?

module-taskframework
type-feature

1. modify job definition add jobClass and remove trigger and sourceId
2. add job_job table to save all tasks scheduled by taskframework
3. add hook ResultHandleService for business to update status and progress

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/TaskFactory.java
```diff
@@ -16,39 +16,21 @@
...
-                case SHADOWTABLE_SYNC:
-                case EXPORT_RESULT_SET:
-                case ONLINE_SCHEMA_CHANGE:
-                case APPLY_PROJECT_PERMISSION:
-                default:
-                    throw new UnsupportedException("Unsupported task type.");
+    public static Task create(String jobClass) {
+        try {
+            Class<?> c = Class.forName(jobClass);
+            if (c.isAssignableFrom(Task.class)) {
```
smallsheeeep, line 28: may `!c.isAssignableFrom(Task.class)`

krihy: done

PR #1102, opened 2023-12-11
fix(full-link-trace): no tags and references in downloaded json file

### LINKED ISSUE #1002: [Bug]: the tags of downloaded full-link-trace json file are empty
### ODC version

4.2.3

### OB version

none

### What happened?

When downloading a json file of full-link-trace, the tags are always empty.

There should be some value in tags.

-

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

When downloading a json file of full-link-trace, the tags are always empty.
This is because some internal classes in `JaegerConverter` do not have getter methods.

Fixes #1002


yhilmare (approved):
lgtm

PR #1104, opened 2023-12-11
enhancement(pl): add async result acquisition api for pl execution

### LINKED ISSUE #1087: [Enhancement]: asynchronously calling functions/procedures

as title saied

### Why is this needed?

cloud api gateway has timeout, we should call pl asynchronously to fix this problem.

yhilmare:
#### What type of PR is this?

type-enhancement

1. rename all interfaces starting with `api/v1/pl` to `/api/v2/pl`
2. remove api `/api/v1/pl/callProcedure` and add `/api/v2/pl/procedure/{sid}/call`, input parameter is same with older one, output is string type.
3. remove api `/api/v1/pl/callFunction` and add `/api/v2/pl/function/{sid}/call`, input parameter is same with older one, output is string type.
4. add `/api/v2/pl/function/{sid}/getResult` and `/api/v2/pl/procedure/{sid}/getResult` to get calling result of function/procedure.

Fixes #1087 

```docs

```


yizhouxw (approved):
lgtm

PR #1110, opened 2023-12-11
fix(osc): osc log is not show totally and  flow task is done unnormally

krihy:
#### What type of PR is this?

type-bug
module-online schema change
1. after`OnlineSchemaChangeFlowableTask` started and executing,  thread  should be wait unit all osc schedule task completed, on `onProgressUpdate` method persist to update task status,   thread executing in `onProgressUpdate` will notify the task started thread.
2. when task percentage or status is changed, then update task
3. fix finalTableDdl for alter sql type

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/OnlineSchemaChangeFlowableTask.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/OnlineSchemaChangeFlowableTask.java
@@ -86,7 +86,6 @@ public class OnlineSchemaChangeFlowableTask extends BaseODCFlowTaskDelegate<Void
     private volatile TaskStatus status;
     private volatile long scheduleId;
     private volatile long creatorId;
-    private volatile long flowTaskId;
     private volatile long organizationId;
     private volatile boolean continueOnError;
     private volatile double percentage;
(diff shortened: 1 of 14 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/OnlineSchemaChangeFlowableTask.java
```diff
@@ -79,14 +79,17 @@ public class OnlineSchemaChangeFlowableTask extends BaseODCFlowTaskDelegate<Void
     @Autowired
     private OrganizationService organizationService;
 
+    private final Object LOCK = new Object();
```
yhilmare, line 82: lock, LOCK is only for static field

krihy: instance lock is only monitor current object,  every osc task relational a new OnlineSchemaChangeFlowableTask object

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/OnlineSchemaChangeFlowableTask.java
```diff
@@ -117,6 +120,12 @@ protected Void start(Long taskId, TaskService taskService, DelegateExecution exe
                 taskHandler.start(scheduleId, tasks.get(0).getId());
                 log.info("Successfully start schedule task with id={}", tasks.get(0).getId());
             }
+            synchronized (LOCK) {
+                log.info("Wait task {} to be terminated.", taskId);
+                LOCK.wait();
```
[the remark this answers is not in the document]
krihy: we set `oscTaskExpiredAfterSeconds` default value is 5*24*3600 seconds in DefaultOnlineSchemaChangeTaskHandler, which represent osc task will be canceled automatic if it's not completed

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/OnlineSchemaChangeFlowableTask.java
yhilmare: I mean that 'LOCK' this upper case name is usually used for static field, nor for local field.

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/OnlineSchemaChangeFlowableTask.java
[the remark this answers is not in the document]
yhilmare: we can not rely on this timeout settings~

yhilmare (approved):
lgtm

PR #1119, opened 2023-12-11
fix(ticket): approvers viewing shadow table sync ticket fails after the ticket is approved/rejected

### LINKED ISSUE #1048: [Bug]: After successfully applying to execute the shadow table, an error message appears when viewing tasks for the admin user
### ODC version

odc423

### OB version

*

### What happened?

image

Can be viewed

1. Performing shadow table synchronization on low permission accounts
2. Agree with Damin account
3. View shadow table synchronization execution tasks

_No response_

### Cloud

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug

Shadow table sync tickets should be seen by creators and approvers but after the tickets are approved by the approval, it would occur an error when the approver clicks the ticket again.
The root cause is that when we query approval candidates, we use 'EXECUTING' status to filter the approval instances, but not filter the 'APPROVED' approval instances.
This PR changed the method signature and could list both approving and approved approval instances.

Fixes #1048 

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/ApprovalPermissionService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/ApprovalPermissionService.java
@@ -17,10 +17,7 @@ package com.oceanbase.odc.service.flow;
 
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -31,7 +28,6 @@ import org.apache.commons.collections4.CollectionUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 
-import com.oceanbase.odc.common.lang.Pair;
 import com.oceanbase.odc.core.authority.util.SkipAuthorize;
 import com.oceanbase.odc.metadb.flow.UserTaskInstanceCandidateEntity;
 import com.oceanbase.odc.metadb.flow.UserTaskInstanceCandidateRepository;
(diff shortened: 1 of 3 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/flow/ApprovalPermissionService.java
```diff
@@ -83,60 +76,6 @@ public class ApprovalPermissionService {
...
     private UserResourceRoleRepository userResourceRoleRepository;
 
-    /**
-     * 获取审批节点可审批用户集合
-     *
-     * @param approvalInstanceId
-     * @param enabled 关联用户或角色是否启用，true -> 角色用户启用; false -> 角色用户禁用; null -> 启用禁用用户均筛选出来
-     * @return list of {@link User}
-     */
-    public Set<User> getCandidates(@NonNull Long approvalInstanceId, Boolean enabled) {
```
MarkPotato777, line 93: Delete the deprecated code without any reference. Same as follows.

yhilmare:
so the related issue is not only for shadow table task but also all types of task?

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/ApprovalPermissionService.java
```diff
@@ -256,18 +130,27 @@ public List<UserTaskInstanceEntity> listApprovableExternalInstances() {
                 .collect(Collectors.toList());
     }
 
+
+    public Map<Long, Set<UserEntity>> getApproverByFlowInstanceIds(@NonNull Collection<Long> flowInstanceIds) {
+        return getUsersByFlowInstanceIdsAndStatus(flowInstanceIds, FlowNodeStatus.getExecutingAndFinalStatuses());
```
yhilmare, line 135: prefer `return getUsersByFlowInstanceIdsAndStatus(flowInstanceIds, Collectors.sington(FlowNodeStatus.EXECUTING));`

MarkPotato777:
> so the related issue is not only for shadow table task but also all types of task?

No, it's just for shadow table sync tasks which is a kind of special. Shadow table sync tasks are not flowable tasks, and they have a separate table to record. Here it uses related flow_instance_id to query all approvers to make them have authority to view the shadow table sync results.

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/ApprovalPermissionService.java
MarkPotato777: I need exactly 'executing' and 'final' status here for querying those approvers who are approving, **approved/rejected**, and this is where the bug is located.

yhilmare (approved):
lgtm

PR #1135, opened 2023-12-12
fix(session): session creation will fail when the oracle schema name is lowercase

### LINKED ISSUE #1092: [Bug]: Failed to create an sql window for oracle schema database name in lower case
### ODC version

ODC 4.2.3

### OB version

OB 4.2.1

### What happened?

Failed to create an sql window for oracle schema database name in lower case
image

Databases with lower-case names can also create sql Windows.

datasource:oracle42222221
database:test新建
Open a sql window

_No response_

### Cloud

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug

When the Oracle schema name is in lowercase, it will fail to create sessions.
The root cause is that when we create sessions, we do not quote the schema name and then the connection plugin will automatically uppercase it, so it will fail when creating sessions using JDBC.
This PR quotes schema names in Oracle mode.

Fixes #1092

```docs

```


yhilmare (dismissed):
lgtm

yhilmare (approved):
lgtm

PR #1136, opened 2023-12-12
feat(partition-plan):support setting scheduling strategy

guowl3:
#### What type of PR is this?
type-feature

The partition plan does not support configuring scheduling time currently. For OB data sources, deleting partitions will cause global index failure. If executed during peak business periods, it may cause performance problems. Therefore, it is necessary to support custom scheduling time to ensure that partition deletion operations are performed during low business periods.

Fixes #

```docs

```


krihy (commented):
lgtm

yhilmare (approved):
lgtm

PR #1139, opened 2023-12-12
fix(result-set-export): there is no data in exported xlsx file

### LINKED ISSUE #1129: [Bug]: Export result set in Excel format, there is no content in the exported file
### ODC version

odc423

### OB version

*

### What happened?

image
image

There is data in the exported result set

Export result set in Excel format, export

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

If export result set in excel format, you will get a xlsx file with no data.
This is caused by setting file path incorrectly.

Fixes #1129


yhilmare (approved):
lgtm

PR #1145, opened 2023-12-12
fix(jdbc): full link trace leads to OOM exception

### LINKED ISSUE #1105: [Bug]: The desktop version does not have the sendConnectionAttributes parameter
### ODC version

odc423

### OB version

*

### What happened?

image

Test link is normal

Desktop version attribute configuration sendConnectionAttributes parameter, click on the test link

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

There is a bug of oceanbase-client 2.4.7.1. If we set `enableFullLinktrace` to true , there may be cause OOM exception.

Fixes #1105

### DIFF
--- a/server/plugins/connect-plugin-ob-mysql/src/main/java/com/oceanbase/odc/plugin/connect/obmysql/OBMySQLConnectionExtension.java
+++ b/server/plugins/connect-plugin-ob-mysql/src/main/java/com/oceanbase/odc/plugin/connect/obmysql/OBMySQLConnectionExtension.java
@@ -117,9 +117,8 @@ public class OBMySQLConnectionExtension implements ConnectionExtensionPoint {
     }
 
     protected Map<String, String> appendDefaultJdbcUrlParameters(Map<String, String> jdbcUrlParams) {
-        if (!jdbcUrlParams.containsKey("enableFullLinkTrace")) {
-            jdbcUrlParams.put("enableFullLinkTrace", "true");
-        }
+        // there is a bug of oceanbase-client 2.4.7.1, setting this to true may cause OOM exception
+        jdbcUrlParams.put("enableFullLinkTrace", "false");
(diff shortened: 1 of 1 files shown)


yhilmare (approved):
lgtm

server/plugins/connect-plugin-ob-mysql/src/main/java/com/oceanbase/odc/plugin/connect/obmysql/OBMySQLConnectionExtension.java
```diff
@@ -117,9 +117,8 @@ protected String getJdbcUrlParameters(Map<String, String> jdbcUrlParams) {
     }
 
     protected Map<String, String> appendDefaultJdbcUrlParameters(Map<String, String> jdbcUrlParams) {
-        if (!jdbcUrlParams.containsKey("enableFullLinkTrace")) {
-            jdbcUrlParams.put("enableFullLinkTrace", "true");
-        }
+        // there is a bug of oceanbase-client 2.4.7.1, setting this to true may cause OOM exception
```
yizhouxw, line 120: will it leads to full link trace not works? when wil oceanbase-client fix this OOM issue

LuckyPickleZZ: No, this setting only affects the stage of jdbc. The remaining stages of obproxy and observer can still be displayed normally.

LuckyPickleZZ: They will fix this in 2.4.8.

PR #1147, opened 2023-12-12
fix(iam): users need re-login to access the individual space after they are granted for individual_space

### LINKED ISSUE #671: [Bug]: The automatic authorization is granted to the user individual_space role, when the user logs in for the first time is individual
### ODC version

ODC 4.2.2

### OB version

Any version of OceanBase.

### What happened?

The role that the user individual_space is automatically authorized. If the user logs in for the first time, he still has no personal space right.Should login again.
image

This permission is granted upon the first successful login.

1.Create an automatic authorization rule that assigns the individual_space role to a successful first login.
2.login in.

_No response_

### Cloud

_No response_

MarkPotato777:
#### What type of PR is this?
type-bug

If the individual_space role was granted to the user, the user must log out and re-login to access the individual space.
The root cause is the individual space created at the user migration step, and the migrate stop trigger while the user login. For the first-time user login, the migration step is skipped due user having no individual space privilege, and the next migration will not be triggered until next time the user login.
This PR changed the migration logic: we just always create the individual space, regardless if the current user has individual space privilege.

Fixes #671 

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/collaboration/DefaultOrganizationResourceMigrator.java
```diff
@@ -100,21 +98,16 @@ public void migrate(@NonNull User user) {
...
-                boolean hasIndividualPermission =
-                        verticalPermissionValidator.implies(individual, Arrays.asList("read", "update"), team.getId());
-                Optional<Organization> organization = organizationService.getByOrganizationTypeAndUserId(
-                        OrganizationType.INDIVIDUAL, user.getId());
-                if (!organization.isPresent() && hasIndividualPermission) {
-                    log.info("Individual organization not found, start to initialize for userId={}", user.getId());
-                    OrganizationEntity saved = individualOrganizationMigrator.migrate(user);
-                    log.info("Initialized individual organization successfully for userId={}", user.getId());
-                }
+                organizationService.getByOrganizationTypeAndUserId(
```
yhilmare, line 101: the return value has never acquired, we may do as follow:
```Java
Optional< Organization> optional = organizationService.getByOrganizationTypeAndUserId(OrganizationType.INDIVIDUAL, user.getId());
if (!optional.isPresent()) {
     log.info("Individual organization not found, start to initialize for userId={}",user.getId());
     OrganizationEntity saved = individualOrganizationMigrator.migrate(user);
     log.info("Initialized individual organization successfully for userId={}, organizationId={}", user.getId(), saved.getId());
}
```

this way may be much more clearer.

by the way, 'verticalPermissionValidator' is never acquired, we'd better delete it.

MarkPotato777: okay, done

yhilmare (approved):
lgtm

PR #1151, opened 2023-12-12
fix(database-change): timeout or oom when upload a large sql files

### LINKED ISSUE #409: [Bug]: when the amount of data executed during database change is too large, OOM occurs

### LINKED ISSUE #886: [Bug]: Timeout in pre-check process node if the SQL files are too large

Add timeout solution for pre-check task. If the pre-check node executing for a long time, then interrupt it and continue execut ing database change task.

### Why is this needed?

If I upload a big SQL file (> 200 MB), then the pre-check node will executing more than 24 hours, and the task status is pending and without any percentage.

smallsheeeep:
#### What type of PR is this?
type-bug
module-database change

In 4.2.3, ODC will read all SQL files to a `List<String>` and then executing following operation. By the way, the executing results are all stored in a `List<SqlExecuteResult>`. So if the SQL files is large (May bigger than 256 MB), OOM may occur, and the pre-check flow node may timeout.
This PR fix it by:
- Use `SqlStatementIterator` replace `List<String>`. The `SqlStatementIterator` hold a `InputStream` and will read SQL file through stream when invoke `next()`. `SqlStatementIterator`. The `InputStream` should be close by caller.
- Write the result to the file after each SQL execution, instead of writing it all at once after all SQLs are executed. This also avoids OOM.
- Add a system config `odc.task.pre-check.max-sql-content-bytes`: maximum bytes of SQL content allowed to be checked during pre-check, exceeding this limit will terminate the pre-check and set the result to the highest risk level, default value: 5 MB.
- Database change task, Pre-check task and RollbackPlan task are adapted. The user input SQL are solved the same as before, while the SQL files are solved using `SqlStatementIterator`.

Fixes #409 
Fixes #886 

```docs

```


server/odc-core/src/main/java/com/oceanbase/odc/core/sql/split/SqlSplitter.java
```diff
@@ -833,4 +848,120 @@ private boolean isIdent(int tokenType) {
...
+                    SqlSplitter splitter = createSplitter();
+                    sqls = splitter.split(buffer.toString()).stream().map(OffsetString::getStr)
+                            .collect(Collectors.toList());
+                    while (sqls.size() > SQL_STATEMENT_BUFFER_SIZE) {
+                        String sql = sqls.remove(0);
+                        holder.addLast(sql);
+                        int index = buffer.indexOf(sql.substring(0, sql.length() - 1));
+                        buffer.delete(0, index + sql.length());
+                        clearUselessPrefix();
+                        delimiter = splitter.getDelimiter();
```
PeachThinking, line 927: Why do we need to reset the delimiter here?

server/odc-core/src/main/java/com/oceanbase/odc/core/sql/split/SqlSplitter.java
```diff
@@ -833,4 +848,120 @@ private boolean isIdent(int tokenType) {
...
+        private String delimiter;
+        private boolean firstLine = true;
+        private List<String> sqls = new ArrayList<>();
+        private long iteratedBytes = 0;
+
+        private static final int SQL_STATEMENT_BUFFER_SIZE = 5;
+        private static final Character SQL_SEPARATOR_CHAR = '/';
+        private static final Character LINE_SEPARATOR_CHAR = '\n';
+        private static final String SQL_MULTI_LINE_COMMENT_PREFIX = "/*";
+        private static final Set<Character> DELIMITER_CHARACTERS = new HashSet<>(Arrays.asList(';', '/', '$'));
```
PeachThinking, line 869: Why are only these three types of delimiters considered here?

server/odc-core/src/main/java/com/oceanbase/odc/core/sql/split/SqlSplitter.java
smallsheeeep: Because the delimiter may change during parse a long SQL content.

server/odc-core/src/main/java/com/oceanbase/odc/core/sql/split/SqlSplitter.java
smallsheeeep: Other customized delimiter is not supported

LuckyPickleZZ:
good job👍

PeachThinking:
👍👍👍

server/odc-service/src/main/java/com/oceanbase/odc/service/regulation/risklevel/RiskLevelService.java
```diff
@@ -96,6 +96,13 @@ public RiskLevel findDefaultRiskLevel() {
                         .orElseThrow(() -> new NotFoundException(ResourceType.ODC_RISK_LEVEL, "level", 0)));
     }
 
+    @SkipAuthorize("internal usage")
+    public RiskLevel findHighestRiskLevel() {
+        return entityToModel(
+                riskLevelRepository.findByOrganizationIdAndLevel(authenticationFacade.currentOrganizationId(), 3)
```
MarkPotato777, line 102: the highest level is 2

smallsheeeep: NoNoNo, it is 3. RiskLevel: default(0), low(1), medium(2), high(3).

server/odc-core/src/main/java/com/oceanbase/odc/core/sql/split/SqlCommentProcessor.java
```diff
@@ -107,13 +108,8 @@ public SqlCommentProcessor(DialectType dialectType, String delimiter) {
...
     public SqlCommentProcessor() {}
 
-    public static SqlStatementIterator iterator(InputStream in, DialectType dialectType,
-            boolean preserveFormat,
-            boolean preserveSingleComments,
-            boolean preserveMultiComments,
-            Charset charset) {
-        return new SqlStatementIterator(in, dialectType, preserveFormat, preserveSingleComments, preserveMultiComments,
-                charset);
+    public static SqlIterator iterator(InputStream in, Charset charset, SqlCommentProcessor processor) {
```
yhilmare, line 111: the calling way of this method is strange:`SqlCommentProcessor.iterator(..., SqlCommentProcessor)`

we may don't change the method signature.

server/odc-core/src/main/java/com/oceanbase/odc/core/sql/split/SqlIterator.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * This interface is used to split SQL files into SQL statements iteratively. The
+ * {@link OffsetString#getOffset()} is used to get the offset of the current SQL statement but may
+ * be not accurate.
+ * 
+ * @author gaoda.xy
+ * @date 2023/12/13 15:04
+ */
+public interface SqlIterator extends Iterator<OffsetString> {
```
yhilmare, line 29: 'SqlStatementIterator' may be better

server/odc-core/src/main/java/com/oceanbase/odc/core/sql/split/SqlSplitter.java
```diff
@@ -833,4 +847,114 @@ private boolean isIdent(int tokenType) {
...
+
+        public SqlStatementIterator(InputStream input, Charset charset, Class<? extends Lexer> lexerType,
+                String delimiter) {
+            this.reader = new BufferedReader(new InputStreamReader(input, charset));
+            this.lexerType = lexerType;
+            this.delimiter = delimiter;
+        }
+
+        @Override
+        public boolean hasNext() {
```
yhilmare, line 878: brilliant

server/odc-core/src/main/java/com/oceanbase/odc/core/sql/split/SqlSplitter.java
```diff
@@ -833,4 +847,114 @@ private boolean isIdent(int tokenType) {
...
+                    } else {
+                        buffer.append(LINE_SEPARATOR_CHAR).append(line);
+                    }
+                    // SqlSplitter is non-reentrant, so we need to create a new one for each loop
+                    SqlSplitter splitter = createSplitter();
+                    sqls = splitter.split(buffer.toString()).stream().map(OffsetString::getStr)
+                            .collect(Collectors.toList());
+                    while (sqls.size() > SQL_STATEMENT_BUFFER_SIZE) {
+                        String sql = sqls.remove(0);
+                        holder.addLast(new OffsetString(0, sql));
```
yhilmare, line 922: give a negative or null value if you don't know the offset

server/odc-core/src/main/java/com/oceanbase/odc/core/sql/split/SqlSplitter.java
```diff
@@ -833,4 +847,114 @@ private boolean isIdent(int tokenType) {
...
+                        buffer.append(line);
+                        firstLine = false;
+                    } else {
+                        buffer.append(LINE_SEPARATOR_CHAR).append(line);
+                    }
+                    // SqlSplitter is non-reentrant, so we need to create a new one for each loop
+                    SqlSplitter splitter = createSplitter();
+                    sqls = splitter.split(buffer.toString()).stream().map(OffsetString::getStr)
+                            .collect(Collectors.toList());
+                    while (sqls.size() > SQL_STATEMENT_BUFFER_SIZE) {
```
yhilmare, line 920: what is 'SQL_STATEMENT_BUFFER_SIZE' used for?

server/odc-core/src/main/java/com/oceanbase/odc/core/sql/split/SqlSplitter.java
```diff
@@ -833,4 +847,114 @@ private boolean isIdent(int tokenType) {
...
+                        String sql = sqls.remove(0);
+                        holder.addLast(new OffsetString(0, sql));
+                        int index = buffer.indexOf(sql.substring(0, sql.length() - 1));
+                        buffer.delete(0, index + sql.length());
```
yhilmare, line 924: we can not consider that the sql-spliter will not add any str to the sql

server/odc-core/src/main/java/com/oceanbase/odc/core/sql/split/SqlSplitter.java
```diff
@@ -833,4 +847,114 @@ private boolean isIdent(int tokenType) {
...
+                if (sqls.isEmpty()) {
+                    return null;
+                }
+                return new OffsetString(0, sqls.remove(0));
+            } catch (Exception e) {
+                throw new RuntimeException("Failed to parse input. reason: " + e.getMessage(), e);
+            }
+        }
+
+        private void clearUselessPrefix() {
```
yhilmare, line 942: we'd better not notice this delimiter

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/PreCheckRuntimeFlowableTask.java
```diff
@@ -236,26 +255,20 @@ private boolean isIntercepted(SqlCheckTaskResult sqlCheckResult,
...
-                new DatabasePermissionCheckResult(unauthorizedDatabaseNames);
-    }
-
-    private void doSqlCheck(TaskEntity preCheckTaskEntity, ConnectionConfig connectionConfig,
-            RiskLevelDescriber describer) {
-        if (Objects.isNull(this.databaseChangeRelatedSqls)) {
-            return;
+    private void doSqlCheckAndDatabasePermissionCheck(TaskEntity preCheckTaskEntity, RiskLevelDescriber describer) {
+        List<OffsetString> sqls = new ArrayList<>();
+        this.overLimit = getSqlContentUntilOverLimit(sqls, sqlCheckProperties.getMaxSqlContentBytes());
```
yhilmare, line 260: sql-check will not get any limit, the caller will control it

server/odc-core/src/main/java/com/oceanbase/odc/core/sql/split/SqlSplitter.java
smallsheeeep: Actually, SqlSplitter will not add or remove inner a complete SQL statement, It only remove some blank characters.

server/odc-core/src/main/java/com/oceanbase/odc/core/sql/split/SqlSplitter.java
smallsheeeep: But some delimiter has different length, SqlSplitter will all change them to `;`, we must deal with it and remove it from buffer.

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/PreCheckRuntimeFlowableTask.java
smallsheeeep: Yeah, I change the class name to `PreCheckTaskProperties` and move it to package com.oceanbase.odc.service.flow.task.model

yhilmare (approved):
lgtm

Issue #1161, opened 2023-12-13
[Bug]: sql check can not prompt user a table without table comment when there exists two table with same name


yhilmare:
### ODC version

4.2.3

### OB version

not related

### What happened?

1. type two table definitions with the same table name:
```SQL
create table asdasd(
    id varchar2(64)
);
create table asdasd(
    id varchar2(64)
);
``` 
2. click `sql-check` button:
<img width="932" alt="截屏2023-12-13 11 46 00" src="https://github.com/oceanbase/odc/assets/22992296/67aa4eb8-8200-45dd-8d50-b659f9ae1950">

3. as you can see: the second table get no violation that table without table comments.
4. the sql-check can give violation if the second table with a different name:
<img width="959" alt="截屏2023-12-13 11 47 29" src="https://github.com/oceanbase/odc/assets/22992296/276e1b0b-ca64-42fc-a28d-f608e7c68a83">

### What did you expect to happen?

the sql-check can prompt user even there exists two tables with same name

### How can we reproduce it (as minimally and precisely as possible)?

follow 'what happened'

yhilmare:
fixed:
<img width="1033" alt="截屏2023-12-13 12 22 12" src="https://github.com/oceanbase/odc/assets/22992296/9be5d2d6-384f-47ac-8685-a11905071c45">

sl01388797:
![Image](https://github.com/oceanbase/odc/assets/143492234/8310bed2-3275-40b9-bbf4-bf782039c6bc)

pass

Issue #1154, opened 2023-12-12
[Bug]: The SQL check content in the approval initiated by executing the statement is incomplete.


tonghan-gsh:
### ODC version

ODC 4.2.3

### OB version

-

### What happened?

![image](https://github.com/oceanbase/odc/assets/46598871/cbd5fbd7-8148-4a0f-aef0-37c5afded2b8)
![image](https://github.com/oceanbase/odc/assets/46598871/3a21f9fc-eb04-4804-a3dd-177384156642)
The three checks that do not require improvement for this statement are not displayed in the SQL check for the database change.

### What did you expect to happen?

The SQL check for the database change should display all the check contents.

### How can we reproduce it (as minimally and precisely as possible)?

create table test(a int,PRIMARY key(a));

yhilmare:
fixed:
<img width="893" alt="截屏2023-12-13 12 19 42" src="https://github.com/oceanbase/odc/assets/22992296/0976ba91-488e-4188-96d8-9cda5171a51d">

sl01388797:
![Image](https://github.com/oceanbase/odc/assets/143492234/565bfe6d-8ee8-47be-bc93-cfc02afff9b2)

pass

Issue #1144, opened 2023-12-12
[Bug]: Execute two identical sql, sql check content is incorrect


tonghan-gsh:
### ODC version

ODC 4.2.3

### OB version

-

### What happened?

![image](https://github.com/oceanbase/odc/assets/46598871/2e935459-e3db-4ecc-a96c-d5ba0a673170)
The second sql test result is shown in the first one.

### What did you expect to happen?

Display the correct sql check results.

### How can we reproduce it (as minimally and precisely as possible)?

Perform a sql check on two same sql.

MarkPotato777:
The root cause is the same as #1118

yhilmare:
fixed:
<img width="901" alt="截屏2023-12-13 11 31 35" src="https://github.com/oceanbase/odc/assets/22992296/ad60bd60-1ed0-498a-a6eb-0fdaec7cad7c">

Huangxiao-mas:
pass

PR #1163, opened 2023-12-13
fix(sql-check): can not give violations related comments normally when there exists same name tables

### LINKED ISSUE #1118: [Bug]: sql check results for tables with the same name are incorrect
### ODC version

ODC 4.2.3

### OB version

-

### What happened?

image
The results for test_null are shown in the results for TEST_NULL.

Display the correct sql check results.

CREATE TABLE "GSH"."TEST_NULL" ("A" NUMBER(38),"B" NUMBER(38));
insert into "GSH"."TEST_NULL"("A","B") values(DEFAULT,2),(1,DEFAULT),(3,3);
create table test_null(a int,b int);
insert into test_null values(1,null),(null,2),(3,3);

Create these tables and execute SQL check.

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

When a table with the same name exists, sql-check cannot accurately indicate that the table has no comments and the column has no comments.

Fixes #1118 #1144 #1154 #1161 

```docs

```


MarkPotato777 (approved):
LGTM

Issue #1162, opened 2023-12-13
[Bug]: The cancellation of allowing the execution of all SQL types is not taking effect


tonghan-gsh:
### ODC version

ODC 4.2.3

### OB version

-

### What happened?

Even though I have cancelled the execution of all allowed SQL types, all types are still able to execute.

### What did you expect to happen?

All sql types cannot be executed.

### How can we reproduce it (as minimally and precisely as possible)?

Deselect all executable sql types.

tonghan-gsh:
![Image](https://github.com/oceanbase/odc/assets/46598871/c1db9760-6510-4711-afc6-5f30e3dcd0f9)

pass

PR #1165, opened 2023-12-13
fix(sql-rule): several sql interception bugs

### LINKED ISSUE #1159: [Bug]: SQL statements that violate SQL window guidelines should be intercepted.
### ODC version

ODC 4.2.3

### OB version

-

### What happened?

image
The SQL statement that violates the SQL window guidelines was not intercepted, and a database change task has been initiated.

SQL statements that violate SQL window guidelines should be intercepted.

Executing SQL statements that violate SQL window guidelines.

_No response_

### Cloud

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug

This PR fixed several bugs:
- if the sql is intercepted by sql-console rules, then sql-check rules will not be checked;
- make the rule 'max-execute-sql-size' apply to the last sql;
- when the rule 'allow-execute-sql-type' value is empty, users could execute any type of sqls;
- when quering data, make sql-console rules not take effect;

Fixes #1159 #1162 #1166

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/integration/ExternalSqlInterceptor.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/integration/ExternalSqlInterceptor.java
@@ -109,7 +109,11 @@ public class ExternalSqlInterceptor extends BaseTimeConsumingInterceptor {
                         .ifPresent(rule -> {
                             Rule violationRule = new Rule();
                             RuleViolation violation = new RuleViolation();
+                            violation.setText(request.getSql());
                             violation.setLevel(2);
+                            violation.setOffset(0);
+                            violation.setStart(0);
+                            violation.setStop(request.getSql().length());
                             violation.setLocalizedMessage(SqlConsoleRules.EXTERNAL_SQL_INTERCEPTOR
                                     .getLocalizedMessage(new Object[] {rule.getProperties()
                                             .get(rule.getMetadata().getPropertyMetadatas().get(0).getName())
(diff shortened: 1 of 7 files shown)


yhilmare:
we may merge this pr #1163 first, It looks like having some conflicts~

server/odc-service/src/main/java/com/oceanbase/odc/service/integration/ExternalSqlInterceptor.java
```diff
@@ -109,7 +109,11 @@ public boolean doPreHandle(@NonNull SqlAsyncExecuteReq request, @NonNull SqlAsyn
                         .ifPresent(rule -> {
                             Rule violationRule = new Rule();
                             RuleViolation violation = new RuleViolation();
+                            violation.setText(request.getSql());
```
MarkPotato777, line 112: if violates the external SQL interception rule, then put all SQLs into one line, and the highlight is all the SQLs.

server/odc-service/src/main/java/com/oceanbase/odc/service/regulation/ruleset/SqlConsoleRuleService.java
```diff
@@ -146,7 +146,7 @@ public <T> Optional<List<T>> getListProperties(@NonNull Long rulesetId, @NonNull
                 return result;
             }).collect(Collectors.toList());
             if (CollectionUtils.isEmpty(properties) || properties.get(0).isEmpty()) {
-                return Optional.empty();
+                return Optional.of(Collections.emptyList());
```
MarkPotato777, line 149: fix #1162

yhilmare (approved):
lgtm

Issue #237, opened 2023-09-07
[Bug]: The table name in the content of the sql downloaded data file does not contain double quotation marks


runzi389205:
### ODC version

4.2.1

### OB version

5.7.25-OceanBase-v4.1.0.1

### What happened?

Download data in sql format for the table named ".",
view downfile,table name without double quotes.
![image](https://github.com/oceanbase/odc/assets/141202865/c6219205-3ba2-4d3e-8dc8-13ead9bccf2b)
![image](https://github.com/oceanbase/odc/assets/141202865/527cc3ff-b0ff-41c9-847f-9fac5fd6c936)
another one：
![image](https://github.com/oceanbase/odc/assets/141202865/841e4e74-7a9e-45b0-ab87-b7073f036f13)

### What did you expect to happen?

view downfile,table name with double quotes.

### How can we reproduce it (as minimally and precisely as possible)?

Enter table data -> execute download data (sql format) -> download file to local

Issue #1158, opened 2023-12-13
[Bug]: Failed to check global snapshot in oracle mode


Huangxiao-mas:
### ODC version

ODC423

### OB version

Oceanbase2277

### What happened?

Failed to check global snapshot in oracle mode
![image](https://github.com/oceanbase/odc/assets/59468957/6cd75a26-63db-4a3e-a627-5b75241ea200)

### What did you expect to happen?

export normally

### How can we reproduce it (as minimally and precisely as possible)?

chat privately

sl01388797:
![Image](https://github.com/oceanbase/odc/assets/143492234/df4a275c-ab2f-4608-b8c9-eb6d71be7e5b)

pass

PR #1172, opened 2023-12-13
fix(data-transfer): only inject sys tenant config when it's configured in datasource

### LINKED ISSUE #1030: [Bug]: Ordinary accounts do not have the permission to modify the data source to initiate an export task. When the sys account clicks the test connection, a permission error is reported.

LuckyPickleZZ:
#### What type of PR is this?
type-bug

There are some issues,  such as #1030 and #1158 , are caused by the sys tenant configs. 
On the one hand, due to the fact that after OceanBase4.x, ob-loader-dump no longer relies on the account and password of sys tenants. On the other hand, this configuration has security issues. 
Therefore, we prohibit users from filling in the sys tenant account password when initiating transfer tasks.

Fixes #1030 and #1158

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/datatransfer/DataTransferService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/datatransfer/DataTransferService.java
@@ -361,22 +361,21 @@ public class DataTransferService {
     }
 
     private void injectSysConfig(ConnectionConfig connectionConfig, DataTransferConfig transferConfig) {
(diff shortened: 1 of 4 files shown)


LuckyPickleZZ:
also fix #237

yhilmare (approved):
lgtm

PR #1173, opened 2023-12-13
feat(taskframework):  add real time log query in task executor

krihy:
#### What type of PR is this?

type-feature
module-taskframework
1.`EmbedServer` bind port for http request.
2.`EmbedServer` add uri mapping for log query,  log uri path is  `/([0-9]+)/tasks/log`
3. All task log file path like  $LOG_INSTALL_PATH/taskId/task.all and  warn log file 
$LOG_INSTALL_PATH/taskId/task.warn

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/ExecutorBizImpl.java
```diff
@@ -0,0 +1,42 @@
...
+ * @date 2023-12-13
+ * @since 4.2.4
+ */
+@Slf4j
+public class ExecutorBizImpl implements ExecutorBiz {
+
+    private static final String LOG_PATH_PATTERN = "%s/%s/task.%s";
+
+    @Override
+    public SuccessResponse<String> log(Long id, String logType) {
```
smallsheeeep, line 36: Why use String rather than enum for `logType`?  Please verify the `logType` before concat `filePath`.

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/logger/LogUtils.java
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * @author yaobin
+ * @date 2023-12-13
+ * @since 4.2.4
+ */
+@Slf4j
+public class LogUtils {
+
+    private static final long MAX_LOG_LINE_COUNT = 10000;
+    private static final long MAX_LOG_BYTE_COUNT = 1024 * 1024;
```
smallsheeeep, line 38: Why only 1 MB? It is too short for task log such as database change

krihy: we just read head line in log, use MAX_LOG_LINE_COUNT or MAX_LOG_BYTE_COUNT limit line size

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/ExecutorBizImpl.java
krihy: done , add OdcTaskLogLevel valid type

PR #1186, opened 2023-12-14
enhancement: use git-lfs to storage datax.zip

LuckyPickleZZ:
#### What type of PR is this?
type-enhancement

Due to the large volume of datax.zip, we use git lfs to store it. We could track the lfs files in `.gitattributes`.

The changes in build_*.yaml is because: Github does not replace the pointer file of LFS with the actual file when executing actions. So we need to manually execute the `git lfs checkout` to obtain it.

I have tested building artifacts, and this passed them.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+datax.zip filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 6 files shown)


yhilmare (approved):
lgtm

PR #1188, opened 2023-12-14
fix(concurrent): remove servlet configuration

LuckyPickleZZ:
remove ServletConfiguration because it may cause concurrent problems


yhilmare:
so，how to solve the locale problem?

yhilmare (approved):
lgtm

PR #1194, opened 2023-12-15
fix(sql-rule): disabling the rule 'allow-execute-sql-types'  does not work

### LINKED ISSUE #1190: [Bug]: After the sql window check specification is closed, the sql check specification is still triggered.
### ODC version

ODC4.2.3

### OB version

independent

### What happened?

After the sql window check specification is closed, the sql check specification is still triggered.
image

After the SQL window inspection specification is closed, the execution results are directly displayed.

chat privately

_No response_

### Cloud

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug

Disabling the rule 'allow-execute-sql-types'  does not work. The root cause is that it doesn't distinguish from empty properties and disabled properties.

Fixes #1190

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/regulation/ruleset/SqlConsoleRuleService.java
```diff
@@ -128,7 +128,7 @@ public <T> Optional<List<T>> getListProperties(@NonNull Long rulesetId, @NonNull
                 return true;
             }).map(rule -> {
                 if (Objects.isNull(rule.getProperties())) {
-                    return Collections.EMPTY_LIST;
+                    return null;
```
yhilmare, line 131: why null, does empty list not work?

MarkPotato777: we need to distinguish 'null' and 'empty list'. 'Null' means that this rule is disabled while 'empty list' means that this rule is enabled but the value is empty.

yhilmare: okay

yhilmare (approved):
lgtm

PR #1198, opened 2023-12-15
feat(datatransfer): support masking data for mysql datatransfer

LuckyPickleZZ:
#### What type of PR is this?
type-feature

In this PR, we support masking data during transfer mysql data. 

DataX allows us to use groovy script to transform data. Therefore, our data mask is also based on this function.
The current temporary solution is: when a column needs to be desensitized, regardless of the mask algorithm configured by the user, we will use the default masking algorithm for desensitization. That is to replace the data of sensitive columns with '******'.

The groovy script templates are configured in `resources/mask_rule_template.yaml`. If we need to increase support for mask algorithms in the future, we can add templates in this file.


smallsheeeep (approved):
LGTM

yhilmare (approved):
lgtm

PR #1202, opened 2023-12-15
fix(database): optimize error message of synchronizing databases failure

### LINKED ISSUE #773: [Bug]: The error message that fails to obtain the lock causes the failure of the database synchronization is inaccurate
### ODC version

4.2.1

### OB version

null

### What happened?

After clicking the synchronous database, error, and the error message causes customers to misunderstand:
image

Optimize performance, optimize error messages

Synchronous database

_No response_

### Cloud

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug

This PR optimizes error message of synchronizing databases failure.

Fixes #773 

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
@@ -354,7 +354,8 @@ public class DatabaseService {
     public Boolean internalSyncDataSourceSchemas(@NonNull Long dataSourceId) throws InterruptedException {
         Lock lock = jdbcLockRegistry.obtain(connectionService.getUpdateDsSchemaLockKey(dataSourceId));
         if (!lock.tryLock(3, TimeUnit.SECONDS)) {
-            throw new ConflictException(ErrorCodes.ResourceModifying, "Can not acquire jdbc lock");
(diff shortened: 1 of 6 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
```diff
@@ -354,7 +354,8 @@ public Boolean syncDataSourceSchemas(@NonNull Long dataSourceId) throws Interrup
     public Boolean internalSyncDataSourceSchemas(@NonNull Long dataSourceId) throws InterruptedException {
         Lock lock = jdbcLockRegistry.obtain(connectionService.getUpdateDsSchemaLockKey(dataSourceId));
         if (!lock.tryLock(3, TimeUnit.SECONDS)) {
-            throw new ConflictException(ErrorCodes.ResourceModifying, "Can not acquire jdbc lock");
+            throw new ConflictException(ErrorCodes.ResourceSynchronizing,
+                    new Object[] {ResourceType.ODC_DATABASE.getLocalizedMessage()}, "Can not acquire jdbc lock");
```
MarkPotato777, line 358: image

yhilmare (approved):
lgtm

PR #1208, opened 2023-12-15
fix(monitor): druid stats use mysql parser

ungreat:
#### What type of PR is this?
fix druid stats may merge sql error by use oceanbase dbType, use mysql parser instead.

Fixes #

```docs

```


yhilmare (approved):
lgtm

Issue #1219, opened 2023-12-18
[Bug]: opening a sql console is not recorded in operating records


MarkPotato777:
### ODC version

4.2.0

### OB version

not related

### What happened?

opening a SQL console is not recorded in operating records.

### What did you expect to happen?

opening a SQL console shoud be recorded in audit events.

### How can we reproduce it (as minimally and precisely as possible)?

1. Opening a SQL console
2. see operating records

Issue #1214, opened 2023-12-18
[Bug]: executing sql is not recorded in audit event


MarkPotato777:
### ODC version

4.2.3

### OB version

not related

### What happened?

I executed SQL in the SQL console, but I cannot see records in audit events.

### What did you expect to happen?

I can see records of executing SQLs in audit events.

### How can we reproduce it (as minimally and precisely as possible)?

As I said in 'When happened'.

sl01388797:
![Image](https://github.com/oceanbase/odc/assets/143492234/45f949ad-6b44-46ba-a2bf-539ed15578ef)

pass

PR #1222, opened 2023-12-18
fix(audit): several operating records issues after ODC V4.2.0

### LINKED ISSUE #1041: [Bug]: No data source displayed in the operation record
### ODC version

odc423

### OB version

*

### What happened?

image

Existing data source

1. Performing operations within the data source
2. View operation records

_No response_

### Cloud

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug

This PR fixed several issues:
1. it shows empty data source information in operating records when creating a task;
2. executing SQLs is not recorded in operating records;
3. opening SQL console is not recorded in operating records;

The reasons for the above issues are the same: the audit module doesn't adapt to ODC V4.2.x.
I added a column `database_id_extract_expression` on `audit_event_meta`, and added `database_id` and `database_name` on `audit_event` because after ODC V4.2.x, a part of operations doesn't relate directly to `DataSource(Connection)`, but relates to `Database`.

Fixes #1041  #1214   #1219 

```docs

```


server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/R4239AuditEventMetaMigrate.java
```diff
@@ -39,9 +39,9 @@
  * @Description: []
  */
 @Slf4j
-@Migratable(version = "4.2.0.1", description = "Built-in audit event meta migrate",
+@Migratable(version = "4.2.3.9", description = "Built-in audit event meta migrate",
```
MarkPotato777, line 42: I changed the version because I need to execute this migration after `V_4_2_3_8__alter_audit_event.sql`

server/odc-service/src/main/java/com/oceanbase/odc/service/audit/AuditEventAspect.java
```diff
@@ -277,29 +283,33 @@ private AuditEvent createAuditEvent(Method method, Object[] args) {
                 .username(authenticationFacade.currentUsername())
                 .detail(parseDetailFromApiParams(apiParams))
                 .build();
-        // 创建连接时，请求中还没有 id，需要通过 VisibleScope 来判断是否为公共连接
```
MarkPotato777, line 280: `VisibileScope` is deprecated after ODC V4.2.0 so I deleted this

server/odc-service/src/main/java/com/oceanbase/odc/service/audit/AuditEventAspect.java
```diff
@@ -438,7 +457,7 @@ private AuditEvent setConnectionRelatedProperties(AuditEvent auditEvent, String
         /**
          * If sid represents static connectionId
          */
-        if (!StringUtils.contains(sid, "-")) {
+        if (!StringUtils.contains(sid, "sid:")) {
```
MarkPotato777, line 460: I changed the '-' to 'sid:' because before V4.2.3, sid's format is like 'sid:1-1' while after V4.2.3 sid is encrypted like 'sid:eyJkYklkIjoxOSwiZHNJZCI6MSwiZnJvbSI6IkItUTk2RE1ENlItMDMzMy5sb2NhbCIsInJlYWxJZCI6IjAxODNhNGYxOGYifQ==:d:testenv', so '-' could not be a flag to distinguish connection id and dynamic sid any more.

server/odc-service/src/main/java/com/oceanbase/odc/service/audit/AuditEventAspect.java
```diff
@@ -450,13 +469,7 @@ private AuditEvent setConnectionRelatedProperties(AuditEvent auditEvent, String
             ConnectionSession session = sessionService.nullSafeGet(SidUtils.getSessionId(sid));
             Object value = ConnectionSessionUtil.getConnectionConfig(session);
             if (value instanceof ConnectionConfig) {
-                /**
-                 * Do NOT record event happened in private connection
-                 */
                 ConnectionConfig config = (ConnectionConfig) value;
-                if (ConnectionVisibleScope.PRIVATE == config.getVisibleScope()) {
```
MarkPotato777, line 457: `VisibleScope` is deprecated so I deleted this.

server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/R4239AuditEventMetaMigrate.java
yizhouxw: the previous migrate may has been executed, and migration history recorded in metadb migration_history.
we would better keep the previous Migrate class, or change to use a 'repeatable Migrate' instead.

MarkPotato777: Actually, this is a 'repeatable Migrate'. I just want to change its execution order to after `V_4_2_3_8__alter_audit_event.sql` otherwise it will execute fail because `V_4_2_3_8__alter_audit_event` alters the table. Even if the previous migration has been executed, I think it will not bring any bad influence. What's your consideration here for keeping the previous Migrate class? For tracing the migration history?

server/odc-service/src/main/java/com/oceanbase/odc/service/audit/AuditEventAspect.java
yhilmare: `DefaultConnectSessionIdGenerator#getKeyFromId`

this method may help you get connection id, datasource id and schema name

MarkPotato777: > `DefaultConnectSessionIdGenerator#getKeyFromId`
> 
> this method may help you get connection id, datasource id and schema name

I do not need to get the connection id and schema name here. I just need to know if this is a dynamic sid or a static connection id and that's enough.

yhilmare (approved):
lgtm

PR #1224, opened 2023-12-18
feat(datasource): show datasource's connect status in team space's SQL console

### LINKED ISSUE #1223: [Feature]: show datasource status in team space's SQL console

In the team space's SQL console, I could double-click the data source which is already invalid. And it will promote an error after I click.

I want to know the data source status in the team space's SQL console just as in the individual space's.

### Additional context

_No response_

MarkPotato777:
#### What type of PR is this?

type-feature

It doesn't show data source status in team space's SQL console

This PR makes the status API adapt to the team space: if the current user has the specific datasource read permission or it joined the project which contains the specific datasource's databases, then I will return the status of this specific datasource.

Fixes #1223 

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
```diff
@@ -432,34 +454,29 @@ public Map<Long, CheckState> getStatus(@NonNull Set<Long> ids) {
         List<Permission> granted = res2Actions.entrySet().stream()
                 .map(e -> securityManager.getPermissionByActions(e.getKey(), e.getValue()))
                 .collect(Collectors.toList());
+        Set<Long> projectRelatedConnectionIds = databaseService.statsConnectionConfig().stream()
```
smallsheeeep, line 457: The `databaseService.statsConnectionConfig()` will return all project related connection ids? This method name and parameters are hard to understand without any relevant information.

MarkPotato777: Yes. Any advice for the method name?

MarkPotato777: btw, `databaseService.statsConnectionConfig()` returns the ConnectionConfigs rather than connectionIds. And I believe `stats` is the exact meaning, the thing that may confuse you is that `statsConnectionConfig` is better placed in `ProjectService`. Will it be better if `statsConnectionConfig` is placed in `ProjectService`?

yhilmare (commented):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
smallsheeeep: No, users can also gain access to datasources without going through the project.

MarkPotato777: So this is why it should be placed in `ProjectService` rather than like `ConnectionService` or other classes.

yhilmare (approved):
lgtm

PR #1225, opened 2023-12-18
feat(taskframework): retry when failed to report task results

smallsheeeep:
#### What type of PR is this?
type-feature
module-task framework

When a task is finished, we must sure that all task related info has been reported to ODC server successfully. So, in this PR, I add retry for report task results.
By the way, I add a field `finished` in `TaskResult` for ODC server to judge whether the Pod could be expired.

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/BaseTask.java
```diff
@@ -127,35 +126,81 @@ public FlowTaskResult result() {
...
-        this.status = status;
+    private void doFinal() {
+        onUpdate();
+        if (this.status == TaskStatus.DONE) {
+            this.progress = 1.0;
+        }
+        log.info("Task finished with status: {}, id: {}, start to report final result",
+                context.getJobIdentity().getId(), status);
+        DefaultTaskResult finalResult = buildCurrentResult();
+        finalResult.setFinished(true);
```
krihy, line 167: finalResult finished is true , but log is not uploaded at this moment, 
we should split two step, first is upload terminal status like done/failed/canceled,  second upload finished  until log upload completed

smallsheeeep: Okay, the `doFinal()` change to 3 steps:
1. report final result (progress, status, resultJson)
2. upload log file to OSS (TODO)
3. report `finished` signal to ODC server

krihy (approved):
lgtm

PR #1228, opened 2023-12-19
fix(database-object):modify the prompt that prompts users about the risk of index changes

### LINKED ISSUE #1142: [Bug]: Provide relevant prompts to users for high-risk operations such as index changes when generate update table ddl
### ODC version

4.2.3

### OB version

all

### What happened?

Provide relevant prompts to users for high-risk operations such as index changes when generate update table ddl

Provide relevant prompts to users for high-risk operations such as index changes when generate update table ddl

White screen creates or deletes an index on a table

_No response_

### Cloud

_No response_

PeachThinking:
#### What type of PR is this?
type-bug
module-database object

modify the prompt that prompts users about the risk of index changes

Fixes #1142

```docs

```


yizhouxw (approved):
lgtm

PR #1243, opened 2023-12-20
fix(datasource): it occurs 'duplicate data source name' error when creating a data source in team space

### LINKED ISSUE #1241: [Bug]: it occurs 'duplicate datasource name' while creating a team space's datasource after upgrading from ODC 4.1.x to 4.2.x
### ODC version

4.2.x

### OB version

not related

### What happened?

After I upgraded ODC from V4.1.2 to V4.2.2, I encountered an error message that said 'duplicate data source name' when was creating a team space's data source however I did not create any team space's data source before.

I could create team space's data sources successfully when I have no duplicated data sources in team space.

1. start up an ODC V4.1.2;
2. user_a creates a private connection named 'ds_a';
3. upgrade ODC to V4.2.2;
4. log in ODC use admin account;
5. create a data source named 'ds_a' in team space;using

The root cause is that private connection migration happens when each user logs in. When we create a data source, ODC will check if the data source name is occupied in the current organization, and it is possible that the user acutally has a private connection with the same name and this user has not logged in yet. Then the name occupation check will fail.

### Cloud

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug

After I upgraded ODC from V4.1.2 to V4.2.2, I encountered an error message that said 'duplicate data source name' when was creating a team space's data source however I did not create any team space's data source before.

The root cause is that private connection migration happens when each user logs in. When we create a data source, ODC will check if the data source name is occupied in the current organization, and it is possible that the user actually has a private connection with the same name and this user has not logged in yet. Then the name occupation check will fail.

In this PR, I added the VisibleScope.ORGANIZATION into condition when checking data source name duplication. In this way, even if there exists a private connection that has not been migrated yet, it will not affect to creation of the same name data source in the team space.

Fixes #1241 

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
@@ -801,6 +801,7 @@ public class ConnectionService {
         ConnectionEntity entity = new ConnectionEntity();
         entity.setOrganizationId(organizationId);
         entity.setName(name);
+        entity.setVisibleScope(ConnectionVisibleScope.ORGANIZATION);
         return repository.exists(Example.of(entity));
     }
 


yhilmare (approved):
lgtm

PR #1247, opened 2023-12-20
fix(partition plan):failed to disable table partition plan

### LINKED ISSUE #1234: [Bug]: Tables that do not set a partition strategy also perform partition planning tasks
### ODC version

ODC 4.2.3

### OB version

-

### What happened?

image
image
Tables that do not set a partition strategy also perform partition planning tasks.

Tables that do not set a partition strategy do not perform partition planning tasks.

Follow what happened.

_No response_

### Cloud

_No response_

guowl3:
#### What type of PR is this?
type-bug

When the partition plan task is executed, disabled partition plans are not filtered out, which will cause the paused partition plan to become invalid.

Fixes #1234

```docs

```


krihy (approved):
lgtm

PR #1249, opened 2023-12-20
fix(schema-plugin): show partition name with identifiers

### LINKED ISSUE #1244: [Bug]: show table partition name with "`"
### ODC version

4.2.2、4.2.3

### OB version

ob 4.2.1

### What happened?

If the partition name in the table's ddl contains an identifier, the displayed partition name also contains the identifier.
image

partition name do not contains the identifier.

in ob mysql 4.2.1 create a parition table and the partition name in the table's ddl contains an identifier

_No response_

### Cloud

_No response_

PeachThinking:
#### What type of PR is this?

type-bug
module-database object

remove identifiers before set partition name 

Fixes #1244

```docs

```


yhilmare (approved):
lgtm

PR #1253, opened 2023-12-20
fix(datasource):  convert the type of ob-mysql-sharding data source to ob-mysql

### LINKED ISSUE #1252: [Bug]: The back-end DialectType has added ODP sharding, which may cause the front-end page to crash
### ODC version

4.2.3

### OB version

odp sharding mysql

### What happened?

The back-end `DialectType` has added ODP sharding MySQL, which may cause the front-end page to crash.

There is no ODP sharding mysql in the DialectType returned by the backend to the frontend.

This hidden bug has not been triggered so far

_No response_

### Cloud

_No response_

PeachThinking:
#### What type of PR is this?
type-bug
module-datasource

The back-end `DialectType` has added `ODP_SHARDING_OB_MYSQL`, which may cause the front-end page to crash. This hidden bug has not been triggered so far.
In order to avoid this bug, `ODP_SHARDING_OB_MYSQL` is converted to `OB_MYSQL` during serialization.

Fixes #1252

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/config/BeanConfiguration.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/config/BeanConfiguration.java
@@ -74,9 +74,8 @@ public class BeanConfiguration {
     public ObjectMapper objectMapper(SensitivePropertyHandler sensitivePropertyHandler) {
         CustomOutputSerializer customOutputSerializer = new CustomOutputSerializer()
                 .addSerializer(Internationalizable.class, new I18nOutputSerializer());
-        DialectTypeOutputSerializer dialectTypeOutputSerializer = new DialectTypeOutputSerializer();
         SimpleModule dialectTypeModule =
-                new SimpleModule().addSerializer(DialectType.class, dialectTypeOutputSerializer);
+                new SimpleModule().addSerializer(DialectType.class, new DialectTypeOutputSerializer());
         return JacksonFactory.unsafeJsonMapper()
                 .registerModule(JacksonModules.sensitiveInputHandling(sensitivePropertyHandler::decrypt))
                 .registerModule(JacksonModules.customOutputHandling(customOutputSerializer))
@@ -184,8 +183,10 @@ public class BeanConfiguration {
 
     /**
      * In order to adapt to the fact that there is no ODP_SHARDING_OB_MYSQL in DialectType of the
-     * front-end, ODP_SHARDING_OB_MYSQL is converted to OB_MYSQL during serialization.
+     * front-end, ODP_SHARDING_OB_MYSQL is converted to OB_MYSQL during serialization. Will be removed
+     * in the future
      */
+    @Deprecated
     private static class DialectTypeOutputSerializer extends JsonSerializer<DialectType> {
 
         @Override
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/config/BeanConfiguration.java
```diff
@@ -171,4 +182,23 @@ public OBDatabaseUser getSysTenantUser(String instanceId) {
...
+    /**
+     * In order to adapt to the fact that there is no ODP_SHARDING_OB_MYSQL in DialectType of the
+     * front-end, ODP_SHARDING_OB_MYSQL is converted to OB_MYSQL during serialization.
+     */
+    private static class DialectTypeOutputSerializer extends JsonSerializer<DialectType> {
+
+        @Override
+        public void serialize(DialectType dialectType, JsonGenerator jsonGenerator,
+                SerializerProvider serializerProvider) throws IOException {
+            if (Objects.nonNull(dialectType)) {
```
yhilmare, line 194: what if null?

server/odc-service/src/main/java/com/oceanbase/odc/config/BeanConfiguration.java
```diff
@@ -67,9 +74,13 @@ public PasswordEncoder passwordEncoder() {
     public ObjectMapper objectMapper(SensitivePropertyHandler sensitivePropertyHandler) {
         CustomOutputSerializer customOutputSerializer = new CustomOutputSerializer()
                 .addSerializer(Internationalizable.class, new I18nOutputSerializer());
+        DialectTypeOutputSerializer dialectTypeOutputSerializer = new DialectTypeOutputSerializer();
+        SimpleModule dialectTypeModule =
```
yhilmare, line 78: `SimpleModule dialectTypeModule = new SimpleModule().addSerializer(DialectType.class, new DialectTypeOutputSerializer());` may be better

server/odc-service/src/main/java/com/oceanbase/odc/config/BeanConfiguration.java
```diff
@@ -171,4 +182,23 @@ public OBDatabaseUser getSysTenantUser(String instanceId) {
         }
     }
 
+    /**
+     * In order to adapt to the fact that there is no ODP_SHARDING_OB_MYSQL in DialectType of the
+     * front-end, ODP_SHARDING_OB_MYSQL is converted to OB_MYSQL during serialization.
+     */
+    private static class DialectTypeOutputSerializer extends JsonSerializer<DialectType> {
```
yhilmare, line 189: add `@Deprecated` annotation to notice others that this is a temporary solution

server/odc-service/src/main/java/com/oceanbase/odc/config/BeanConfiguration.java
PeachThinking: done

server/odc-service/src/main/java/com/oceanbase/odc/config/BeanConfiguration.java
PeachThinking: done

server/odc-service/src/main/java/com/oceanbase/odc/config/BeanConfiguration.java
PeachThinking: Added logic for handling null ：`jsonGenerator.writeNull();`

yhilmare (approved):
lgtm

Issue #1268, opened 2023-12-21
[Bug]: Native MySQL export result set failed


sl01388797:
### ODC version

odc423

### OB version

*

### What happened?

![image](https://github.com/oceanbase/odc/assets/143492234/57381d40-c1e1-4118-a4ba-4025c6c2015b)

### What did you expect to happen?

Export result set successful

### How can we reproduce it (as minimally and precisely as possible)?

Native MySQL export result set

sl01388797:
![Image](https://github.com/oceanbase/odc/assets/143492234/b92a62d3-13ed-4fbc-9cc2-802d9fb70e3b)

pass

PR #1275, opened 2023-12-21
fix(result-export): failed to export mysql data

### LINKED ISSUE #1271: [Bug]: Native MySQL import CSV file error
### ODC version

odc423

### OB version

*

### What happened?

image
log
[text (17).log](https://github.com/oceanbase/odc/files/13739589/text.17.log)

Import was successful

Import CSV files into native MySQL

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

1. when import csv file with blank spaces in filename like 'da ta.csv', thie URL#getFile will get 'da%20da.csv', causing FileNotFoundException. So I use URL#toUri instead. 
2. there are some bugs in result-set-export for mysql datasource , I fixed them...

Fixes #1271 #1268

### DIFF
--- a/server/plugins/task-plugin-mysql/src/main/java/com/oceanbase/odc/plugin/task/mysql/datatransfer/job/datax/model/parameter/MySQLReaderPluginParameter.java
+++ b/server/plugins/task-plugin-mysql/src/main/java/com/oceanbase/odc/plugin/task/mysql/datatransfer/job/datax/model/parameter/MySQLReaderPluginParameter.java
@@ -31,12 +31,12 @@ public class MySQLReaderPluginParameter implements PluginParameter {
      * export all columns
      */
     private List<String> column = Collections.singletonList("*");
-    private List<Connection> connection;
+    private List<DataXConnection> connection;
     // TODO private String splitPk;
 
     @Data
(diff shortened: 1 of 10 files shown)


server/plugins/task-plugin-mysql/src/main/java/com/oceanbase/odc/plugin/task/mysql/datatransfer/job/datax/model/parameter/MySQLReaderPluginParameter.java
```diff
@@ -33,17 +33,21 @@ public class MySQLReaderPluginParameter implements PluginParameter {
...
     // TODO private String splitPk;
-    /**
-     * must be singleton. If this field is specified, the DataX will ignore the other configurations
-     * such as table and column, and directly use the content of this configuration to filter the data.
-     */
-    private List<String> querySql;
 
     @Data
     @AllArgsConstructor
     public static class Connection {
```
LuckyPickleZZ, line 39: rename

yhilmare (approved):
lgtm

yhilmare (approved):
lgtm

PR #1280, opened 2023-12-22
fix(datasource): make ODP_SHARDING_OB_MYSQL not be converted to OB_MYSQL in some special cases

### LINKED ISSUE #1279: [Bug]: The optionally supported data source types in the security specification are incorrect
### ODC version

4.2.3

### OB version

all

### What happened?

The optionally supported data source types in the security specification are incorrect：
image

show ODP_sharding_mysql type

as above

_No response_

### Cloud

_No response_

PeachThinking:
#### What type of PR is this?
type-bug

Fix the bug introduced by this PR：https://github.com/oceanbase/odc/pull/1253
There are some special scenarios, such as the optional DialectType in security rules and the DialectType displayed in operation records. In these cases, ODP_SHARDING_OB_MYSQL should not be converted to OB_MYSQL.

Fixes #1279

```docs

```

### DIFF
--- a/server/odc-common/src/main/java/com/oceanbase/odc/common/json/NormalDialectTypeOutput.java
+++ b/server/odc-common/src/main/java/com/oceanbase/odc/common/json/NormalDialectTypeOutput.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright (c) 2023 OceanBase.
+ *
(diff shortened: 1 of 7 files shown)


yhilmare (approved):
lgtm

PR #1281, opened 2023-12-22
fix(sql-rule): adjust several sql-console rules' default values

MarkPotato777:
please hold~ I am testing...

#### What type of PR is this?

type-bug

This PR adjusts some sql-console rules' default values if users have not updated the value.

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #1288, opened 2023-12-22
fix(result-set-export): failed to rewrite sql for mysql

### LINKED ISSUE #1282: [Bug]: Native MySQL export result set selection CSV and Excel format selection without column header cannot take effect
### ODC version

odc423

### OB version

*

### What happened?

9a2e91cc9f75a69da0a8a401d48d22fe
8e1632b60e90b3154aab00c77dee1928

The exported result set has no column header

Native MySQL export result set settings do not include column headers

_No response_

### Cloud

_No response_

### LINKED ISSUE #1283: [Bug]: The native MySQL export result set query result limit cannot take effect
### ODC version

odc423

### OB version

*

### What happened?

6d17d4542831750944f77fe692f63648
50aa8ffcec084f7821fe3c27cf91e797

The limit on the number of query results is in effect

The limit on the number of query results is set to be smaller than the table data

_No response_

### Cloud

_No response_

LuckyPickleZZ:
Fix #1283 . When exporting result set, we would rewrite query sql into like `select * from (<original sql>) limit xxx` . 
In ob-mysql , this works normally. But in native mysql, this rewrite would get a syntax error 'Every derived table must have its own alias'. 
Therefore, we will add a `as  __ODC_INTERNAL_RESULT_SET__` into query to fix this.

Fix #1282 . When user set skip header, the `TxtWriterPluginParameter#header` field should not be set.

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/flowtask/ResultSetExportPreprocessor.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/flowtask/ResultSetExportPreprocessor.java
@@ -77,7 +77,7 @@ public class ResultSetExportPreprocessor implements Preprocessor {
         BasicResult r = factory.buildAst(sql).getParseResult();
         Verify.verify(ParserUtil.getGeneralSqlType(r) == GeneralSqlType.DQL, "Invalid sql type, query must be DQL!");
 
-        parameter.setSql(sql);
+        parameter.setSql(sqls.get(0));
     }
 
 }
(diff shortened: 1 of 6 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/flowtask/ResultSetExportPreprocessor.java
```diff
@@ -53,15 +53,15 @@ public void process(CreateFlowInstanceReq req) {
...
+        preCheckSql(parameter, connectionConfig.getDialectType());
 
         PreConditions.validArgumentState(FILE_NAME_PATTERN.matcher(parameter.getFileName()).matches(),
                 ErrorCodes.BadArgument, null,
                 "File name must contain only letters, numbers, Chinese characters and \"._-\" ");
     }
 
-    private void preCheckSql(String sql, DialectType dialectType) {
+    private void preCheckSql(ResultSetExportTaskParameter parameter, DialectType dialectType) {
+        String sql = parameter.getSql();
```
yhilmare, line 64: is the new code different from old version?

LuckyPickleZZ: I added `parameter.setSql(sqls.get(0));` at line 80.

LuckyPickleZZ:
also fix #1292 , when table name is in Chinese, there are no content in exported file.

yhilmare (approved):
lgtm

PR #1298, opened 2023-12-25
feat(sql-execute): add size limit for sql console query

LuckyPickleZZ:
#### What type of PR is this?
type-feature

If the size of query result se is very large, there may be some unexpected exceptions.

So I added a limit for query result set, the default value is -1, indicating no restriction.
This param could be configured in metadb, and it's value must be a positive number.


yizhouxw (approved):
lgtm

PR #1299, opened 2023-12-25
feat(dlm): upgrade dlm's version to 1.0.8

### LINKED ISSUE #1295: [Bug]: there is illegal log printing for dlm task
### ODC version

4.2.3

### OB version

not related

### What happened?

1. start odc app
2. view log dir: `log/scheduleTask`
截屏2023-12-25 10 01 11

1. there is illegal log file and dir: `${ctx:jobName}-${ctx:jobGroup}`
2. the log dir name should be lower case, 'scheduleTask' is not so good, 'scheduletask' or 'schedule-task' will be better

follow 'what happened'

_No response_

### Cloud

_No response_

guowl3:
#### What type of PR is this?
type-bug

Fixed an issue where monitoring logs were routed to non-specified files.

Fixes #1295

```docs

```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -119,7 +119,7 @@
         <mina.version>2.1.6</mina.version>
 
         <!-- data-lifecycle-manager version -->
-        <data-lifecycle-manager.version>1.0.7</data-lifecycle-manager.version>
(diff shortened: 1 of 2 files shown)


yhilmare (approved):
lgtm

PR #1302, opened 2023-12-25
fix(sql-rule): the rule 'allow-sql-types' of dev environment is disabled by default

### LINKED ISSUE #1301: [Bug]: When upgrading 422 to 423, the enabled value is not changed to the default value.
### ODC version

ODC 4.2.3

### OB version

-

### What happened?

Version 422 does not change the default value of the sql check specification. When upgrading to 423, the "SQL types allowed to be executed" is not enabled by default.

"Allow execution sql types" is enabled by default when upgrading to 423

Do not modify the default value of SQL specification in version 422, upgrade to version 423

_No response_

### Cloud

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug

The rule 'allow-sql-types' of the dev environment is disabled by default. The root cause is that the default value was wrongly set to disabled in ODC V4.2.0.
This PR corrects it to 'enabled' if this rule is not edited by users.

Fixes #1301 

```docs

```

### DIFF
--- a/server/odc-migrate/src/main/resources/migrate/common/V_4_2_3_10__adjust_sql_console_rules.sql
+++ b/server/odc-migrate/src/main/resources/migrate/common/V_4_2_3_10__adjust_sql_console_rules.sql
@@ -3,6 +3,7 @@
 update
   regulation_rule_applying rra
 set
+  rra.is_enabled = 1,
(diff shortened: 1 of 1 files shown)


yizhouxw (approved):
lgtm

PR #1309, opened 2023-12-26
enhancement: add internal-endpoint for cloud storage configuration

MarkPotato777:
#### What type of PR is this?

type-enhancement
module-objectstorage

Support internal-endpoint in cloud storage configuration. If internal-endpoint is set, then it will use internal-endpoint; otherwise, it will use public-endpoint.

Fixes #

```docs

```


server/odc-server/src/main/resources/data.sql
```diff
@@ -532,6 +532,9 @@ INSERT INTO config_system_configuration(`key`, `value`, `description`) VALUES('o
  'CHANGE_ME', '云存储服务 RegionId') ON DUPLICATE KEY UPDATE `id`=`id`;
 INSERT INTO config_system_configuration(`key`, `value`, `description`) VALUES('odc.cloud.object-storage.endpoint',
  'CHANGE_ME', '云存储服务 Endpoint') ON DUPLICATE KEY UPDATE `id`=`id`;
+UPDATE config_system_configuration SET `key` = 'odc.cloud.object-storage.public-endpoint', `description`='云存储服务公网 Endpoint' WHERE `key`='odc.cloud.object-storage.endpoint';
```
yizhouxw, line 535: insert a new record instead, copy value from previous one, default value set as empty.

in code layer, if public-endpoint/internal-endpoint not set, use value of endpoint (preivous key) .

yizhouxw (approved):
lgtm

yizhouxw (dismissed):
lgtm

yizhouxw (approved):

PR #1314, opened 2023-12-27
fix(object-storage): remove dependency on OssTaskReferManager

LuckyPickleZZ:
#### What type of PR is this?
type-bug

Now ODC would upload some task results to cloud object storage if it's available. But the `objectName` is storaged in memory of `OssTaskReferManager`. If the odc-server is reboot, these all data will be lost.

So we remove the dependency on this cache in this PR.


yhilmare (approved):
lgtm

PR #1316, opened 2023-12-27
fix(database-change): query task details throw file not found exception

### LINKED ISSUE #1315: [Bug]: View database changes task reports file path not found
### ODC version

4.2.3

### OB version

any

### What happened?

If the result file in a database change task is missing, viewing the task details will result in an error: xxx.json file not found.

Even if the file is missing, it should not report an error when viewing the work order details.

Initiate a database change task, then delete the JSON result file generated by the task, and then view the task details.

_No response_

### Cloud

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
module-database change task

Fixes #1315 

Self-test passed.

```docs

```


yhilmare (approved):
lgtm

PR #1318, opened 2023-12-27
feat(osc): add project list  adapter oms new api

krihy:
#### What type of PR is this?

type-feature
module-onlineschemachange

oms api changed, so wo should adapter the new api

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/oms/enums/ProjectTypeEnum.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package com.oceanbase.odc.service.onlineschemachange.oms.enums;
+
+/**
+ * @author yaobin
+ * @date 2023-12-27
+ * @since 4.2.4
+ */
+public enum ProjectTypeEnum {
```
smallsheeeep, line 24: What is this "Project"? I mean it should be different from ODC's project, so better to add a prefix (Such as "Oms") to specify.

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/oms/openapi/ProjectOpenApiService.java
```diff
@@ -30,6 +32,15 @@
  * @since 4.2.0
  */
 public interface ProjectOpenApiService {
```
smallsheeeep, line 34: Ditto.

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/oms/enums/ProjectTypeEnum.java
smallsheeeep: Enum suffix is no need.

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/oms/openapi/ProjectOpenApiService.java
```diff
@@ -30,6 +32,15 @@
...
+
+    /**
+     * 查询项目
+     *
+     * @param request
+     * @return
+     */
+    List<ProjectResponse> listProjects(ListProjectRequest request);
```
smallsheeeep, line 42: Is this interface currently unavailable anywhere?

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/oms/enums/ProjectTypeEnum.java
krihy: okey, done, rename to OmsProjectType

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/oms/openapi/ProjectOpenApiService.java
krihy: this class suggest reserve, it's referenced by many class and name contain `OpenApi` which can be different from odc project

krihy: done , all osc class with project* rename to OmsProject*

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/oms/openapi/ProjectOpenApiService.java
krihy: it's method be called in cloud project

smallsheeeep (approved):
LGTM

PR #1325, opened 2023-12-28
fix(database-change): query task details throw flow instance not found exception

### LINKED ISSUE #1326: [Bug]: Query database change task details throw flow instance not found exception
### ODC version

4.2.3

### OB version

any

### What happened?

When an project owner query the details of database change task created by SQL Plan. Then not found exception is thrown.

Work as usual and no exception thrown.

Use project common user create a SQL plan, and then use project owner query ticket list and query details of the database change task created by the SQL plan.

_No response_

### Cloud

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
module-ticket

When a project OWNER query the details of a database change task created by a SQL plan, then not found exception will be thrown. 
The root cause is that ODC will judge the user permission, the current user must be the creator or approver for this task, otherwise the task will not be readable. But in the latest ODC, the owner can view all tickets. This PR fix it by add a judgement: if the current user is OWNER of the project, then will skip auth.
By the way, there is a unauthorized issue in method `ApprovalPermissionService#getApprovableApprovalInstances()` and this PR fix it.

Fixes #1326

```docs

```

### DIFF
--- a/server/integration-test/src/test/java/com/oceanbase/odc/service/iam/UserServiceTest.java
+++ b/server/integration-test/src/test/java/com/oceanbase/odc/service/iam/UserServiceTest.java
@@ -699,6 +699,22 @@ public class UserServiceTest extends MockedAuthorityTestEnv {
         Assert.assertTrue(passwordEncoder.matches(request.getNewPassword(), changePasswordUser.getPassword()));
     }
 
+    @Test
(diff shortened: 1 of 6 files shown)


yhilmare (approved):
lgtm

PR #1341, opened 2024-01-03
fix(sql-check): failed to check statement when connect to a lower case schema 

### LINKED ISSUE #1337: [Bug]: Oracle mode lowercase database fails to perform database change pre-check
### ODC version

ODC 4.2.3

### OB version

-

### What happened?

I used the oracle mode lowercase database to execute the statement in the sql window successfully, but the pre-check for the database change task failed.
image

Precheck does not fail.

datasource: oracle420
database: odctest
sql: create table test(a int);
insert into test values(1);
select * from test;

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

Fixes #1337 

```docs

```


MarkPotato777 (approved):
LGTM

PR #1342, opened 2024-01-03
feat(structure-compare): structural comparison task interface layer code implementation

### LINKED ISSUE #816: [Feature]: It is hoped that ODC could support structural comparison.

In daily work, I often need to ensure that the structures of development databases and test databases used for the same business are consistent, but because too many people use these two types of databases, I cannot determine whether their structures are truly consistent.

I hope that ODC could provide a method for comparing structures so that I can periodically check for differences between these two databases.

### Additional context

_No response_

PeachThinking:
#### What type of PR is this?
type-feature
module-structure-compare

This PR only involves the implementation of the structure comparison task interface layer code and does not involve the implementation of the structure comparison service layer.
1. `FlowInstanceDetailResp` ：The current task order is associated with a `Database`. The three fields `connection`, `databaseName`, and `databaseId` in `FlowInstanceDetailResp` jointly represent a data source. In fact, a `Database`  can be used to replace these fields. The following is the modification plan：
- Use `private Database database;` instead of the above three fields to represent the database to which the task order belongs.
- Add `private Database relatedDatabase;` to represent the database associated with the task order. In the structure comparison task, it represents the target database.
```
@Data
public class FlowInstanceDetailResp {
    
    // delete：
    private InnerConnection connection;
    private String databaseName;
    private Long databaseId;
    
    // add:
    private Database database;
    private Database relatedDatabase;
```
This change has been debugged and commissioned with the front-end.

2. Added 2 new interfaces：
- `/api/v2/schema-sync/structureComparison/{taskId} `：Get a list of db structure comparison task results.
-  `/api/v2/schema-sync/structureComparison/{taskId}/{structureComparisonId}` : Get details of a specific database object structure comparison result

Fixes #816

```docs

### CODE REVIEW

File: server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/SchemaSyncController.java [hunk 1]
```diff
@@ -63,4 +67,16 @@ public ListResponse<TableComparing> skipTableComparing(@PathVariable Long id,
             @RequestBody SetSkippedReq setSkippedReq) {
         return Responses.list(shadowTableComparingService.setSkipTableComparing(id, setSkippedReq));
     }
+
+    @RequestMapping(value = "/structureComparison/{taskId}", method = RequestMethod.GET)
+    public SuccessResponse<DBStructureComparisonResult> listStructureComparisonResult(@PathVariable Long taskId,
```
LuckyPickleZZ, line 72: Why does this task require a separate controller instead of being placed in a flow related controller

PeachThinking: The existing `SchemaSyncController` is the controller of the interface for database synchronization related tasks. I think the structure comparison task is also part of the database synchronization task, so the interface for obtaining the structure comparison results is placed in `SchemaSyncController`, but the result of the structure comparison flowable task is still in the interface under `FlowInstanceController`

yhilmare (dismissed):
lgtm

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/SchemaSyncController.java
```diff
@@ -63,4 +67,16 @@ public ListResponse<TableComparing> skipTableComparing(@PathVariable Long id,
             @RequestBody SetSkippedReq setSkippedReq) {
         return Responses.list(shadowTableComparingService.setSkipTableComparing(id, setSkippedReq));
     }
+
+    @RequestMapping(value = "/structureComparison/{taskId}", method = RequestMethod.GET)
```
yhilmare, line 71: use id replace

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/model/DBObjectStructureComparisonResp.java
```diff
@@ -0,0 +1,40 @@
...
+ * @author jingtian
+ * @date 2024/1/3
+ * @since ODC_release_4.2.4
+ */
+@Data
+public class DBObjectStructureComparisonResp {
+    /**
+     * Refer to structure_comparison.id
+     */
+    private Long id;
```
yhilmare, line 33: no need to add this id

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/model/DBStructureComparisonResp.java
```diff
@@ -0,0 +1,53 @@
...
+import lombok.Data;
+
+/**
+ * @author jingtian
+ * @date 2024/1/3
+ * @since ODC_release_4.2.4
+ */
+@Data
+public class DBStructureComparisonResp {
+    private Long taskId;
```
yhilmare, line 31: us id replace

yhilmare (approved):
lgtm

PR #1345, opened 2024-01-04
fix(mock-data): failed to upload file to oss

yhilmare:
#### What type of PR is this?

type-bug

odc will shutdown mock data task when mock task is succeed or failed and mock data task will set status when upload operation is not completed, so that the upload operation will be interrupted.

Fixes #

```docs

```


tonghan-gsh (approved):
lgtm

PR #1357, opened 2024-01-08
fix(osc): osc task don't show manual swap table name when full migrate is completed

krihy:
#### What type of PR is this?

type-bug
module-online schema change
fix osc percentage continue execute when `isManualSwapTableEnabled` is true, we should update it to task result json, the swap table button show or not based on `isManualSwapTableEnabled`

```docs

```


guowl3 (approved):
LGTM

PR #1366, opened 2024-01-09
feat(database-permission): adapt permission framework

smallsheeeep:
#### What type of PR is this?
type-feature
module-permission framework

This PR adapt permission framework to DB permission control:
1. Add column `expire_time`, `authorization_type` and `ticket_id` to `iam_permission` table for supporting expire time and permission related to ticket.
2. Add `DatabasePermission` and add `query`, `change`, `export` action.
3. All permission query SQL add where clause `expire_time is null or expire_time > now()`. And expired permission will be cleared by `PermissionSchedules`.

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/metadb/iam/PermissionEntity.java
```diff
@@ -45,6 +48,7 @@
 @Data
 @Entity
 @EqualsAndHashCode(exclude = {"updateTime", "createTime"})
+@Where(clause = "expire_time is null or expire_time > now()")
```
MarkPotato777, line 51: Will it affect the global settings? What should I do if I just want to query all permissions

smallsheeeep: It will affect all query except for using customized SQL. If you want to query all permissions without any conditions, please use `findAllNoCareExpireTime()` or other customized methods.

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/PermissionSchedules.java
```diff
@@ -0,0 +1,51 @@
...
+
+    @Autowired
+    private PermissionRepository permissionRepository;
+
+    @Value("${odc.iam.permission.expired-retention-time-seconds:7776000}")
+    private Long expiredRetentionTimeSeconds;
+
+    @Scheduled(initialDelay = 10 * 1000L,
+            fixedDelayString = "${odc.iam.permission.expired-clean-interval-millis:180000}")
+    public void clearExpiredPermission() {
```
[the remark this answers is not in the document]
smallsheeeep: Even if we don't delete the expired permissions, it will not affect the existing forensics, at most it will lead to query performance degradation. The main consideration here is that we don't want to have too many expired permissions in the user permissions list, which affects the user experience. Therefore, a compromise is to retain permissions that expire within 90 days by default. Users are allowed to customize this retention period, so it is also possible to never delete.

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/PermissionSchedules.java
```diff
@@ -0,0 +1,51 @@
...
+
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * @author gaoda.xy
+ * @date 2024/1/3 11:22
+ */
+@Slf4j
+@Component
+public class PermissionSchedules {
```
yhilmare, line 35: how about deleting the expired permission when attempt to use them

server/odc-core/src/main/java/com/oceanbase/odc/core/authority/permission/DatabasePermission.java
```diff
@@ -0,0 +1,126 @@
...
+ * @author gaoda.xy
+ * @date 2024/1/2 14:19
+ */
+public class DatabasePermission extends ResourcePermission {
+
+    private static final String DATABASE_QUERY_ACTION = "query";
+    private static final String DATABASE_CHANGE_ACTION = "change";
+    private static final String DATABASE_EXPORT_ACTION = "export";
+
+    private static final int QUERY = 0x80;
```
yhilmare, line 41: the permission mask is too big, prefer smaller one

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/PermissionSchedules.java
smallsheeeep: I got you. I will delete this schedules and delete expired permission inner list database permission method (in following PR).

yhilmare (approved):
lgtm

Issue #1336, opened 2024-01-02
[Bug]: failed to approval or reject a ticket when ConnectionConfig has been changed


yhilmare:
### ODC version

4.2.3

### OB version

not related

### What happened?

1. upgrade odc from 4.2.2 to 4.2.3
2. there are several tickets pending
3. approval or reject any one of them and an exception will be thrown:
```shell
详情
预期外异常，错误详情：Unhandled exception, type=PersistenceException, message= ### Error querying database. Cause: org.flowable.common.engine.api.FlowableException: Couldn't deserialize object in variable 'connectionConfig' ### Cause: org.flowable.common.engine.api.FlowableException: Couldn't deserialize object in variable 'connectionConfig'
```

### What did you expect to happen?

approval or reject the ticket normally

### How can we reproduce it (as minimally and precisely as possible)?

follow 'what happened'

yhilmare:
flowable use `ObjectOutputStream`/`ObjectInputStream` serialize/deserialize an object and we don't set `serialVersionUID` manually so that flowable can not deserialize the object when the object has been changed

jonas303615108:
great

jonas303615108:
great

zijiacj:
great

Huangxiao-mas:
test pass

PR #1372, opened 2024-01-10
feat(database-permission): add apply database permission ticket

smallsheeeep:
#### What type of PR is this?
type-feature
module-Database Permission

This PR adapt apply database permission to current ticket module.

By the way, this PR fix a issue in apply project permission task: when user terminate task while the task is executing, then the ticket is canceled but the project permission is still also granted to the user.

Fixes #

Self-test passed by using Postman.

```docs

```


server/odc-migrate/src/main/resources/migrate/common/V_4_2_4_3__alter_iam_permission.sql
yhilmare, line 1: 4.2.4.3 is occupied in this pr #1373

server/odc-service/src/main/java/com/oceanbase/odc/service/permission/database/ApplyDatabaseFlowableTask.java
```diff
@@ -0,0 +1,251 @@
...
+        List<Long> databaseIds =
+                parameter.getDatabases().stream().map(ApplyDatabase::getId).collect(Collectors.toList());
+        Map<Long, DatabaseEntity> id2databaseEntities = databaseRepository.findByIdIn(databaseIds).stream()
+                .collect(Collectors.toMap(DatabaseEntity::getId, d -> d, (d1, d2) -> d1));
+        for (Long id : databaseIds) {
+            if (!id2databaseEntities.containsKey(id)) {
+                log.warn("Database not found, id={}", id);
+                throw new NotFoundException(ResourceType.ODC_DATABASE, "id", id);
+            }
+            if (!id2databaseEntities.get(id).getProjectId().equals(projectId)) {
```
MarkPotato777, line 244: It may cause potential NPE as databases could not belong to any project.

smallsheeeep: If it's a white screen operation, the databases that can be selected are the ones that are already associated with the project. However, in order to prevent a packet grabbing attack, it would be more reasonable to add a null value validation.

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/factory/FlowResponseMapperFactory.java
```diff
@@ -305,35 +306,31 @@ private FlowInstanceMapper generateMapper(@NonNull Collection<Long> flowInstance
         /**
          * Get Database associated with each TaskEntity
          */
-        Set<Long> databaseIds = taskId2TaskEntity.values().stream().filter(entity -> entity.getDatabaseId() != null)
```
yhilmare, line 308: why should you change these codes? is there any merging problem during the conflict solving?

yhilmare (dismissed):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/permission/database/model/ApplyDatabaseParameter.java
```diff
@@ -0,0 +1,95 @@
...
+import com.oceanbase.odc.service.permission.project.ApplyProjectParameter.ApplyProject;
+
+import lombok.Data;
+
+/**
+ * @author gaoda.xy
+ * @date 2024/1/3 13:54
+ */
+@Data
+public class ApplyDatabaseParameter implements Serializable, TaskParameters {
```
yhilmare, line 37: we'd better add serialVersionUID, there is already a related issue #1336

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/factory/FlowResponseMapperFactory.java
smallsheeeep: Yes, there is a bug of parameter validate failed. Because some type of task has no databaseId.

server/odc-service/src/main/java/com/oceanbase/odc/service/permission/database/model/ApplyDatabaseParameter.java
smallsheeeep: U are right, done!

yhilmare (approved):
lgtm

PR #1376, opened 2024-01-11
fix(dlm): the task log file does not exist

guowl3:
#### What type of PR is this?
type-bug

Fix the problem that the log file does not exist.

Fixes #

```docs

```


krihy (approved):
LGTM

PR #1377, opened 2024-01-11
feat(notification): add controller and service layer for notification

LuckyPickleZZ:
#### What type of PR is this?
type-feature

In this PR , we add controller and service layer for notification.
There is an update of migrate script, which is  column `channel_name` of table `notification_message`. This is to address the issue of channels being deleted but requiring viewing in the "message" history.
The unit test has covered all methods.


server/odc-common/src/main/java/com/oceanbase/odc/common/jpa/JsonObjectConverter.java
```diff
@@ -0,0 +1,41 @@
...
+import javax.persistence.Converter;
+
+import com.oceanbase.odc.common.json.JsonUtils;
+
+/**
+ * @author liuyizhuo.lyz
+ * @date 2024/1/10
+ */
+@Converter
+public class JsonObjectConverter implements AttributeConverter<Object, String> {
```
yizhouxw, line 31: nice tool !

server/odc-migrate/src/main/resources/migrate/common/V_4_2_4_2__alter_notification_schema.sql
```diff
@@ -67,6 +67,7 @@ CREATE TABLE IF NOT EXISTS `notification_message` (
   `title` varchar(512) NOT NULL COMMENT 'message title',
   `content` text NOT NULL COMMENT 'message content',
   `channel_id` bigint(20) NOT NULL COMMENT 'channel id, reference to notification_channel.id',
+  `channel_name` varchar(128) NOT NULL COMMENT 'channel name',
```
yizhouxw, line 70: why add channel name here ?

yizhouxw (commented):
the concept 'policy` may not suitable here,

often policy may contains a set of rule,  how about EventRoutingRule  ?

server/odc-migrate/src/main/resources/migrate/common/V_4_2_4_2__alter_notification_schema.sql
yizhouxw: got, for deleting scenario

LuckyPickleZZ:
> the concept 'policy` may not suitable here,
> 
> often policy may contains a set of rule, how about EventRoutingRule ?

We've discussed this yet. I don't think this entity should be renamed. There are these reasons:
1. Although currently each policy only represents one rule, this is determined by the front-end interaction. Our backend model does not have this limitation. If we expand our interaction in the future, we can match multiple events for each policy through `match_expression`. 
2.  In my opinion, if it is only an entity of "EventRoutingRule", it doesn't even require `match_expression` field, and only needs to do 1:1 matching according to the `event` to meet the requirements.
3. Last but not least, The APIs and models for both the front-end and back-end have been completed, and if further modifications are made, the cost will be relatively high.

yizhouxw (approved):
lgtm

PR #1378, opened 2024-01-12
fix(osc): osc job query connection config by id throw Access Denied

krihy:
#### What type of PR is this?

type-bug
module-online schema change

when osc job query connectionConfig,  we should set user info by SecurityContextUtils#setCurrentUser in current thread

fix osc job query connection config by id throw Access Denied

```docs

```


MarkPotato777 (approved):
LGTM

yhilmare (approved):
lgtm

PR #1392, opened 2024-01-15
feat(database-permission): service implementation

smallsheeeep:
#### What type of PR is this?
type-feature
module-database permission

This PR adds DB permission management implementations:
1. DB permission CRUD;
2. DB permission check when create a flow instance;
3. Delete DB permission when DB is deleted or transfered, or permission is expired.

Fixes #

DB permission check in SQL executing will be completed in next PR.

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/permission/database/DatabasePermissionService.java
```diff
@@ -42,32 +77,165 @@
...
+
     @Transactional(rollbackFor = Exception.class)
-    @PreAuthenticate(hasAnyResourceRole = {"OWNER", "DBA", "DEVELOPER", "SECURITY_ADMINISTRATOR", "PARTICIPANT"},
-            resourceType = "ODC_PROJECT", indexOfIdParam = 0)
+    @SkipAuthorize("permission check inside")
     public Page<UserDatabasePermission> list(@NotNull Long projectId, @NotNull QueryDatabasePermissionParams params,
             Pageable pageable) {
-        throw new NotImplementedException();
+        boolean hasPermission;
+        if (params.getUserId() == null || params.getUserId() != authenticationFacade.currentUserId()) {
```
yizhouxw, line 112: seems we can extract a ProjectPermissionValidator component, for simpiliy related permission verify

server/odc-service/src/main/java/com/oceanbase/odc/service/permission/database/model/DatabasePermissionType.java
```diff
@@ -40,6 +46,67 @@ public enum DatabasePermissionType implements Translatable {
...
+        if ("change".equalsIgnoreCase(action)) {
+            return DatabasePermissionType.CHANGE;
+        }
+        if ("export".equalsIgnoreCase(action)) {
+            return DatabasePermissionType.EXPORT;
+        }
+        return null;
+    }
+
+    public static DatabasePermissionType from(SqlType sqlType) {
```
yizhouxw, line 66: should add null param check as it is a public method

yizhouxw (dismissed):
lgtm

server/odc-migrate/src/main/resources/migrate/common/V_4_2_4_6__add_list_user_database_permission_view.sql
```diff
@@ -0,0 +1,46 @@
+--
+-- Add column `resource_type`, `resource_id` to `iam_permission` table
+--
+ALTER TABLE `iam_permission` ADD COLUMN `resource_type` VARCHAR ( 32 ) DEFAULT NULL COMMENT 'Resource type of the permission, the default value is null which means not assigned';
```
yhilmare, line 4: How to set historical data?

server/odc-migrate/src/main/resources/migrate/common/V_4_2_4_6__add_list_user_database_permission_view.sql
```diff
@@ -0,0 +1,46 @@
...
+ALTER TABLE `iam_permission` ADD COLUMN `resource_id` BIGINT ( 20 ) DEFAULT NULL COMMENT 'Resource ID of the permission, the default value is null which means not assigned';
```
yhilmare, line 5: prefer varchar type

server/odc-migrate/src/main/resources/migrate/common/V_4_2_4_6__add_list_user_database_permission_view.sql
```diff
@@ -0,0 +1,46 @@
...
+
+--
+-- Add `list_user_database_permission_view` view for query user database permissions list
+--
+create or replace view `list_user_database_permission_view` as
```
yhilmare, line 10: You need to consider the performance issues of this view when the amount of data is large

server/odc-service/src/main/java/com/oceanbase/odc/metadb/iam/PermissionRepository.java
```diff
@@ -92,21 +90,26 @@ Optional<PermissionEntity> findByOrganizationIdAndActionAndResourceIdentifier(Lo
...
     @Query(value = "delete from iam_permission p where p.id in (:ids)", nativeQuery = true)
     int deleteByIds(@Param("ids") Collection<Long> ids);
 
     @Modifying
     @Transactional
-    @Query(value = "delete from iam_permission p where p.expire_time is not null and p.expire_time < :expireTime",
-            nativeQuery = true)
-    int deleteByExpireTimeBefore(@Param("expireTime") Date expireTime);
+    @Query(value = "delete from iam_permission p where 1=1", nativeQuery = true)
+    void deleteAll();
```
yhilmare, line 105: why should we need this method?
`CrudRepository#deleteAll`?

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
```diff
@@ -476,6 +475,11 @@ private Map<Long, CheckState> getTeamSpaceStatus(Set<Long> ids, Map<Long, Connec
         return connId2State;
     }
 
+    @SkipAuthorize("odc internal usage")
+    public ConnectionConfig nullSafeGet(@NonNull Long id) {
```
yhilmare, line 479: why should we need this? where is the auth check?

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/ProjectPermissionValidator.java
```diff
@@ -0,0 +1,85 @@
...
+        return hasProjectRole(Collections.singleton(projectId), roleNames);
+    }
+
+    @SkipAuthorize("internal usage only")
+    public boolean hasProjectRole(@NonNull Collection<Long> projectIds, @NonNull List<ResourceRoleName> roleNames) {
+        if (projectIds.isEmpty()) {
+            return false;
+        }
+        if (roleNames.isEmpty()) {
+            return true;
```
yhilmare, line 76: why?

server/odc-service/src/main/java/com/oceanbase/odc/service/permission/database/DatabasePermissionHelper.java
```diff
@@ -0,0 +1,153 @@
...
+                toCheckDatabaseIds.add(e.getId());
+            }
+        }
+        Map<Long, Set<DatabasePermissionType>> id2PermissionTypes = getDatabaseId2PermissionTypes(databaseIds);
+        for (Long databaseId : toCheckDatabaseIds) {
+            Set<DatabasePermissionType> authorized = id2PermissionTypes.get(databaseId);
+            if (CollectionUtils.isEmpty(authorized)) {
+                throw new AccessDeniedException(String.format("No permission for the database with id %s", databaseId));
+            }
+            Set<DatabasePermissionType> unAuthorized =
```
yhilmare, line 87: What should we do if there is an implicit relationship between permissions? (Of course there is none now)

server/odc-service/src/main/java/com/oceanbase/odc/service/permission/database/DatabasePermissionService.java
```diff
@@ -42,32 +78,165 @@
...
-    @PreAuthenticate(hasAnyResourceRole = {"OWNER", "DBA", "DEVELOPER", "SECURITY_ADMINISTRATOR", "PARTICIPANT"},
-            resourceType = "ODC_PROJECT", indexOfIdParam = 0)
-    public UserDatabasePermission relinquish(@NotNull Long projectId, @NotNull Long permissionId) {
-        throw new NotImplementedException();
+    private Specification<UserDatabasePermissionEntity> getSpecificationByStatus(PermissionStatus status,
+            Date date) {
+        switch (status) {
+            case EXPIRED:
+                return UserDatabasePermissionSpec.expireTimeBefore(date);
+            case EXPIRING:
```
yhilmare, line 232: There is overlap between the definitions of not-expired and expiring

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/ProjectPermissionValidator.java
yhilmare: prefer returning false
a man without any roles can pass the permission checking, it is not reasonable~

server/odc-service/src/main/java/com/oceanbase/odc/service/permission/database/DatabasePermissionService.java
smallsheeeep: Yes, I refine it by change this enums to `ExpirationStatusFilter` for better readability.

server/odc-service/src/main/java/com/oceanbase/odc/service/permission/database/DatabasePermissionHelper.java
smallsheeeep: For now, let's keep it that way, for the sake of O(N) time complexity. If we use `Permission#implies()`, the complexity is O(N^2).

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/ProjectPermissionValidator.java
smallsheeeep: I check all invoker, you are right, I fix it.

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
smallsheeeep: It is used for filtering unauthorized databases. For only internal usage, so without any auth check.

server/odc-service/src/main/java/com/oceanbase/odc/metadb/iam/PermissionRepository.java
smallsheeeep: Because I add `@Where(clause = "expire_time > now()")` upper `PermissionEntity`. So all not customized query method will auto add this where clause for filtering expired permission. 
In this case, we want to delete all records, so I need to add this customized method.
By the way, this method only used in unit test.

server/odc-migrate/src/main/resources/migrate/common/V_4_2_4_6__add_list_user_database_permission_view.sql
smallsheeeep: I delete these two column.

server/odc-migrate/src/main/resources/migrate/common/V_4_2_4_6__add_list_user_database_permission_view.sql
smallsheeeep: The test results show that the performance will work under private cloud and will be poorer in public cloud scenarios. The performance of the authentication system is needed, but that's something for later.

server/odc-migrate/src/main/resources/migrate/common/V_4_2_4_6__add_list_user_database_permission_view.sql
smallsheeeep: This column is deleted.

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
smallsheeeep: I rename this method to `getBasicWithoutPermissionCheck` for better readability.

yhilmare (approved):
lgtm

PR #1405, opened 2024-01-16
fix(osc): exists horizontal overstep access data permission when swap table manual

krihy:
#### What type of PR is this?

type-bug
module-onlineSchemaChange
exists horizontal overstep access data permission when swap table manual

add permission valid, only flow instance  creator can operate swap table name.

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/OscService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/OscService.java
@@ -118,9 +118,13 @@ public class OscService {
         Optional<FlowInstance> optional = flowFactory.getFlowInstance(oscParameters.getFlowInstanceId());
         FlowInstance flowInstance = optional.orElseThrow(
                 () -> new NotFoundException(ResourceType.ODC_FLOW_INSTANCE, "id", oscParameters.getFlowInstanceId()));
-        permissionValidator.checkCurrentOrganization(flowInstance);
+        try {
+            permissionValidator.checkCurrentOrganization(flowInstance);
+        } finally {
+            flowInstance.dealloc();
+        }
 
-        // check permission, only creator can swap table manual
+        // check user permission, only creator can swap table manual
         PreConditions.validHasPermission(
                 Objects.equals(authenticationFacade.currentUserId(), scheduleEntity.get().getCreatorId()),
                 ErrorCodes.AccessDenied,
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/OscService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/OscService.java
@@ -124,7 +124,7 @@ public class OscService {
         PreConditions.validHasPermission(
                 Objects.equals(authenticationFacade.currentUserId(), scheduleEntity.get().getCreatorId()),
                 ErrorCodes.AccessDenied,
-                "no permission swap table name, schedule task id " + scheduleTaskId);
+                "no permission swap table.");
 
         OnlineSchemaChangeScheduleTaskResult result = JsonUtils.fromJson(scheduleTask.getResultJson(),
                 OnlineSchemaChangeScheduleTaskResult.class);
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/OscService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/OscService.java
@@ -115,16 +115,16 @@ public class OscService {
         OnlineSchemaChangeParameters oscParameters = JsonUtils.fromJson(scheduleEntity.get().getJobParametersJson(),
                 OnlineSchemaChangeParameters.class);
 
+        Optional<FlowInstance> optional = flowFactory.getFlowInstance(oscParameters.getFlowInstanceId());
+        FlowInstance flowInstance = optional.orElseThrow(
+                () -> new NotFoundException(ResourceType.ODC_FLOW_INSTANCE, "id", oscParameters.getFlowInstanceId()));
+        permissionValidator.checkCurrentOrganization(flowInstance);
+
         // check permission, only creator can swap table manual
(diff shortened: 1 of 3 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/model/OnlineSchemaChangeParameters.java
```diff
@@ -56,6 +56,7 @@ public class OnlineSchemaChangeParameters implements Serializable, TaskParameter
 
     private List<String> lockUsers;
     private SwapTableType swapTableType;
+    private Long flowInstanceId;
```
smallsheeeep, line 59: Please add `serialVersionUID` manually in this object since you have change the field.

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/OscService.java
```diff
@@ -102,6 +115,17 @@ public OscSwapTableVO swapTable(@PathVariable Long scheduleTaskId) {
         OnlineSchemaChangeParameters oscParameters = JsonUtils.fromJson(scheduleEntity.get().getJobParametersJson(),
                 OnlineSchemaChangeParameters.class);
 
+        // check permission, only creator can swap table manual
+        PreConditions.validHasPermission(
+                Objects.equals(authenticationFacade.currentUserId(), scheduleEntity.get().getCreatorId()),
+                ErrorCodes.AccessDenied,
+                "no permission swap table name, schedule task id " + scheduleTaskId);
+
+        Optional<FlowInstance> optional = flowFactory.getFlowInstance(oscParameters.getFlowInstanceId());
```
smallsheeeep, line 124: The `scheduleEntity` has a `organizationId` field and seems no need `Optional<FlowInstance>`

krihy: scheduleEntity is not implements OrganizationIsolated

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/model/OnlineSchemaChangeParameters.java
krihy: Serializable is unnecessary, object be saved as json string , do not save as object byte array

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/OscService.java
```diff
@@ -102,6 +115,17 @@ public OscSwapTableVO swapTable(@PathVariable Long scheduleTaskId) {
         OnlineSchemaChangeParameters oscParameters = JsonUtils.fromJson(scheduleEntity.get().getJobParametersJson(),
                 OnlineSchemaChangeParameters.class);
 
+        // check permission, only creator can swap table manual
+        PreConditions.validHasPermission(
+                Objects.equals(authenticationFacade.currentUserId(), scheduleEntity.get().getCreatorId()),
+                ErrorCodes.AccessDenied,
+                "no permission swap table name, schedule task id " + scheduleTaskId);
```
yizhouxw, line 122: error message should be readble, user may not undertand schedule task id here.

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/OscService.java
```diff
@@ -102,6 +115,17 @@ public OscSwapTableVO swapTable(@PathVariable Long scheduleTaskId) {
...
+
+        Optional<FlowInstance> optional = flowFactory.getFlowInstance(oscParameters.getFlowInstanceId());
+        FlowInstance flowInstance = optional.orElseThrow(
+                () -> new NotFoundException(ResourceType.ODC_FLOW_INSTANCE, "id", oscParameters.getFlowInstanceId()));
+        permissionValidator.checkCurrentOrganization(flowInstance);
```
yizhouxw, line 127: move checkCurrentOrganization to higher priority

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/model/OnlineSchemaChangeParameters.java
krihy: add `serialVersionUID` for flow save byte array

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/OscService.java
krihy: okey, done

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/OscService.java
smallsheeeep: ScheduleEntity is a entity object, so it dose not implements OrganizationIsolated. But the field ogranizationId is valid and could be used.
Anyway, your current code could work as you want.

smallsheeeep (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/OscService.java
```diff
@@ -102,6 +115,17 @@ public OscSwapTableVO swapTable(@PathVariable Long scheduleTaskId) {
         OnlineSchemaChangeParameters oscParameters = JsonUtils.fromJson(scheduleEntity.get().getJobParametersJson(),
                 OnlineSchemaChangeParameters.class);
 
+        Optional<FlowInstance> optional = flowFactory.getFlowInstance(oscParameters.getFlowInstanceId());
```
yhilmare, line 118: we should dealloc flow instance when you finish using it

krihy: okey, done

PR #1410, opened 2024-01-17
fix(data-security): data masking failed when using nesting case-when clause

### LINKED ISSUE #1391: [Bug]: nested case...when... constructs,Causing desensitization to fail
### ODC version

4.2.x

### OB version

all

### What happened?

image
nested case...when... constructs,Causing desensitization to fail

Normal desensitization

Query sensitive columns in the following manner：
SELECT column_name,
       CASE 
           WHEN condition1 THEN result1
           WHEN condition2 THEN 
               CASE
                   WHEN condition2_1 THEN result2_1
                   WHEN condition2_2 THEN result2_2
               END
           WHEN condition3 THEN result3
           ELSE default_result
       END AS alias_name
FROM table_name;

_No response_

### Cloud

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
module-data security

When executing following SQL (with nesting case-when), then masking failed due to extract sensitive column failed:
```sql
SELECT
  id,
  name,
  CASE
    WHEN birthday < '1990-01-01' THEN 1
    WHEN birthday < '2010-01-01' THEN CASE
      WHEN birthday < '2005-01-01' THEN 21
      WHEN birthday < '2010-01-01' THEN 22
    END
    WHEN birthday < '2020-01-01' THEN 3
    ELSE 4
  END AS decade
FROM
  test_data_masking_1;
```

The root cause is that the column extractor does not consider syntactic scenarios where case-when is nested within case-when. This PR fix it and add unit test.

By the way, the incorrect file formatting in `.editconfig` causing indentation configuration to not take effect, incidentally fixed.

Fixes #1391 

Self-test passed.

```docs

```


yhilmare (approved):
lgtm

PR #1411, opened 2024-01-17
fix(sql-execution): precision loss when displaying datetime type 

### LINKED ISSUE #1388: [Bug]: querying a column of type datetime(3), the result set displays a precision of only 2 bits
### ODC version

4.2.3

### OB version

3.2.4.3 （May any version）

### What happened?

ODC query for a column of type datetime(3) shows only two millisecond values, while the command line displays them normally.

Demonstrate correct accuracy.

(Ditto)

_No response_

### Cloud

_No response_

zijiacj:
#### What type of PR is this?
type-bug
module-sql execution

This  PR fixes an issue where odc displays datetime type precision loss

Fixes #1388 

```docs

```


yhilmare (approved):
lgtm

PR #976, opened 2023-12-04
feat(sql-execute): supports locating specific issue locations in multiple sqls during sql interception stage and pre-check stage

MarkPotato777:
#### What type of PR is this?

type-feature
module-sql-execute
#### Which issue(s) this PR fixes:

Fixes #359 

#### Special notes for your reviewer:

This PR mainly supports locating the specific issue locations in the whole SQL script. The main solution is to record the first character order of each SQL as the offset when splitting SQLs. To achieve this, I use `OffsetString` and `OrderChar` to wrap the primitive `String` and `char` datatype: 
1. Iterate the original SQLs and map each character to OrderChar to record the original position of each char;
2. When recognizing a SQL we use the first `OrderChar.order` of the line as the `OffsetString.offset`;

For now, we have two sets of logic to split SQLs: `SqlCommentProcessor#split` and `SqlSplitter#split`. I change the return type from` List<String>` to `List<OffsetString>` to offer the first char position of each split SQL. For scenarios where the offset is not necessary, use `SqlUtils#split` which will map `OffsetString` to primitive String for the caller to ensure that most current scenarios of splitting do not need to be modified.

#### Additional documentation e.g., usage docs, etc.:

```docs

```

PR #1414, opened 2024-01-17
fix(sql-execution): can not set a delimiter longer than 2

### LINKED ISSUE #1412: [Bug]: SQL checks containing delimiters will report an error
### ODC version

ODC 4.2.3

### OB version

-

### What happened?

image
No error will be reported when checking the sql statement of the delimiter alone, but an error will be reported when checking together with pl.

Checking it together with pl will not report an error.

delimiter //;
create function `fun_loop` (p1 int) returns int(11) begin
      declare v1 int;
      set v1 = 0;
      loop1:LOOP
      IF v1 > 10 THEN
      LEAVE loop1;
      END IF;
      set v1 = v1 + p1;
      END LOOP;
      return v1;
      end;

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

we can not set a delimiter longer than 2:

```SQL
delimiter $$$
select 1
from dual $$$
```
which cause an out of bound exception, this issue may cause by this pr #976 

Fixes #1412 

```docs

```

### DIFF
--- a/server/odc-core/src/main/java/com/oceanbase/odc/core/sql/split/SqlCommentProcessor.java
+++ b/server/odc-core/src/main/java/com/oceanbase/odc/core/sql/split/SqlCommentProcessor.java
@@ -28,7 +28,6 @@ import java.util.Objects;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


MarkPotato777 (approved):
LGTM

PR #1419, opened 2024-01-17
fix(flow): can not set task status correctly when creating task concurrently

yhilmare:
#### What type of PR is this?

type-bug

add null checking

Fixes #

```docs

```


krihy (commented):
LGTM

krihy (approved):
LGTM

PR #1420, opened 2024-01-18
feat(deployment): add datetime output for ob-odc-web.std.log

### LINKED ISSUE #1353: [Enhancement]: add datetime output for ob-odc-web.std.log

 add datetime output for ob-odc-web.std.log.

### Why is this needed?

if no datetime output in log file, it is verfy hard to trouble shooting.

yizhouxw:
#### What type of PR is this?

type-feature

sometimes odc-server may crash, ob-odc-web.std.log offer valuable information for diagnose.
however there is no datetime info in ob-odc-web.std.log. this pr add datetime.

Fixes #1353 

```docs

```

### DIFF
--- a/script/start-odc.sh
+++ b/script/start-odc.sh
@@ -54,17 +54,25 @@ function usage() {
     echo "- SERVER_PORT: deprecated environment variable, same as ODC_SERVER_PORT, for compatibility"
 }
 
+function log_info() {
+    echo 1>&2 "$(date +"%Y-%m-%dT%H:%M:%S.%Z") [INFO]" "$*"
+}
+
+function log_error() {
+    echo 1>&2 "$(date +"%Y-%m-%dT%H:%M:%S.%Z") [ERROR]" "$*"
+}
+
 function check_env_value_set() {
     local name=$1
     local value=$2
     if [ -z "$value" ]; then
-        echo "FATAL ERROR!, environment variable <${name}> not set, cannot start odc-server"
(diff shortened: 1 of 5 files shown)


script/start-odc.sh
```diff
@@ -54,17 +54,25 @@ function usage() {
     echo "- SERVER_PORT: deprecated environment variable, same as ODC_SERVER_PORT, for compatibility"
 }
 
+function log_info() {
+    echo 1>&2 "$(date +"%Y-%m-%dT%H:%M:%S.%Z") [INFO]" "$*"
```
smallsheeeep, line 58: Why do INFO level logs need to be redirected to standard error output as well?

yizhouxw: good question,  the code copy from where, here we may not redirect stdout to stderr.
fixed now.

server/odc-common/src/test/java/com/oceanbase/odc/common/security/SensitiveDataUtilsTest.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Collection;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameter;
+import org.junit.runners.Parameterized.Parameters;
+
+@RunWith(Parameterized.class)
```
smallsheeeep, line 28: 👍🏻

smallsheeeep (approved):
LGTM

PR #1427, opened 2024-01-22
feat(data-transfer): upgrade ob-loader-dumper to 4.2.8-RELEASE

LuckyPickleZZ:
Upgrade ob-loader-dumper to 2.4.8-RELEASE. There are two main changes:

1. They removed `ControlManager#register` method, but will restore it in 2.4.8.1. So we will temporarily delete the call to this method and add it back later. 
This has resulted in data-mask being temporarily unavailable.

2. Ob-loader-dumper uses `session.config.json` file to init session, we added it into `server/plugins/task-plugin-ob-mysql/src/main/resources/session.config.json`


server/plugins/task-plugin-ob-mysql/src/main/java/com/oceanbase/odc/plugin/task/obmysql/datatransfer/factory/DumpParameterFactory.java
```diff
@@ -220,7 +220,9 @@ private void setMaskConfig(DumpParameter parameter, DataTransferConfig transferC
                 controlDescription.add(function);
                 controlContext.add(controlDescription);
             }
-            controlManager.register(entry.getKey().getSchemaName(), entry.getKey().getTableName(), controlContext);
+            // TODO ob-loader-dumper will restore this method in version 4.2.8.1
+            // controlManager.register(entry.getKey().getSchemaName(), entry.getKey().getTableName(),
```
smallsheeeep, line 224: I got you!

smallsheeeep (approved):
LGTM

PR #1430, opened 2024-01-22
feat(partition-plan): add service implementation and plugin api

yhilmare:
#### What type of PR is this?

type-feat

1. add apis for plugin layer
2. add implementation for creating a partition plan

Fixes #

```docs

```


server/plugins/task-plugin-api/src/main/java/com/oceanbase/odc/plugin/task/api/partitionplan/PartitionPlanExtensionPoint.java
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * {@link PartitionPlanExtensionPoint}
+ *
+ * @author yh263208
+ * @date 2023-01-11 20:23
+ * @since ODC_release_4.2.4
+ * @see org.pf4j.ExtensionPoint
+ */
+public interface PartitionPlanExtensionPoint extends ExtensionPoint {
```
yizhouxw, line 40: 'partition plan' is a concept at odc schedule task level, the plugin may responsible for generate ddl for range partition table creation/dropping, at least there is no 'plan' here.

how about AutomaticRangePartitionExtensionPoint ?

server/plugins/task-plugin-api/src/main/java/com/oceanbase/odc/plugin/task/api/partitionplan/PartitionPlanExtensionPoint.java
```diff
@@ -0,0 +1,58 @@
...
+
+    boolean supports(@NonNull DBTablePartition dbTablePartition);
+
+    String unquoteIdentifier(@NonNull String identifier);
+
+    List<String> getCreatePartitionDdls(List<DBTablePartitionDefinition> definitions);
```
yizhouxw: generate may better than get

server/plugins/task-plugin-api/src/main/java/com/oceanbase/odc/plugin/task/api/partitionplan/PartitionPlanExtensionPoint.java
```diff
@@ -0,0 +1,58 @@
...
+
+    List<String> getDropPartitionDdls(List<DBTablePartitionDefinition> definitions);
+
+    DropPartitionGenerator getDropPartitionGeneratorByName(@NonNull String name);
+
+    PartitionExprGenerator getPartitionExpressionGeneratorByName(@NonNull String name);
+
+    PartitionNameGenerator getPartitionNameGeneratorGeneratorByName(@NonNull String name);
+
+    List<DataType> getDataTypes(@NonNull Connection connection, @NonNull DBTablePartition partition);
```
yizhouxw: I canot understand method `getDataTypes`, what is DataType means here?

server/plugins/task-plugin-api/src/main/java/com/oceanbase/odc/plugin/task/api/partitionplan/PartitionPlanExtensionPoint.java
yhilmare: done, rename to `AutoPartitionExtensionPoint`

server/plugins/task-plugin-api/src/main/java/com/oceanbase/odc/plugin/task/api/partitionplan/PartitionPlanExtensionPoint.java
yhilmare: done:
`getCreatePartitionDdls` -> `generateCreatePartitionDdls`
`getDropPartitionDdls` -> `generateDropPartitionDdls`

server/plugins/task-plugin-api/src/main/java/com/oceanbase/odc/plugin/task/api/partitionplan/PartitionPlanExtensionPoint.java
yhilmare: the code is outdated, I have rename this method to `getPartitionKeyDataTypes`

yizhouxw (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/partitionplan/PartitionPlanServiceV2.java
```diff
@@ -0,0 +1,219 @@
...
+        }
+        DBTablePartition partition = dbTable.getPartition();
+        if (partition == null || partition.getPartitionOption() == null) {
+            throw new IllegalArgumentException("Partition is null for " + schema + "." + tableName);
+        } else {
+            DBTablePartitionType partitionType = partition.getPartitionOption().getType();
+            if (partitionType == DBTablePartitionType.UNKNOWN
+                    || partitionType == DBTablePartitionType.NOT_PARTITIONED) {
+                throw new IllegalArgumentException("Table " + schema + "." + tableName + " is not partitioned");
+            } else if (autoPartitionExtensionPoint.supports(partition)) {
```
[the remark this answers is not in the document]
yhilmare: great, added

server/plugins/task-plugin-api/src/main/java/com/oceanbase/odc/plugin/task/api/partitionplan/AutoPartitionExtensionPoint.java
```diff
@@ -0,0 +1,58 @@
...
+ * @author yh263208
+ * @date 2023-01-11 20:23
+ * @since ODC_release_4.2.4
+ * @see org.pf4j.ExtensionPoint
+ */
+public interface AutoPartitionExtensionPoint extends ExtensionPoint {
+
+    boolean supports(@NonNull DBTablePartition dbTablePartition);
+
+    String unquoteIdentifier(@NonNull String identifier);
```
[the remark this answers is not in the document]
yhilmare: two differences:
1. StringUtil is just a tool, we should not use it at the service layer, cause that we don't know which dialect type is so that we don't know which method should we call (you may think that we can get dialect type and use if-else block to do that, but this is too customized).
2. we should process upper-lower case problem, that is to say, if oracle identifiers are like these `"aaa"`, `aaa`,  and we can get new identifiers by calling this method: `aaa`, `AAA`.

PR #1438, opened 2024-01-23
fix(dlm): the data cleaning task scheduling failed after editing the rate limit configuration

guowl3:
…iguration

#### What type of PR is this?
type-bug

The schedule entity will no longer be modified when updating the limiting configuration to ensure the correctness of task parameters.

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/model/ScheduleDetailResp.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/model/ScheduleDetailResp.java
@@ -175,31 +175,7 @@ public class ScheduleDetailResp implements OrganizationIsolated {
             if (datasource != null) {
                 resp.setDatasource(new InnerConnection(datasource));
             }
-            switch (entity.getJobType()) {
-                case DATA_ARCHIVE: {
-                    DataArchiveParameters parameters =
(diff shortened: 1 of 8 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/model/ScheduleDetailResp.java
```diff
@@ -163,8 +175,31 @@ public ScheduleDetailResp map(@NonNull ScheduleEntity entity) {
...
+                    break;
+                }
+                case DATA_DELETE: {
+                    DataDeleteParameters parameters =
+                            JsonUtils.fromJson(entity.getJobParametersJson(), DataDeleteParameters.class);
+                    RateLimitConfiguration rateLimitConfig = getDLMRateLimitConfigurationById.apply(entity.getId());
+                    if (rateLimitConfig != null) {
+                        parameters.setRateLimit(rateLimitConfig);
+                    }
+                    resp.setJobParameters(JsonUtils.toJson(parameters));
```
krihy, line 196: look like many duplicate code DATA_DELETE with DATA_ARCHIVE in switch

guowl3: Simplified code implementation

krihy (approved):
lgtm

PR #1456, opened 2024-01-26
feat(partition-plan): add implementation for mysql mode

yhilmare:
#### What type of PR is this?

type-mysql

1. this pr supply an implementation of partition plan for mysql mode.
2. add several test cases for plugin layer and service layer.

Fixes #

```docs

```


yhilmare:
half content of this pr is test cases

server/plugins/task-plugin-api/src/main/java/com/oceanbase/odc/plugin/task/api/partitionplan/datatype/TimeDataType.java
```diff
@@ -37,8 +41,13 @@ public class TimeDataType extends GeneralDataType {
     public static final int MINUTE = 0x10 | HOUR;
     public static final int SECOND = 0x20 | MINUTE;
```
smallsheeeep, line 42: Why `MONTH = 0x2 | YEAR`? What is the purpose of such a design? Is it used for removing excess precision?

smallsheeeep (commented):
👍🏻

server/plugins/task-plugin-api/src/main/java/com/oceanbase/odc/plugin/task/api/partitionplan/datatype/TimeDataType.java
yhilmare: partition plan is designed for time type, but the precision of a time type is different, eg.
1. date's precision is only up to year, month and day.
2. datetime's precision can be up to hour, minute and second.

so that this field is used to indicate the precision of the time type. There is a priori knowledge here: High precision must include low precision, so I adopted this design to express the inclusion relationship between precisions.

> Is it used for removing excess precision?

yes

smallsheeeep: I got you!

server/plugins/task-plugin-api/src/main/java/com/oceanbase/odc/plugin/task/api/partitionplan/util/TimeDataTypeUtil.java
```diff
@@ -0,0 +1,92 @@
...
+ * {@link TimeDataTypeUtil}
+ *
+ * @author yh263208
+ * @date 2024-01-26 10:16
+ * @since ODC_release_4.2.4
+ */
+public class TimeDataTypeUtil {
+
+    public static Date getNextDate(Date base, int interval, int intervalPrecision) {
+        Calendar calendar = Calendar.getInstance();
```
yizhouxw, line 36: java 8 time API will be easier, e.g. LocalDateTime.plus()

server/plugins/task-plugin-api/src/main/java/com/oceanbase/odc/plugin/task/api/partitionplan/util/TimeDataTypeUtil.java
```diff
@@ -0,0 +1,92 @@
...
+        } else if ((intervalPrecision & TimeDataType.YEAR) == TimeDataType.YEAR) {
+            calendar.add(Calendar.YEAR, interval);
+        } else {
+            throw new IllegalArgumentException("Interval precision is illegal, " + intervalPrecision);
+        }
+        return calendar.getTime();
+    }
+
+    public static List<Date> removeExcessPrecision(List<Date> candidates, int realPrecision) {
+        return candidates.stream().map(candidate -> {
```
yizhouxw, line 57: same with LocalDateTime.truncateTo ?

yizhouxw (commented):
nice abstraction!

server/plugins/task-plugin-api/src/main/java/com/oceanbase/odc/plugin/task/api/partitionplan/util/TimeDataTypeUtil.java
yhilmare: same as above

server/plugins/task-plugin-api/src/main/java/com/oceanbase/odc/plugin/task/api/partitionplan/util/TimeDataTypeUtil.java
yhilmare: okay, I will learn this new class. btw, I am not familiar with this now~~😂 , so can we follow this current impl now?

yizhouxw (approved):

PR #1458, opened 2024-01-26
fix(database-permission): failed to verify database permission in personal space

smallsheeeep:
#### What type of PR is this?
type-bug
module-database permission control

In perosnal space, user has all permission for all databases but there is no permission records in metaDB. So the DB permission will failed and user cannot create session. This PR fix it by skipping permission check in personal space.

Fixes #

```docs

```


yhilmare (approved):
lgtm

Issue #829, opened 2023-11-14
[Enhancement]: The desktop version of ODC currently does not support exporting objects exceeding 4GB in size.


Jane201510:
### What would you like to be added?

Export failures can occur when ODC encounters objects exceeding the maximum file size limitation imposed by the ZIP format.

### Why is this needed?

It is suggested to exclude ZIP compression from ODC exports to prevent potential failures caused by file size limitations.

yizhouxw:
released in v4.2.4

PR #1459, opened 2024-01-26
feat(data-transfer): not compress export output in client mode

LuckyPickleZZ:
#### What type of PR is this?
type-feature

Now we would compress export output to zip format no mater on the web mode or the client mode.
However, the zip format supports up to 4GB of compressed data. And we don't have a limit on exporting data size in the desktop version. So in the client mode, we will not compress it.

Feat #829


yhilmare (approved):
lgtm

PR #1464, opened 2024-01-29
fix(data-transfer): fix wrong object type names were used

### LINKED ISSUE #1333: [Bug]: Object type not displayed when exporting package body and common synonyms
### ODC version

odc423

### OB version

*

### What happened?

image

Display object types

Export package body and common synonyms

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

When use ob-loader-dumper to transfer data, we should use `ObjectType#name` as displayed name. But `ObjectType#getName` was used instead, causing miss match with front-end.

Fixes #1333


yhilmare (approved):
lgtm

PR #1465, opened 2024-01-29
fix(data-transfer): do not create os user in client mode

### LINKED ISSUE #1300: [Bug]: Ubuntu command line window cannot be opened
### ODC version

ODC4.2.3

### OB version

-

### What happened?

image
image
If the ubuntu user does not have useradd permissions, the command line window cannot be opened.

Remove this check

Open odc using a user without useradd permissions

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

When odc are running on ubuntu as client mode , we will create a new os user to run obclient. But if current os user doesn't have access to create user, an exception will be thrown and obclient could not be opened.
In fact we don't need to create a user in client mode , so I changed this assert.

Fixes #1300


yhilmare (approved):
lgtm

PR #1468, opened 2024-01-29
fix(sql-execute): do not follback execute when manual commit enabled

### LINKED ISSUE #1352: [Bug]: sql statement execute failed will result into transaction rollback
### ODC version

4.1.3~4.2.3

### OB version

3.2.3 and all other version

### What happened?

if set manual commit in personal preferences, sql statement execute failed will result into transaction rollback 

sql statement execute failed should take no affect to  transaction rollback

1. set manual commit in personal prefences.
2. open a new db session.
3. insert one row , such as 'insert into t1 values(100);'
4. execute a syntax error statements, such as 'select not_exist_col from t1;'
5. query the inserted row 'select * form t1'
6. the row inserted at step 3 should exists

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

Now when odc user has enabled manual commit, the transaction will still be rollbacked when exception occurs.
However this action is incorrect. Transactions should only end when the user manually rolls back or commits. ODC should not control transaction status.

Fixes #1352

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/session/OdcStatementCallBack.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/session/OdcStatementCallBack.java
@@ -202,7 +202,7 @@ public class OdcStatementCallBack implements StatementCallback<List<JdbcGeneralR
 
         } catch (Exception e) {
             try {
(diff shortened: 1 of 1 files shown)


yhilmare (approved):
lgtm

PR #1470, opened 2024-01-29
fix(pre-check): load uploaded files failed

### LINKED ISSUE #1469: [Bug]: Pre-check in database change task executing failed when upload SQL files
### ODC version

4.2.4

### OB version

any

### What happened?

Pre-check in database change task executing failed when upload SQL files

Pre-check successfully.

.

The reason is that the `bucket` is using wrong `userId`:
```java
this.creatorId = FlowTaskUtil.getTaskCreator(execution).getCreatorId();
```

### Cloud

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
module-pre check

A small bug, this PR fix it.

Fixes #1469 

```docs

```


yhilmare (approved):
lgtm

PR #1474, opened 2024-01-29
fix(sql-execute): executing anonymous block causes NPE in the team space

### LINKED ISSUE #1422: [Bug]: Open the SQL type that is allowed to be executed, then execute the anonymous block and an error will be reported.
### ODC version

ODC 4.2.3

### OB version

-

### What happened?

Open the SQL type that is allowed to be executed in the window specification, and then execute the anonymous block and an error will be reported.
image
image

Execute anonymous blocks without reporting errors.

Follow what happened.

_No response_

### Cloud

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug

This PR recognized executing anonymous block sql and set its sql type to `OTHERS`, otherwise, it will cause NPE.

Fixes #1422

```docs

```


yhilmare (approved):
lgtm

PR #1487, opened 2024-01-30
fix(sql-execute): fail to execute statement on OceanBase 2.2.30

### LINKED ISSUE #1475: [Bug]: When using the ob2230 version, an error occurs when executing sql.
### ODC version

ODC 4.2.3

### OB version

OB 2.2.3.0

### What happened?

When using the ob2230 version, an error occurs when executing sql.
image
image

Use ob2230 version to execute sql successfully.

Follow what happened.

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

`userenv('sessionid')` are not supported on OB 2230, so we use `OceanBaseConnection#getServerThreadId` instead.
But in `BaseSqlExecuteCallable` , the connection is generated from `SingleConnectionDataSource`. It's not instance of `OceanBaseConnection`. So we should get the real connection to get the id.

Fixes #1475

### DIFF
--- a/server/plugins/connect-plugin-ob-oracle/src/main/java/com/oceanbase/odc/plugin/connect/oboracle/OBOracleSessionExtension.java
+++ b/server/plugins/connect-plugin-ob-oracle/src/main/java/com/oceanbase/odc/plugin/connect/oboracle/OBOracleSessionExtension.java
@@ -15,6 +15,7 @@
  */
 package com.oceanbase.odc.plugin.connect.oboracle;
 
+import java.lang.reflect.Proxy;
 import java.sql.Connection;
 import java.sql.SQLException;
 import java.util.Objects;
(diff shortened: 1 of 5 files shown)


server/plugins/connect-plugin-ob-oracle/src/main/java/com/oceanbase/odc/plugin/connect/oboracle/OBOracleSessionExtension.java
```diff
@@ -67,6 +69,9 @@ public String getConnectionId(Connection connection) {
         } catch (Exception e) {
             if (connection instanceof OceanBaseConnection) {
                 connectionId = ((OceanBaseConnection) connection).getServerThreadId() + "";
+            } else if (Proxy.isProxyClass(connection.getClass())) {
```
yhilmare, line 72: OBMySQLSessionExtension has the same problem

yhilmare (approved):
lgtm

PR #1492, opened 2024-02-01
fix(taskframework): start process failed when local odc server start by java -jar

krihy:
#### What type of PR is this?

type-bug
module-taskframework

1. fix `java -jar` command start odc executor by odc job
2. fix command `SystemUtils#isProcessRunning` 
3. start odc executor set memory properties `-Xms1024m`

Fixes #

1. test in linux on mac ok
2. windows should be test again

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/ExecutorProcessBuilderFactory.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/ExecutorProcessBuilderFactory.java
@@ -34,7 +34,7 @@ import com.oceanbase.odc.service.task.util.JobUtils;
  */
 public class ExecutorProcessBuilderFactory {
 
-    private static final Pattern ODC_SERVER_EXECUTABLE_JAR = Pattern.compile("^.*odc-server-.*-executable\\.jar$");
+    private static final Pattern ODC_SERVER_EXECUTABLE_JAR = Pattern.compile("^.*odc-server-.*executable\\.jar$");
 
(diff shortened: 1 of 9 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/ExecutorProcessBuilderFactory.java
```diff
@@ -31,35 +34,46 @@
...
         options.add("-XX:+PrintGCTimeStamps");
         options.add("-XX:+PrintGCDateStamps");
+        options.add(String.format("-Xloggc:%s/gc.log", environments.get(JobEnvKeyConstants.ODC_LOG_DIRECTORY)));
         options.add("-XX:+UseGCLogFileRotation");
         options.add("-XX:GCLogFileSize=50M");
         options.add("-XX:NumberOfGCLogFiles=5");
         options.add("-XX:+ExitOnOutOfMemoryError");
         options.add("-Xmx2048m");
-        options.add("-Xms2048m");
+        options.add("-Xms1024m");
```
yizhouxw, line 76: avoid hard code heap size, what if a task require more memory in prod env ?

krihy: ok, task can set jvm memory by  `SingleJobProperties`  when  submit JobDefinition to  task-framework, this will modify in later feature pr。

yizhouxw (approved):
lgtm

PR #1495, opened 2024-02-01
fix(partition-plan):delete job failed if the associated trigger does not exist

guowl3:
…d trigger does not exist.

#### What type of PR is this?
type-bug

When deleting the operation, currently the Trigger is used to find the associated job and delete it. If the Trigger does not exist, the expected job deletion cannot be achieved.

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #806, opened 2023-11-13
feat(datasource): ODC connection module supports oracle11g

PeachThinking:
#### What type of PR is this?
type-feature
module-Datasource

#### What this PR does / why we need it:
Currently, this branch can create and synchronize Oracle data sources, create sessions, and execute SQL in the SQL window.
1. Connection module support for Oracle11g: add `connect-plugin-oracle`.
2. create new datasource support for Oracle11g:
add `sid`、`serviceName`、`userRole` column in table `connect_connection`, these columns are need for oracle to create a new connection.
3. Sync database and create database support for oracle: add `schema-plugin-oracle`.
4. Create session support oracle: make `DefaultConnectSessionFactory` adapt to oracle. query 
 session variable method in `DefaultConnectSessionFactory#setNlsFormat`  does not work with oracle mode，so a `getVariable` method is abstracted in the `SessionExtensionPoint` interface.
6. query and set `autocommit` variable method in `SessionSettingsService` does not work with oracle mode. because Oracle does not have an `autocommit` session variable, so jdbc is used to obtain and set the `autocommit` of the current connection.    I have verified that oceanbase、mysql and oralce's jdbc all support this method.
```
boolean autoCommit = connection.getAutoCommit();
connection.setAutoCommit(false);
```

7. fix do not connect to correct database when create session by database.  because oracle cannot set defaultSchema in jdbcUrl, so  we cannot get  connect schema by `JdbcUrlParser#getSchema`  in `DefaultConnectSessionFactory#registerConsoleDataSource`. In this case, oracle's `JdbcUrlParser#getSchema` will will return `null`, which means the current connection does not specify a database to connect to, so `ConnectionSessionUtil.setConnectSchema(session, null);`, and `SwitchSchemaInitializer` will switch schema to current schema.

 
#### Which issue(s) this PR fixes:

Fixes #396

#### Additional documentation e.g., usage docs, etc.:

```docs

```

PR #1520, opened 2024-02-04
feat(datasource): session management adapt for oracle11g

### LINKED ISSUE #1565: [Bug]: Obtained incorrect ddl while viewing Oracle table structure
### ODC version

4.2.4

### OB version

oracle

### What happened?

image

Show the correct ddl

Create oracle tables with index.

_No response_

### Cloud

_No response_

PeachThinking:
#### What type of PR is this?
type-feature

1. make OdcDBSession#sessionId to String data type, because oracle kill session comman is like ：
```
ALTER SYSTEM KILL SESSION 'sid,serial#';
```
image

2. set query limit  in `queryTableOrViewData` interface adapt for oracle and adaption for editor single table resultSet.
3. fix create index ddl dose not end with ";" in oracle mode.
4. Adaption for oracle timestampTZ and timestampLTZ datatype，in oracle mode, `TimeZone.getDefault()` is used as session time zone.

Fixes #806
Fixes #1565 

```docs

```


server/odc-test/src/main/java/com/oceanbase/odc/test/database/TestDBConfigurations.java
```diff
@@ -48,6 +48,9 @@ public class TestDBConfigurations {
 
     private TestDBConfigurations() {
         for (TestDBType type : TestDBType.values()) {
+            if (type == TestDBType.DORIS) {
```
yhilmare, line 51: is this expected?

PeachThinking: sorry, forget to delete it

yhilmare (approved):
lgtm

PR #1530, opened 2024-02-04
reactor(taskframework): reactor pod config about region and image name

krihy:
#### What type of PR is this?

type-reactor
module-task framework

1. get region from environments when system configuration is null
2. modify job image name provider, first get from system config, then get from environment.
3. destroy-executor-job-cron-expression` and `jdestroy-executor-job-cron-expression` cron expression set to every 1 seconds

Fixes #

```docs

```


smallsheeeep (approved):
LGTM

guowl3 (approved):
LGTM

yhilmare (approved):
lgtm

PR #1541, opened 2024-02-06
fix(taskframework): k8s system config from data.sql is empty string 

krihy:
#### What type of PR is this?

type-feature
module-taskframework

1. read system config is empty string but not null,  use `StringUtils.isNotBlank` to predicate string is blank or not
2. start pod command by start-job.sh

test passed

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #1569, opened 2024-02-20
feat(rollback-plan): adapt to task framework

smallsheeeep:
#### What type of PR is this?
type-feature
module-task framework
module-rollback plan

This PR adapt auto-generate-rollback-plan in database change task to task framework:
1. `RollbackPlanRuntimeFlowableTaskCopied` is copied from `RollbackPlanRuntimeFlowableTask` and rewritten to adapt task framework.
2. `RollbackPlanTask` is the core codes for running generate-rollback-plan task, and the related parameters are transferred through `RollbackPlanTaskParameters`.
3. Because the task framework cannot access meta DB, and have not provide a object storage abilitity, so I save or upload rollback plan result SQL files by task itself. If OSS is supported, then upload file and get a download URL. Otherwise, save the file to `${workdir}/data/ROLLBACK_PLAN/${uuid}.sql` and generate a local download URL like "/api/v2/flow/flowInstances/${flowInstanceId}/tasks/rollbackPlan/download".

Self-test passed through process mode. The cloud scenarios have not tested, and will be tested later.

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/RollbackPlanRuntimeFlowableTaskCopied.java
```diff
@@ -0,0 +1,208 @@
...
+        rollbackPlanTaskParameters.setDelimiter(p.getDelimiter());
+        return rollbackPlanTaskParameters;
+    }
+
+    /**
+     * Copy the rollback properties bean in case GSON cannot serialize Spring bean
+     * 
+     * @return A new rollback properties simple Java object
+     */
+    private RollbackProperties copyRollbackPropertiesBean() {
```
krihy, line 197: reference ObjectUtil#deepCopy

server/odc-service/src/main/java/com/oceanbase/odc/service/task/runtime/RollbackPlanTask.java
```diff
@@ -0,0 +1,276 @@
...
+    private void handleRollbackResult(String rollbackResult) {
+        if (StringUtils.isNotBlank(rollbackResult)) {
+            long totalSize = rollbackResult.getBytes().length;
+            if (totalSize > parameters.getRollbackProperties().getMaxRollbackContentSizeBytes()) {
+                log.warn("Rollback plan result file size exceeds maximum, totalSize={}, taskId={}", totalSize, taskId);
+                throw new UnsupportedSqlTypeForRollbackPlanException("Rollback plan result file size exceeds maximum");
+            }
+            try {
+                String resultFileDownloadUrl = String.format(
+                        "/api/v2/flow/flowInstances/%s/tasks/rollbackPlan/download", parameters.getFlowInstanceId());
```
krihy, line 240: multi same url path can be exacted bo JobUrlConstants

server/odc-service/src/main/java/com/oceanbase/odc/service/task/runtime/RollbackPlanTask.java
```diff
@@ -0,0 +1,276 @@
...
+                    } catch (UnsupportedSqlTypeForRollbackPlanException unsupportedSqlTypeException) {
+                        log.info(unsupportedSqlTypeException.getMessage());
+                    } catch (Exception e) {
+                        // Continue to generate rollback plan for the next sql
+                        log.warn("Failed to generate rollback plan for sql:{}, error message:{}", sql, e.getMessage());
+                    }
+                }
+                handleRollbackResult(rollbackPlans.toString());
+                this.success = true;
+            } finally {
```
krihy, line 156: it may not be nice code try catch nested

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/RollbackPlanRuntimeFlowableTaskCopied.java
smallsheeeep: The `RollbackProperties` is not a simple Java object but a Spring proxy bean. It cannot be deep copied by `ObjectUtil#deepCopy`.

server/odc-service/src/main/java/com/oceanbase/odc/service/task/runtime/RollbackPlanTask.java
smallsheeeep: These code is copied from `RollbackPlanRuntimeFlowableTask`, These code works right.

server/odc-service/src/main/java/com/oceanbase/odc/service/task/runtime/RollbackPlanTask.java
smallsheeeep: This url is used by specific task, no need to put it in `JobUrlConstants`.

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/RollbackPlanRuntimeFlowableTask.java
```diff
@@ -201,38 +196,46 @@ protected RollbackPlanTaskResult start(Long taskId, TaskService taskService, Del
...
+                    } finally {
+                        OdcFileUtil.deleteFiles(tempFile);
+                    }
+                }
+                this.resultFileId = resultFileId;
+                this.resultFileDownloadUrl = resultFileDownloadUrl;
+                this.isSuccess = true;
+                return RollbackPlanTaskResult.success(resultFileId, resultFileDownloadUrl);
+            } catch (Exception e) {
+                log.warn("Failed to put generated rollback plan file for taskId={}", getTaskId());
```
PeachThinking, line 230: It is recommended to print the stack information of the exception to the log to facilitate locating the problem.

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowTaskInstanceService.java
```diff
@@ -380,27 +381,48 @@ public List<BinaryDataResult> download(@NonNull Long flowInstanceId, String targ
...
-                        .singletonList(new ByteArrayDataResult("download.data", dispatchResponse.getContent()));
-            }
-            return Collections.singletonList(new ByteArrayDataResult(fileName, dispatchResponse.getContent()));
+            // The task is not executing on current machine, need to forward the request
+            return dispatchRequest(taskEntity);
         }
         return internalDownload(taskEntity, targetFileName);
     }
 
+    public List<BinaryDataResult> downRollbackPlanResult(@NonNull Long flowInstanceId) throws IOException {
```
PeachThinking, line 390: Is there any authentication logic required here?

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/RollbackPlanRuntimeFlowableTask.java
smallsheeeep: Yes you are right, I fix it.

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowTaskInstanceService.java
smallsheeeep: Same as other file download API, there is no permission check yet. The URL is added in `OrganizationAuthenticationInterceptor` as a whitelist.

PeachThinking:
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/DatabaseChangeRuntimeFlowableTaskCopied.java
```diff
@@ -123,13 +131,14 @@ protected DatabaseChangeResult start(Long taskId, TaskService taskService, Deleg
         }
         jobEntity = taskFrameworkService.find(jobId);
         result = JsonUtils.fromJson(jobEntity.getResultJson(), DatabaseChangeResult.class);
+        result.setRollbackPlanResult(rollbackPlanTaskResult);
         JobStatus status = jobEntity.getStatus();
         isSuccessful = status == JobStatus.DONE;
         isFailure = !isSuccessful;
+        taskService.succeed(taskId, result);
```
krihy, line 138: task may be failed here

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/RollbackPlanRuntimeFlowableTaskCopied.java
krihy: use `RollbackProperties rollbackProperties = new RollbackProperties()` is normal bean, here should be changed if  RollbackProperties add property, 
btw: 
why need copy a new bean in `rollbackPlanTaskParameters.setRollbackProperties(copyRollbackPropertiesBean())`;

smallsheeeep: The autowired `RollbackProperties` is a Spring proxy bean, which cannot be serialized by GSON. So I `new` a new Java object for this class. I have try to use `ObjectUtil#deepCopy`, but this method return because this method depends on JsonUtil and Fastjson cannot serialize this Spring bean. So the right and easiest solution is new a object and copy properties.

krihy (commented):
LGTM

yhilmare (approved):
lgtm

PR #1578, opened 2024-02-20
feat(databaseChange): Automatically adjust the timeout if the database change task involves time-consuming index change operations

### LINKED ISSUE #772: [Feature]: Auto-recognize time-consuming SQL and prompts when creating database change task

I create a database change task in which create a index for big table and consume long time (> 48h). 
Then the task will timeout and failed, the index may rollback.

May create a new SQL check rule to delect create index operation and prompt user to increase the task timeout?

### Additional context

_No response_

PeachThinking:
#### What type of PR is this?
type-feature
module-database change

Automatically adjust the timeout if the database change task involves time-consuming index change operations.
1. create `IndexChangeTimeConsumingExists` to check if sqls involve time-consuming operations like creating index and modifying primary key. This SQL check result should not be displayed to the user, so this check result will be filtered out at the end.
2. if involves, `SqlCheckTaskResult#timeConsumingSqlExists` will be true, and `DatabaseChangeResult#isAutoModifyTimeout` will be true.

`ServiceTaskInstanceRepository#findByFlowInstanceId` is a bug, it would return multiple rows of data，But this method is not actually called. Delete this method.

Fixes #772 

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/DatabaseChangeRuntimeFlowableTask.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/DatabaseChangeRuntimeFlowableTask.java
@@ -18,6 +18,7 @@ package com.oceanbase.odc.service.flow.task;
 import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 
+import org.apache.commons.lang3.Validate;
 import org.flowable.engine.delegate.DelegateExecution;
 import org.springframework.beans.factory.annotation.Autowired;
 
@@ -28,8 +29,11 @@ import com.oceanbase.odc.core.session.ConnectionSessionUtil;
 import com.oceanbase.odc.core.shared.Verify;
 import com.oceanbase.odc.core.shared.constant.DialectType;
 import com.oceanbase.odc.core.shared.constant.FlowStatus;
+import com.oceanbase.odc.core.shared.constant.ResourceType;
+import com.oceanbase.odc.core.shared.exception.NotFoundException;
 import com.oceanbase.odc.core.sql.split.SqlCommentProcessor;
 import com.oceanbase.odc.metadb.task.TaskEntity;
+import com.oceanbase.odc.metadb.task.TaskRepository;
 import com.oceanbase.odc.service.connection.model.ConnectProperties;
 import com.oceanbase.odc.service.connection.model.ConnectionConfig;
 import com.oceanbase.odc.service.datasecurity.DataMaskingService;
@@ -37,8 +41,10 @@ import com.oceanbase.odc.service.datasecurity.accessor.DatasourceColumnAccessor;
 import com.oceanbase.odc.service.flow.exception.ServiceTaskCancelledException;
 import com.oceanbase.odc.service.flow.exception.ServiceTaskError;
 import com.oceanbase.odc.service.flow.exception.ServiceTaskExpiredException;
+import com.oceanbase.odc.service.flow.model.PreCheckTaskResult;
 import com.oceanbase.odc.service.flow.task.model.DatabaseChangeParameters;
 import com.oceanbase.odc.service.flow.task.model.DatabaseChangeResult;
+import com.oceanbase.odc.service.flow.task.model.FlowTaskProperties;
 import com.oceanbase.odc.service.flow.task.model.RollbackPlanTaskResult;
 import com.oceanbase.odc.service.flow.util.FlowTaskUtil;
 import com.oceanbase.odc.service.objectstorage.ObjectStorageFacade;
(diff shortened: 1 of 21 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/DatabaseChangeRuntimeFlowableTaskCopied.java
```diff
@@ -218,4 +232,26 @@ private JobDefinition buildJobDefinition(DelegateExecution execution) {
                 .build();
     }
 
+    private void modifyTimeoutIfInvolveIndexChange(DelegateExecution execution, DatabaseChangeParameters parameters,
+            DatabaseChangeTaskParameters taskParameters, TaskEntity taskEntity) {
+        Long preCheckTaskId = FlowTaskUtil.getPreCheckTaskId(execution);
+        TaskEntity preCheckTask = taskRepository.findById(preCheckTaskId)
```
krihy, line 238: why not use TaskService#detail to query

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/DatabaseChangeRuntimeFlowableTask.java
```diff
@@ -193,4 +208,23 @@ private DatabaseChangeThread generateOdcAsyncTaskThread(Long taskId, DelegateExe
         return returnVal;
     }
 
+    private void modifyTimeoutIfInvolveIndexChange(DelegateExecution execution, DatabaseChangeParameters parameters) {
+        Long taskId = FlowTaskUtil.getTaskId(execution);
+        Long preCheckTaskId = FlowTaskUtil.getPreCheckTaskId(execution);
+        TaskEntity preCheckTask = taskRepository.findById(preCheckTaskId)
```
krihy, line 214: same as above

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/model/SqlCheckTaskResult.java
```diff
@@ -48,9 +48,10 @@ public class SqlCheckTaskResult implements Serializable, FlowTaskResult {
     private String error;
     private String fileName;
     private List<CheckResult> results = new ArrayList<>();
+    private boolean involveIndexChange;
```
krihy, line 51: suggest: involveIndexChange -> isInvolveIndexChanged

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/IndexChangeTimeConsumingExists.java
```diff
@@ -0,0 +1,76 @@
...
+import lombok.NonNull;
+
+/**
+ * {@link IndexChangeTimeConsumingExists}
+ *
+ * @author jingtian
+ * @date 2024/2/19
+ * @since ODC_release_4.2.4
+ */
+public class IndexChangeTimeConsumingExists implements SqlCheckRule {
```
krihy, line 43: suggest add test case for IndexChangeTimeConsumingExists

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/DatabaseChangeRuntimeFlowableTaskCopied.java
PeachThinking: you are right! I have modified the code

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/model/SqlCheckTaskResult.java
PeachThinking: -> involveTimeConsumingSql

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/SqlCheckService.java
```diff
@@ -94,7 +94,7 @@ public List<CheckResult> check(@NotNull ConnectionSession session,
     }
 
     public List<CheckViolation> check(@NotNull Long environmentId, @NonNull String databaseName,
-            @NotNull List<OffsetString> sqls, @NotNull ConnectionConfig config) {
+            @NotNull List<OffsetString> sqls, @NotNull ConnectionConfig config, List<SqlCheckRule> defaultRules) {
```
smallsheeeep, line 97: How to understand the `defaultRules`? Additional rules? Why not create a new method such as `public List<CheckViolation> check(@NotNull Long environmentId, @NonNull String databaseName, @NotNull List<OffsetString> sqls, @NotNull ConnectionConfig config, List<SqlCheckRule> rules)`. In this method, you can specify what rules you want to use.

PeachThinking: you are right! Done.

smallsheeeep (approved):
LGTM

yhilmare (approved):
lgtm

PR #1592, opened 2024-02-21
fix(database-permission): DB permission interceptor invalid when executing PL in the SQL console

### LINKED ISSUE #1581: [Bug]: There is no database change permission, and the creation of pl objects is not intercepted.
### ODC version

ODC 4.2.4

### OB version

-

### What happened?

image
There is no database change permission, and the creation of pl objects is not intercepted.

Without database change permissions, creating pl objects should be intercepted and prompted to apply for database permissions.

1. Turn on the switch that allows the sql window to create pl
2. The user does not have permission to change the database
3. create function `fun_tinyint1` (`p1` tinyint(4)) returns tinyint(4) begin      return p1; end;

_No response_

### Cloud

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
module-database permission

When user has no DB `change` permission, and executing a PL/SQL in SQL  console such as:
```sql
CREATE FUNCTION `gaoda_test_data`.`func2` (
	`str1` VARCHAR ( 45 ),
	`str2` VARCHAR ( 45 )) RETURNS VARCHAR ( 128 ) BEGIN
RETURN ( SELECT concat( str1, str2 ) FROM DUAL );
END;
```
Then the DB permission interceptor will be invalid, the SQL will be executed.

The root cause is that the `SchemaExtractor` does not care about PL/SQL, this PR add these solution and fix this bug.

Fixes #1581 

Slef-test passed. 
Note that if the database is accessed from within PL, it may not be detected.

```docs

```


yhilmare (approved):
lgtm

PR #1494, opened 2024-02-01
feat(datasource): support new datasource of doris

ISADBA:
#### What type of PR is this?
type-feature

Optionally add one or more of the following kinds if applicable:
datasource

#### What this PR does / why we need it:
Support new datasource of Doris

#### Which issue(s) this PR fixes:
Fixes #1448

#### Special notes for your reviewer:
have frontend pr

#### Additional documentation e.g., usage docs, etc.:

```docs
Support adding Doris as a separate data source to the ODC data source, facilitating customization of Doris data source-related functionalities.

Support adding Doris schema to the project.

Support viewing Doris databases, tables, and table information in the SQL window, enabling data queries through the SQL window.

Doris adapts to work order functionalities:
a. Supports export.
b. Supports exporting result sets.
c. Supports SQL work orders (DDL, DML).
d. Supports SQL plans.

Doris adapts to security capabilities:
a. Supports 8 built-in SQL window validation rules.
b. Supports 3 built-in DML-related SQL check rules (invalid WHERE condition, no WHERE condition, syntax error).
c. Supports library-level permission validation.
```

PR #1594, opened 2024-02-22
fix(mock-data): failed to mock any data for ob-mysql mode

yhilmare:
#### What type of PR is this?

type-bug

this issue is involved by this pr #1494 
and this pr set an error dialect type when ob-mysql

Fixes #

```docs

```


smallsheeeep (approved):
LGTM

PR #1599, opened 2024-02-22
enhancement(unit-test): on-demand loading of unit test dependent data sources

### LINKED ISSUE #1598: [Enhancement]: On-demand loading of unit test dependent data sources

ODC support multiple type of database system. But in current unit test framework, it load all type database environment at one time, if I do not have some available database, then the unit test cannot start even though I only need one database.

### Why is this needed?

On-demand loading of unit test dependent data sources.

smallsheeeep:
#### What type of PR is this?
type-enhancement

Enhancement the unit test framework.

Fixes #1598 

```docs

```


yhilmare (approved):
lgtm

PR #1601, opened 2024-02-22
fix(taskframework):  use default entry point to replace init cmd 

krihy:
#### What type of PR is this?

type-bug
module-taskframework
1.task executor use start-odc.sh and when environment contain `ODC_BOOT_MODE` and value is `TASK_EXECUTOR`, pod will be started as task executor
2.fix log query NPE when job is `RUNNING` and log bucket name and object id is not uploaded at this period time

Fixes #

```docs

```


guowl3 (dismissed):
Write only task framework internal logs to executor.log

krihy:
> Write only task framework internal logs to executor.log

done

smallsheeeep (approved):
LGTM

PR #1606, opened 2024-02-23
feat(dlm): supports review sql

guowl3:
#### What type of PR is this?
type-feature

It is currently impossible to view the filtering conditions after ODC parsing when creating a dlm task, and there is a certain risk. This PR provides a SQL preview function to help customers understand the SQL to be executed.

Fixes #

```docs

```

### DIFF
--- a/server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DLMController.java
+++ b/server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DLMController.java
@@ -15,18 +15,16 @@
  */
 package com.oceanbase.odc.server.web.controller.v2;
 
-import java.util.List;
-
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RestController;
 
+import com.oceanbase.odc.service.common.response.ListResponse;
 import com.oceanbase.odc.service.common.response.Responses;
-import com.oceanbase.odc.service.common.response.SuccessResponse;
 import com.oceanbase.odc.service.dlm.DLMService;
-import com.oceanbase.odc.service.dlm.model.GetRealSqlListReq;
+import com.oceanbase.odc.service.dlm.model.PreviewSqlStatementsReq;
 
 /**
  * @Author：tinker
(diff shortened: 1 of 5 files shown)


server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DLMController.java
```diff
@@ -0,0 +1,48 @@
...
+
+@RestController
+@RequestMapping("/api/v2/dlm")
+public class DLMController {
+
+    @Autowired
+    private DLMService dlmTaskPrepareService;
+
+    @RequestMapping(value = "/getRealSqlList", method = RequestMethod.POST)
+    public SuccessResponse<List<String>> getRealSqlList(@RequestBody GetRealSqlListReq req) {
```
yizhouxw, line 45: getRealSqlList, seems hard to understand,

how about previewSqlStatements

smallsheeeep, line 45: Why not use `ListResponse`?

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DLMController.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+@RestController
+@RequestMapping("/api/v2/dlm")
+public class DLMController {
+
+    @Autowired
+    private DLMService dlmTaskPrepareService;
+
+    @RequestMapping(value = "/getRealSqlList", method = RequestMethod.POST)
```
smallsheeeep, line 44: Better to use `GET` ?

guowl3: The GET method cannot use @RequestBody, but the request parameters are more complex.

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DLMController.java
guowl3: modified function name from getRealSqlList to previewSqlStatements.

guowl3: I've updated the function to return ListResponse as suggested.

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DLMController.java
smallsheeeep: I new tips for me, Thank you!

smallsheeeep (approved):
LGTM

PR #1610, opened 2024-02-25
fix(cloud): add a CacheManager bean which allows null values

MarkPotato777:
#### What type of PR is this?

type-bug

In some scenery, the business logic needs a null-free cache, so I added an `allowNullCacheManager` in this PR.

Fixes #

```docs

```


yizhouxw (approved):
lgtm

PR #1611, opened 2024-02-26
feat(osc): support add column in new ddl

krihy:
type-feature
module-online schema change

origin ddl is 
`
CREATE TABLE `a3` (
  `a` int(11) NOT NULL,
  `c` varchar(110) DEFAULT NULL,
  UNIQUE KEY `f` (`a`) BLOCK_SIZE 16384 LOCAL
);
`
support add column on `a3`, test passed

`alter table a3 add column   `b` int(11) DEFAULT NULL;
`

```docs

```


guowl3 (approved):
LGTM

PR #1612, opened 2024-02-26
feat(taskframework): support modify job parameters

krihy:
#### What type of PR is this?

type-feature
module-taskframework
job parameters can be modified when job is not terminated

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/BaseJobCaller.java
```diff
@@ -142,6 +144,51 @@ protected void afterStopFailed(JobIdentity ji, Exception e) throws JobException
...
+            throw new JobException("Modify job parameters not succeed, jobId={0}, response={1}", ji.getId(),
+                    JsonUtils.toJson(e.getMessage()));
+        }
+    }
+
+    private String getExecutorPoint(JobIdentity ji, JobConfiguration jobConfiguration, JobEntity jobEntity)
+            throws JobException {
+        String executorEndpoint = jobEntity.getExecutorEndpoint();
+        if (executorEndpoint == null) {
+            try {
```
guowl3, line 176: I don't think retrying is necessary

server/odc-service/src/main/java/com/oceanbase/odc/service/task/schedule/JobScheduler.java
```diff
@@ -45,15 +47,24 @@ public interface JobScheduler {
...
     void cancelJob(Long jobId) throws JobException;
 
+    /**
+     * modify job parameters, new parameters will be delivery to executor if job is running
+     *
+     * @param jobId job id
+     * @param jobParameters job parameters
+     * @throws JobException throw JobException if job is terminated
+     */
+    void modifyJobParameters(Long jobId, Consumer<Map<String, String>> jobParameters) throws JobException;
```
guowl3, line 57: The input parameters of the method are a bit complicated.

server/odc-service/src/main/java/com/oceanbase/odc/service/task/util/JobUtils.java
```diff
@@ -69,6 +72,22 @@ public static <T> T fromJson(String json, Class<T> clazz) {
         return new Gson().fromJson(json, clazz);
     }
 
+    public static <T> T fromJson(String json, Type type) {
```
guowl3, line 75: Is there a better solution?

server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/BaseJobCaller.java
krihy: okey, done

server/odc-service/src/main/java/com/oceanbase/odc/service/task/schedule/JobScheduler.java
krihy: modify method parameter, remove consumer

server/odc-service/src/main/java/com/oceanbase/odc/service/task/util/JobUtils.java
krihy: this is solution is more easier

guowl3 (approved):
LGTM

PR #1613, opened 2024-02-26
fix(osc): osc will be failed if check oms step accumulate failed time bigger than threshold

krihy:
#### What type of PR is this?

module-osc
type-bug
1.guarantee osc task can be completed, handle osc status to failed if check failed times bigger than threshold, threshold default value is `120`.

test passed

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/pipeline/ProjectStepResultChecker.java
```diff
@@ -139,7 +140,16 @@ private void checkPreCheckStepResult() {
         }
     }
 
-    private boolean checkProjectFinished() {
+    private boolean isProjectReleased() {
```
yizhouxw, line 143: delete, releasing status mapping to released, seems hard to understand ?

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/pipeline/ProjectStepResultChecker.java
```diff
@@ -181,7 +191,14 @@ private boolean checkStepFailed() {
                 if (currentProjectStepMap.get(stepName).getExtraInfo() != null) {
                     checkerResult.setErrorMsg(currentProjectStepMap.get(stepName).getExtraInfo().getErrorMsg());
                 }
-                isProjectFailed = true;
+
+                int failedTimes = checkFailedTimes.compute(stepName, (k, v) -> v == null ? 1 : ++v);
```
yizhouxw, line 195: what is the check interval? 
if the check interval set as a small value, then we may need change the threshold, how to maintains threshold setting?l

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/subtask/OmsResourceCleanHandler.java
```diff
@@ -46,16 +47,24 @@ public boolean checkAndReleaseProject(OmsProjectControlRequest projectControl) {
...
         OmsProjectControlRequest controlRequest = new OmsProjectControlRequest();
         controlRequest.setId(projectControl.getId());
         controlRequest.setUid(projectControl.getUid());
-        OmsProjectProgressResponse progressResponse = projectOpenApiService.describeProjectProgress(controlRequest);
-        if (progressResponse.getStatus() == OmsProjectStatusEnum.RELEASED ||
-                progressResponse.getStatus() == OmsProjectStatusEnum.RELEASING ||
-                progressResponse.getStatus() == OmsProjectStatusEnum.DELETED) {
-            released = true;
+        try {
+            OmsProjectProgressResponse progressResponse = projectOpenApiService.describeProjectProgress(controlRequest);
```
yizhouxw, line 54: duplicatd relased judge logic, how about add a method in OmsProjectStatuaEnum ?

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/pipeline/ProjectStepResultChecker.java
krihy: Current only encounter  `FULL_TRANSFER` may cause oms project be blocked during data transfer, and the failed result cannot be recovered in general,  the threshold times is enough.

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/pipeline/ProjectStepResultChecker.java
krihy: ok, these status can mapping to `Destroyed`

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/pipeline/ProjectStepResultChecker.java
krihy: use 120 s to record `FULL_TRANSFER` failed

server/odc-server/src/main/resources/data.sql
```diff
@@ -675,6 +675,10 @@ INSERT INTO config_system_configuration(`key`, `value`, `description`) VALUES('o
   'CHANGE_ME', 'oms region')
   ON DUPLICATE KEY UPDATE `id`=`id`;
 
+INSERT INTO config_system_configuration(`key`, `value`, `description`) VALUES('odc.osc.oms.check-project-step-failed-threshold-time',
```
yizhouxw, line 678: odc.osc.oms.check-project-step-failed-timeout-seconds

krihy: Okey, done

LuckyPickleZZ (approved):
LGTM

PR #1614, opened 2024-02-26
fix(dlm):create target table failed

guowl3:
#### What type of PR is this?
type-bug

Create target table failed because of  invalid cursor position

Fixes #

```docs

```


krihy (approved):
LGTM

PR #1619, opened 2024-02-26
fix(database-permission): delete related permission records when deleting data source

### LINKED ISSUE #1618: [Bug]: No cleanup of related database permission records when deleting a data source
### ODC version

4.2.4

### OB version

any

### What happened?

When I delete a datasource, then all database are deleted, but the records in table `iam_permission` still contains these deleted database info.

Delete related database permission records when deleting a data source.

Grant DB permission to a user, then delete the related datasource, look the meta DB for iam_permission table.

_No response_

### Cloud

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
module-database permission

When a data source is deleted, then the related database permission records should be deleted. This PR just do it.

Fixes #1618 

Self-test passed

```docs

```


yhilmare (approved):
lgtm

PR #1621, opened 2024-02-26
enhancement(sql-rule): optimize the rule 'not-allow PL debug''s description

### LINKED ISSUE #1339: [Enhancement]: SQL window specification Adjustments

 PL debugging is not allowed by default in a production environment.If a user enables PL debugging, alert them to the potential risks.
提示示例：PL 调试会有工作线程卡住的风险，恐影响生产系统的稳定，不建议启用 PL 调试。

image

### Why is this needed?

In the debugging scenario for OB PL, initiating debugging may cause the working thread to become stuck.
It is not recommended to enable PL debugging in a production environment.

MarkPotato777:
#### What type of PR is this?

type-enhancement

PL debug on OceanBase maybe a high-level risk action, so I updated the description of this rule.

Fixes #1339

```docs

```


yhilmare (approved):
lgtm

PR #1625, opened 2024-02-26
fix(integration): garbled code exists when using Chinese in request body

### LINKED ISSUE #1624: [Bug]: Garbled code exists when using Chinese in request body of external integration configuration
### ODC version

4.2.4

### OB version

any

### What happened?

When I create a external integration config such as:
```yaml
approvalTimeoutSeconds: 86400
http:
  connectTimeoutSeconds: 5
  socketTimeoutSeconds: 30
api:
  start:
    # Request method, supported values: GET | POST | PUT | PATCH
    method: POST
    # Request URI
    url: http://localhost:45678/start
    # Request headers, the format is: <key: value>
    headers:
      # Sample of headers
      Content-Type: application/json;charset=UTF-8
      Accept: application/json
    # Request query parameters, the format is: <key: value>
    queryParameters: ~
    # Request body
    body:
      # Type of request body, supported values: FORM_DATA | RAW
      type: RAW
      # Content of request body, the format is: <String> for RAW type or <key: value> for FORM_DATA type
      # Sample of request content in type of RAW
      content: |-
        {
          "userId": "${user.id}",
          "riskLevel": "高风险",  # using Chinese
          "userName": "${user.account}",
          "taskType": "${task.type}",
          "connection": "${connection.name}",
          "tenant": "${connection.tenant}",
          "sqlContent": "${sql.content}",
          "taskDetails": "${task.details}"

        }
    # Mark whether the request body is encrypted
    requestEncrypted: false
    # Expression to judge whether the request is successful based on the response body analysis, using the SPEL syntax
    requestSuccessExpression: '[success] == true'
    # Mark whether the response body is encrypted
    responseEncrypted: false
    # Expression to extract ID of process instance created by the external system based on the response body analysis, using the SPEL syntax
    extractInstanceIdExpression: '[content][processInstanceId]'
```
Then the actual request body exists garbled code such as:
image

Avoid exist garbled code.

Ditto.

_No response_

### Cloud

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
type-external integration

There is a bug: Garbled code exists when using Chinese in request body of external integration configuration.

Root cause: When the content-type is RAW, then we build HTTP request entity through `EntityBuilder.create().setText(encryptedBody).build()`. Then the default charset is `ISO_8859_1` rather than `UTF_8` causing garbled code for Chinese.

This PR fix it by using `StringEntity(bodyString, StandardCharsets.UTF_8)`.

Fixes #1624 

Self-test Passed.

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/integration/HttpOperationService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/integration/HttpOperationService.java
@@ -16,19 +16,19 @@
 package com.oceanbase.odc.service.integration;
 
 import java.io.UnsupportedEncodingException;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
 
-import org.apache.http.HttpEntity;
 import org.apache.http.NameValuePair;
 import org.apache.http.client.config.RequestConfig;
-import org.apache.http.client.entity.EntityBuilder;
 import org.apache.http.client.entity.UrlEncodedFormEntity;
 import org.apache.http.client.methods.HttpUriRequest;
 import org.apache.http.client.methods.RequestBuilder;
+import org.apache.http.entity.StringEntity;
 import org.apache.http.message.BasicNameValuePair;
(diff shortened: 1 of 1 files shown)


yizhouxw (approved):
lgtm

PR #1636, opened 2024-02-27
merge from main into dev/4.2.x

guowl3:

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowTaskInstanceService.java
krihy, line 1: getLogByTaskFramework method in FlowTaskInstanceService is not used

krihy (approved):
LGTM

PR #1641, opened 2024-02-27
fix(security): share public and private key when ODC is deployed on multiple nodes

### LINKED ISSUE #664: [Bug]: When ODC is deployed in a multi-node environment, new datasource connections cannot be created.
### ODC version

ob cloud

### OB version

all

### What happened?
When ODC is deployed in a multi-node environment, new datasource connections cannot be created
image

The cause is：
 The request routing rule in OB-Cloud is to reach the same ODC based on uid, but the uid is not sent when the api/v1/info interface requests, making it impossible to ensure that api/v1/info and create session interface are sent to the same ODC.
image

Multi-cloud environment ODC can create connections normally when deploying multiple nodes

In Multi-cloud environment， create new datasource when odc deploying multiple nodes

_No response_

### Cloud

_No response_

PeachThinking:
#### What type of PR is this?
type-bug
module-security

When ODC is deployed on multiple nodes, the public and private keys are shared by metaDB:

- If ODC can obtain the public and private keys from metaDB at startup, it will directly use the keys configured in metaDB.

- If ODC cannot obtain the public and private keys from metaDB at startup, it will preemptively write the keys to metaDB using a distributed lock. If it fails to acquire the lock, it will attempt to read the keys from metaDB again.

Fixes #664 

```docs

```


smallsheeeep (approved):
LGTM

yhilmare (approved):
lgtm

PR #1644, opened 2024-02-27
fix(config): wrong reason while full link trace not enabled

### LINKED ISSUE #1616: [Feature]: User settings center for several self defined configuration

User settings center for several self defined configuration 

new user confgifation API , new user configuraiton GUI 

### Additional context

_No response_

yizhouxw:
#### What type of PR is this?

type-bug

fix some problems foudn during debug.

1. The value for language modification doesn't match between frontend and backend: the frontend uses hyphens (-), while the backend employs underscores (_). For example, if the frontend sets 'en-US', the corresponding backend locale value would be 'en_US'.

2. fter reverting to default configurations, the shortcut keys are not F8 and F9.

3. When End-to-end TRACE is turned off, it indeed becomes inaccessible; however, the prompt message still reads "Full-stack tracing is only available on OceanBase versions higher than 4.0." This should be modified to state that "Full link trace is currently not enabled."

4. There's an error where the default value for continuing execution after an error is incorrect: it should be False but currently is set to True.

Fixes #1616 

```docs

```

### DIFF
--- a/server/odc-core/src/main/java/com/oceanbase/odc/core/shared/constant/ErrorCodes.java
+++ b/server/odc-core/src/main/java/com/oceanbase/odc/core/shared/constant/ErrorCodes.java
@@ -233,6 +233,7 @@ public enum ErrorCodes implements ErrorCode {
     ObGetPlanExplainEmpty,
     ObGetFullLinkTraceFailed,
     ObFullLinkTraceNotSupported,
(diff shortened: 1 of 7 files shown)


yhilmare (approved):
lgtm

yhilmare (approved):
lgtm

PR #1646, opened 2024-02-27
fix(ldap):ldap not have login attempt limit

### LINKED ISSUE #1587: [Bug]: LDAP login needs to remove the five login limit
### ODC version

ODC424

### OB version

independent

### What happened?

LDAP login needs to remove the five login limit
image

remove this restriction

chat privately

_No response_

### Cloud

_No response_

ungreat:
#### What type of PR is this?
Added a limit for the number of failed login attempts for LDAP integration, and this limit on failed login attempts is isolated from the limit for local password login attempts. Furthermore, the limit will be reset after a successful login.

Fixes #1587  Instead of remove login failed message, add login apptempt times to LDAP integrate config.

some change:
`FailedLoginAttemptLimiter`: add method `attemptFailedByException` to  increase `failedAttempt` by exception, which can avoid determine the type of exception to `reduceFailedAttemptCount` in `CustomAuthenticationFailureHandler#onAuthenticationFailure`.This allows FailedLoginAttemptLimiter transparent to CustomAuthenticationFailureHandler.

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/iam/auth/AttemptableUsernamePasswordAuthenticationToken.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/iam/auth/AttemptableUsernamePasswordAuthenticationToken.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2023 OceanBase.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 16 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/iam/auth/AttemptableUsernamePasswordAuthenticationToken.java
```diff
@@ -0,0 +1,50 @@
...
+package com.oceanbase.odc.service.iam.auth;
+
+import java.util.Collection;
+
+import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
+import org.springframework.security.core.GrantedAuthority;
+
+import lombok.Getter;
+
+public class AttemptableUsernamePasswordAuthenticationToken extends UsernamePasswordAuthenticationToken {
```
MarkPotato777, line 25: Your solution is kind of complex for me, could you please kindly add some detail and explanation in your PR description? For example, why do we need another authentication token?

ungreat: AttemptableUsernamePasswordAuthenticationToken is use to distinguish whether need to reset the login limit after login successful.Let me add a few more details ~

ungreat:
It's not necessary to implement this feature at the moment.

Issue #1675, opened 2024-02-29
[Enhancement]: Provide internationalized error messages when lacking database permissions


smallsheeeep:
### What would you like to be added?

Provide internationalized error messages when lacking database permissions

### Why is this needed?

Current error messages are poorly readable:
![image](https://github.com/oceanbase/odc/assets/84772536/bbc8aee1-0242-47f5-90e0-f9d866947b0d)

PR #1678, opened 2024-02-29
fix(database-permission): permission intecept failed when switch schema or call PL

### LINKED ISSUE #1659: [Bug]: Only has the export permission of the database,running a function containing select type can return results.
### ODC version

ODC 4.2.4

### OB version

-

### What happened?

image
Only has the export permission of the database,running a function containing select type can return results.

If the user does not have query permissions for the database, an error will be reported when running pl of the query class.

```
create table test(a int);

insert into test values(1),(2),(1),(3),(3);

create function `get_a_column_data` () returns text BEGIN
    DECLARE result TEXT;
    
    SELECT GROUP_CONCAT(a SEPARATOR ', ') INTO result FROM test;
    
    RETURN result;
    end;
```

_No response_

### Cloud

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
module-database permission

There are following bugs:
- When call function or other PL, then DB permission intecept failed;
- When switch schema by executing SQL in console, then DB permission intecept failed;
- The error message is poorly readable.

This PR fix above issues by:
1. Add DB permission check logic in `DBPLService`;
2. Add new parser token solution for extract schema name in `SchemaExtractor`;
3. Update current schema of connection session when executing switch schema SQL in console. this is done by using `SwitchDatabaseInterceptor`;
4. Define a new error code for DB permission.

Fixes #1659 #1675 

Self-test passed.

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/SchemaExtractor.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/SchemaExtractor.java
@@ -22,6 +22,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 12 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/SchemaExtractor.java
```diff
@@ -203,6 +226,21 @@ public RelationFactor visitDatabase_factor(Database_factorContext ctx) {
         public RelationFactor visitCreate_database_stmt(Create_database_stmtContext ctx) {
             return null;
         }
+
+        @Override
+        public RelationFactor visitSimple_func_expr(Simple_func_exprContext ctx) {
```
yhilmare, line 231: this ast node is not correct, there are much more other node that can be referenced to function accessing~

yhilmare: we'd better visit bit_expr node and use MySQLExpressionFactory to judge if this node is a function calling~

server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/SchemaExtractor.java
```diff
@@ -60,6 +75,30 @@
...
+        Optional<String> schemaName = Optional.empty();
+        for (SqlTuple sqlTuple : sqlTuples) {
+            try {
+                AbstractSyntaxTree ast = sqlTuple.getAst();
+                if (ast == null) {
+                    sqlTuple.initAst(AbstractSyntaxTreeFactories.getAstFactory(dialectType, 0));
+                    ast = sqlTuple.getAst();
+                }
+                BasicResult basicResult = ast.getParseResult();
+                if (((dialectType.isMysql() || dialectType.isDoris()) && basicResult.getSqlType() == SqlType.USE_DB)
```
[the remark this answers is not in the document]
smallsheeeep: The original purpose is used for filterring. I find there is a bug when using Oracle and I have fixed it.

server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/SchemaExtractor.java
smallsheeeep: Done!

PR #1690, opened 2024-03-01
fix(dlm):job scheduler not found

guowl3:
#### What type of PR is this?
type-bug

The bean should only be obtained if the task framework is enabled.

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #1702, opened 2024-03-01
fix(database-change): create database change task failed in private sapce

### LINKED ISSUE #1693: [Bug]: Personal spatial database change execution failed
### ODC version

odc424

### OB version

*

### What happened?

image

Execution succeeded

Personal space execution database changes

_No response_

### Cloud

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug

This PR(#1578) cause a bug that cannot create a database change task in private space. The root cause is that flow instance created in private space has no PreCheck task. This PR just fix it.

Fixes #1693 

Self-test passed.

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
@@ -732,6 +732,10 @@ public class FlowInstanceService {
             ConnectionConfig connectionConfig) {
         log.info("Start creating flow instance, flowInstanceReq={}", flowInstanceReq);
         TaskType taskType = flowInstanceReq.getTaskType();
+        if (taskType == TaskType.ASYNC) {
+            DatabaseChangeParameters taskParameters = (DatabaseChangeParameters) flowInstanceReq.getParameters();
+            taskParameters.setModifyTimeoutIfTimeConsumingSqlExists(false);
+        }
         TaskEntity taskEntity = taskService.create(flowInstanceReq, (int) TimeUnit.SECONDS
                 .convert(flowTaskProperties.getDefaultExecutionExpirationIntervalHours(), TimeUnit.HOURS));
         Verify.notNull(taskEntity.getId(), "TaskId can not be null");


server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/DatabaseChangeRuntimeFlowableTask.java
```diff
@@ -208,20 +210,27 @@ private void modifyTimeoutIfTimeConsumingSqlExists(DelegateExecution execution,
...
-        if (Objects.nonNull(preCheckResult.getSqlCheckResult())
-                && preCheckResult.getSqlCheckResult().isTimeConsumingSqlExists()
-                && autoModifiedTimeout > parameters.getTimeoutMillis()) {
-            this.autoModifyTimeout = true;
-            parameters.setTimeoutMillis(autoModifiedTimeout);
-            TaskEntity databaseChangeTaskEntity = taskService.detail(taskId);
-            databaseChangeTaskEntity.setParametersJson(JsonUtils.toJson(parameters));
-            taskService.updateParametersJson(databaseChangeTaskEntity);
+        Optional<Long> preCheckTaskId;
+        try {
```
PeachThinking, line 214: we can set this parameter to true when create a database change task in personal space instead of by try catch: `parameters.isModifyTimeoutIfTimeConsumingSqlExists()`

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
```diff
@@ -732,6 +732,10 @@ private FlowInstanceDetailResp buildWithoutApprovalNode(CreateFlowInstanceReq fl
             ConnectionConfig connectionConfig) {
         log.info("Start creating flow instance, flowInstanceReq={}", flowInstanceReq);
         TaskType taskType = flowInstanceReq.getTaskType();
+        if (taskType == TaskType.ASYNC) {
```
yizhouxw, line 735: it also affect database change task generated by sql plan task ?

smallsheeeep: Yes! All task generated by this method has no PRE_CHECK task, so need add this `if` block.

smallsheeeep:
BTW: The "modify timeout if time consuming SQL exists" is completed by using a built-in SQL check rule and will not take affect in Individual Space. This is not match user's needs and will be fixed in later PR by @PeachThinking

yhilmare (approved):
lgtm

PR #1733, opened 2024-03-04
fix(db-browser):failed to get table when column has default value in oracle11g mode

### LINKED ISSUE #1727: [Bug]: After modifying the precision of the NUMBER type, the ORACLE data source fails to open.
### ODC version

ODC424

### OB version

independent

### What happened?

After modifying the precision of the NUMBER type, the ORACLE data source fails to open.
image

。

`
CREATE TABLE HX.NEWTABLE_10 (
	COLUMN1 NUMBER(38,125) DEFAULT NULL NOT NULL,
	CONSTRAINT SYS_C0017388 CHECK ("COLUMN1" IS NOT NULL)
);`
change into NUMBER(38,15)

_No response_

### Cloud

_No response_

PeachThinking:
#### What type of PR is this?
type-bug

All LONG or LONG RAW columns have to be retrieved from the ResultSet prior to all the other columns or oracle jdbc will throw “Stream has already been closed” Exception.

Fixes #1727

```docs

```


yhilmare (approved):
lgtm

PR #1748, opened 2024-03-05
fix(partition-plan): npe will be thrown when input is invalid

yhilmare:
#### What type of PR is this?

type-bug

avoid npe when input sql expression is invalid. as you can see, if we input an invalid sql expression, the execution result of this expr is null:
截屏2024-03-05 12 24 09

Fixes #

```docs

```


yizhouxw (approved):
lgtm

LuckyPickleZZ (approved):
LGTM

PR #1750, opened 2024-03-05
feat(partition-plan): disable the previous flow instance when a partition plan is related to an exists flow

yhilmare:
#### What type of PR is this?

type-feat

disable the previous flow instance when a partition plan is related to an exists flow

Fixes #

```docs

```


guowl3 (approved):
LGTM

PR #1768, opened 2024-03-06
fix(taskframework): fix task update final status when completed

krihy:
#### What type of PR is this?

type-bug
module-taskframework
modify `doStart` return boolean to update task status.

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #1778, opened 2024-03-06
fix(dlm): interrupt job failed.

guowl3:
#### What type of PR is this?
type-bug

After using the task framework, we need to use the task framework interface to terminate the task.

Fixes #

```docs

```


yhilmare (approved):
lgtm

Issue #1746, opened 2024-03-05
[Bug]: Export the stored procedure structure to a table, DELIMITER $$ does not have line breaks


Huangxiao-mas:
### ODC version

ODC424

### OB version

independent

### What happened?

Export the stored procedure structure to a table, DELIMITER $$ does not have line breaks
![image](https://github.com/oceanbase/odc/assets/59468957/c304eb19-180e-445d-bde2-c54a821602bc)

### What did you expect to happen?

`
DROP PROCEDURE GENERATE_TARCHIVES_BULK_DATA;
DELIMITER $$

CREATE OR REPLACE PROCEDURE "GENERATE_TARCHIVES_BULK_DATA" (p_record_count IN NUMBER) IS
BEGIN
  FOR i IN 1..p_record_count LOOP
    INSERT INTO TC_TARCHIVES (
      VC_ARCHIVES_NO,
      VC_REQUEST_NO,
      C_ARCHIVES_TYPE,
      B_ARCHIVES_CONTENT,
      VC_OP_TIME,
      VC_OPERATOR_NO,
      C_AUDIT_FLAG,
      VC_AUDIT_NO,
      C_ARCHIVES_STATE,
      VC_CUST_NO,
      L_SERIAL_NO,
      VC_DESCRIBE,
      VC_STORAGE_FILE_NAME,
      VC_TRADE_ACCO,
      VC_MANAGER_CODE,
      VC_REJECT_REASON,
      VC_REMARK,
      VC_PRODUCT_CODE,
      VC_STORAGE_FILE_PATH,
      VC_STORAGE_FILE_NO,
      VC_BATCH_NO
    ) VALUES (
      'ARCHIVES_' || LPAD(TO_CHAR(i), 10, '0'), -- 示例档案编号
      'REQ_' || LPAD(TO_CHAR(i), 10, '0'), -- 示例请求编号
      'TY', -- 档案类型示例值
      EMPTY_BLOB(), -- BLOB字段的模拟数据，这里是空BLOB
      TO_CHAR(SYSDATE, 'YYYYMMDDHH24MISS'), -- 操作时间
      'OPER_' || LPAD(TO_CHAR(i), 5, '0'), -- 操作员编号
      'Y', -- 审计标志示例值
      'AUD_' || LPAD(TO_CHAR(i), 5, '0'), -- 审计编号
      'A', -- 档案状态示例值
      LPAD(TO_CHAR(i), 12, '0'), -- 客户编号
[...]

PR #1786, opened 2024-03-06
fix(data-transfer): export Oracle object ddl without line breaks

### LINKED ISSUE #1671: [Bug]: precision is lost when displaying timestamps
### ODC version

4.2.4

### OB version

not related

### What happened?

precision is lost when displaying timestamps

precision should not be lost

1. create a table and insert table:
```sql
create table ts_tbl(
    col_ts_3 timestamp(3),
    col_ts_6 timestamp(6)
);
insert into ts_tbl values('2022-12-12 12:12:12.0','2022-12-12 12:12:12.0');
```
2. query this data:
截屏2024-02-29 16 43 07

we will see more precision if we query this data by obclient:
截屏2024-02-29 16 44 11

_No response_

### Cloud

_No response_

### LINKED ISSUE #1742: [Bug]: When exporting synonyms, after checking CREATE and adding DROP before it, the content of the exported file has a syntax error.
### ODC version

ODC424

### OB version

independent

### What happened?

When exporting synonyms, after checking CREATE and adding DROP before it, the content of the exported file has a syntax error.
image

correct
`DROP PUBLIC SYNONYM EXISTING_SYNONYM;

CREATE OR REPLACE PUBLIC SYNONYM SMTPTransport FOR schema.SMTPTransport;`

。

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

Fixes #1742  #1746  . Export Oracle object ddl without line breaks.
Fixes #1671  . When start a export task ,the front-end don't send `stopWhenError` param in request, so it would use false as it's value. If a object failed, the other would continue transfering.

### DIFF
--- a/server/plugins/task-plugin-api/src/main/java/com/oceanbase/odc/plugin/task/api/datatransfer/model/DataTransferConfig.java
+++ b/server/plugins/task-plugin-api/src/main/java/com/oceanbase/odc/plugin/task/api/datatransfer/model/DataTransferConfig.java
@@ -63,7 +63,7 @@ public class DataTransferConfig implements TaskParameters, Serializable {
     private EncodingType encoding = EncodingType.UTF_8;
     private CsvConfig csvConfig;
     private List<CsvColumnMapping> csvColumnMappings;
-    private boolean stopWhenError;
+    private boolean stopWhenError = true;
     private String exportFilePath;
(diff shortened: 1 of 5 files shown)


yhilmare (approved):
lgtm

PR #1797, opened 2024-03-07
fix(datasource): connecting failed in OceanBase Oracle after correcting the username

### LINKED ISSUE #1626: [Bug]: Modify the sys user of the data source in oracle mode, open the link after saving, and the data source is empty.
### ODC version

ODC424

### OB version

independent

### What happened?

Modify the sys user of the data source in oracle mode, open the link after saving, and the data source is empty.
image
image

modify successfully

1. First create the wrong data source information. For example, the correct user is sys. When creating, write it as sy and save it directly.
2. Edit the wrong data source link, change the wrong user name to sys, and click Test Connection. After passing, save and submit.
3. Enter the data source and find that there is no information

_No response_

### Cloud

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug

`ConnectionConfig.defaultSchema` is deprecated at the API level, so this PR adds `@JsonIgnore` on this field to prevent the issue which is that when users update the data source username, the front-end will pass the `defaultSchema` which is obtained from the `detail` API, and this will cause some logic error in the back-end: the odc server will not update the default schema using the username, it will use the old default schema passed by the front-end.

Fixes #1626 

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/ConnectionConfig.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/ConnectionConfig.java
@@ -222,6 +222,7 @@ public class ConnectionConfig
      * 默认 schema，对应 /api/v1 的 defaultDBName 字段
      */
     @Size(max = 128, message = "Schema name is out of range [0, 128]")
+    @JsonIgnore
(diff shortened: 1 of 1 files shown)


yhilmare (approved):
lgtm

PR #1800, opened 2024-03-08
fix(environment): flow instance selects wrong approval flow config after updating risk detect rules

### LINKED ISSUE #1640: [Bug]: The dlm task error message is incomplete and lacks detailed information.
### ODC version

ODC424

### OB version

independent

### What happened?

The dlm task error message is incomplete and lacks detailed information.
image

The dlm task error message is incomplete and lacks detailed information.

image

_No response_

### Cloud

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug

This PR deleted the i18n of the risk detect rule value. The front-end will directly map the i18n key to i18n value which is also passed by the back-end. Otherwise, when updating the environment, the front-end will pass the i18n value to the server rather than the i18n key.

Fixes #1640

```docs

```


yizhouxw (approved):
lgtm

PR #1804, opened 2024-03-08
fix(partition-plan): the first partition value is incorrect when creating partitions sequentially based on the current time

yhilmare:
#### What type of PR is this?

type-bug

I get a table like this:
```sql
create table datetime_range_parti_tbl(
    id int(11),
    create_time datetime
) partition by range columns (create_time) (
    partition p20240301 VALUES LESS than ('2024-03-01 00:00:00'),
    partition p20240401 VALUES LESS than ('2024-04-01 00:00:00')
)
```

and I create a partition plan with such configs:
截屏2024-03-18 10 34 22

I want to create my next 6 partitions and based on the time that the partition plan task executed. What I expected is I can get partitions from 2024-05-01 to 2024-10-01, but what I get is as follow:
截屏2024-03-08 15 22 46

as you can see: there is only 5 partitions, cause that the first generated partition is same with the last partition of this table which is not expected.

here is the current behavior:

截屏2024-03-18 10 35 29

Fixes #

```docs

```


smallsheeeep (approved):
LGTM

LuckyPickleZZ (dismissed):
lgtm

yizhouxw:
next 6 partitions，means 7 partitions，current + next7.

so the expected behaivior should be partition for month 4/5/6/7/8/9 be created,
which result into last partition p202409 less than '2024-10-01 00:00:00'

yizhouxw (dismissed):
lgtm

smallsheeeep (approved):

LuckyPickleZZ (approved):

smallsheeeep (approved):

PR #1808, opened 2024-03-08
fix(taskframework): fix invalid CSRF token when task process report heart to sever

krihy:
#### What type of PR is this?

type-bug
module-taskframework
add csrf white list, when uri match task white list, skip valid csrf token, because task api only be accessed by localhost when task is process model , fix task result heart request return InvalidCsrfToken like:
`[2024-03-08 08:31:52.966][http-nio-8989-exec-1][,,][INFO][com.oceanbase.odc.service.common.util.WebResponseUtils][150]: ODC_REQUEST_TRACE, httpMethod=POST, fullURL=http://localhost:8989/api/v2/task/heart, method=JSON, clientAddress=127.0.0.1, userId=Not-Required, result={"code":"InvalidCsrfToken","data":false,"durationMillis":0,"errCode":"InvalidCsrfToken","errMsg":"Invalid CSRF token","httpStatus":null,"importantMsg":false,"message":"Invalid CSRF token","requestId":null,"server":"ob-testC10-node4","timestamp":0.0,"traceId":""}, userAgent=Apache-HttpClient/4.5.13 (Java/1.8.0_402)
`

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/config/CommonSecurityProperties.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/config/CommonSecurityProperties.java
@@ -133,12 +133,7 @@ public class CommonSecurityProperties {
 
     public String[] getAuthWhitelist() {
         if (taskFrameworkEnabled) {
-            String[] copiedAuthWhitelist = new String[buildInAuthWhitelist.length + 3];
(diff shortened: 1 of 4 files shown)


yhilmare (approved):
lgtm

PR #1809, opened 2024-03-08
fix(connect): sometimes open sql console result in 404 error

yhilmare:
#### What type of PR is this?

type-bug

use url base64 encoder or decoder to encode or decode the session id to avoid '/' char.

Fixes #

```docs

```


yhilmare:
https://stackoverflow.com/questions/43089541/difference-between-basic-and-url-base64-encoding-in-java-8

yizhouxw (approved):
lgtm,

i change the title, for match bug sceanario, not solution

LuckyPickleZZ (approved):
lgtm

yhilmare:
ref: https://stackoverflow.com/questions/43089541/difference-between-basic-and-url-base64-encoding-in-java-8

Issue #1773, opened 2024-03-06
[Bug]: The schema file of the native ORACLE import table is imported into the ODC_TEST library, but it is imported into the SYS library.


Huangxiao-mas:
### ODC version

ODC424

### OB version

independent

### What happened?

The schema file of the native ORACLE import table is imported into the ODC_TEST library, but it is imported into the SYS library.
![image](https://github.com/oceanbase/odc/assets/59468957/0e8e9e69-a6ab-4ee6-81b8-5ca17f48f777)

### What did you expect to happen?

。

### How can we reproduce it (as minimally and precisely as possible)?

`CREATE TABLE "NON_PARTRANGE_HASH_NOPRI2" 
   (	"COL1" NUMBER(*,0), 
	"COL2" NUMBER, 
	"COL3" CHAR(10), 
	"COL4" VARCHAR2(10), 
	"COL5" BLOB, 
	"COL6" CLOB, 
	"COL7" DATE, 
	"COL8" TIMESTAMP (6), 
	"COL11" RAW(10)
   ) PCTFREE 10 PCTUSED 40 INITRANS 1 MAXTRANS 255 
  STORAGE(
  BUFFER_POOL DEFAULT FLASH_CACHE DEFAULT CELL_FLASH_CACHE DEFAULT)
  TABLESPACE "USERS" 
 LOB ("COL5") STORE AS BASICFILE (
  ENABLE STORAGE IN ROW CHUNK 8192 RETENTION 
  NOCACHE LOGGING 
  STORAGE(
  BUFFER_POOL DEFAULT FLASH_CACHE DEFAULT CELL_FLASH_CACHE DEFAULT)) 
 LOB ("COL6") STORE AS BASICFILE (
  ENABLE STORAGE IN ROW CHUNK 8192 RETENTION 
  NOCACHE LOGGING 
  STORAGE(
  BUFFER_POOL DEFAULT FLASH_CACHE DEFAULT CELL_FLASH_CACHE DEFAULT)) 
  PARTITION BY RANGE ("COL2") 
  SUBPARTITION BY HASH ("COL1") 
  SUBPARTITIONS 1
 (PARTITION "P1"  VALUES LESS THAN (0) 
PCTFREE 10 PCTUSED 40 INITRANS 1 MAXTRANS 255 
  STORAGE(
  BUFFER_POOL DEFAULT FLASH_CACHE DEFAULT CELL_FLASH_CACHE DEFAULT)
  TABLESPACE "USERS" 
[...]

PR #1832, opened 2024-03-11
fix(data-transfer): oracle mode import with incorrect splitted sqls

### LINKED ISSUE #1761: [Bug]: Native Oracle import function, after import, some fields in the function are missing
### ODC version

ODC424

### OB version

independent

### What happened?

Native Oracle import function, after import, some fields in the function are missing
image
image

。

[58774_export_file (1).zip](https://github.com/oceanbase/odc/files/14503644/58774_export_file.1.zip)

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

Fixed:
1. When import sql scripts for orarcle mode, we used `SqlCommentProcessor` to split sql. But it should use `SqlSplittter` instead. 
2. In MySQLDataTransferJob, schema jobs and data jobs will be initialized first before task transfer. Actually, this is incorrect because when creating a data job, table information will be queried. If the table has not been imported yet, an error will occur.

Fixes #1761 #1773


yhilmare (approved):
lgtm

PR #1838, opened 2024-03-11
fix(multi-cloud): failed to create new connection with read only account

PeachThinking:
#### What type of PR is this?
type-bug
module-multi-cloud

fix cannot create new connection with read only account

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #1843, opened 2024-03-11
fix(database-permission): return authorized permission types for detail database interface

smallsheeeep:
#### What type of PR is this?
type-feature
module-database permission management

This PR makes the detail database API returns authorized permission types information, so that the front-end can determine whether to display some databse operation button. This is required by the front-end.

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #1847, opened 2024-03-11
fix(schedule): no permission to edit

### LINKED ISSUE #1526: [Bug]: SQL plan cannot be edited, no permissions
### ODC version

ODC424

### OB version

independent

### What happened?

SQL plan cannot be edited, no permissions
image

can edict

edict sqlplan

_No response_

### Cloud

_No response_

guowl3:
#### What type of PR is this?
type-bug

When editing a task, an error of insufficient permissions will be reported.This is because when resourceRoleNames is empty, the verification method will determine that the operation has no permissions.

Fixes #1526

```docs

```


yhilmare (approved):
lgtm

PR #1851, opened 2024-03-12
enhancement(notification): support using variables in http header and body

LuckyPickleZZ:
#### What type of PR is this?
type-feature

The notification channel of ODC currently supports custom webhook channel, allowing users to input custom HTTP headers and body. But this template does not support variables, and some users require this feature. So we are increasing this support.
The specific method is to add event_id into the message entity for querying variables in event labels when sending messages. Using these variables to replacing the placeholders in the body and header.


smallsheeeep (dismissed):
LGTM for codes own by me

yhilmare (approved):
lgtm

MarkPotato777 (approved):
LGTM

PR #1867, opened 2024-03-12
fix(workflow): build front resource when only build client artifact

### LINKED ISSUE #1858: [Bug]: Using github to build client process needs optimization
### ODC version

-

### OB version

-

### What happened?

image
When building the client without checking jar, rpm, or docker, no front-end resources are built.

Also build front-end resources when only client is checked

Only check client

_No response_

### Cloud

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug

Fixes #1858 

```docs

```


yhilmare (approved):
lgtm

PR #1870, opened 2024-03-12
fix(metadb): fix executor meta db

krihy:
#### What type of PR is this?

type-bug
module-taskframework
1. fix meta db config environment name 
2. set meta db type set default is OB_MYSQL

Fixes #

```docs

```


guowl3 (approved):
LGTM

Issue #1871, opened 2024-03-13
[Bug]: Adding an auto-increment column does not match the offline detection


tonghan-gsh:
### ODC version

ODC4.2.4

### OB version

OB4.2

### What happened?

 Adding an auto-increment column does not match the offline detection
![image](https://github.com/oceanbase/odc/assets/46598871/017a3da3-f1aa-4ab8-a70b-b708aa8120e3)

### What did you expect to happen?

Adding an auto-increment column must match the offline detection

### How can we reproduce it (as minimally and precisely as possible)?

create table test_identity(a int);
ALTER TABLE test_identity ADD (id NUMBER GENERATED BY DEFAULT ON NULL AS IDENTITY);

sl01388797:
![Image](https://github.com/oceanbase/odc/assets/143492234/28b8e958-52bd-4873-ab04-34ad9eb60713)

pass

PR #1879, opened 2024-03-13
fix(sql-check): failed to detect pk dropping when drop it as constraint

### LINKED ISSUE #1875: [Bug]: Deleting the primary key through alter table drop constraint is not recognized as offline
### ODC version

ODC 4.2.4

### OB version

OBOracle 4.2

### What happened?

image
Deleting the primary key through alter table drop constraint is not recognized as offline

Delete the primary key identified as offline through alter table drop CONSTRAINT

create table test_identity(a int);
ALTER TABLE test_identity ADD CONSTRAINT constraint_name PRIMARY KEY (a);
ALTER TABLE test_identity DROP CONSTRAINT constraint_name;

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

primary key can be defined as a constraint, just like this:
```sql
create table abcd(
    id varchar2(64),
    constraint abcd primary key (id)
);
```
so that we can drop this pk by dropping constraint:
```sql
alter table abcd drop constraint abcd;
```

this pr consider of this situation~

Fixes #1875 #1871 #1891 #1893 #1906 

```docs

```


PeachThinking (dismissed):
LGTM

yizhouxw (approved):

PeachThinking (approved):

PR #1904, opened 2024-03-14
fix(taskframework):  fix retry log attribute

krihy:
#### What type of PR is this?

module-taskframework
type-bug
if job is retrying, log attribute will replicated, so update the new value is exists

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #1908, opened 2024-03-14
fix(iam): failed to batch import user with roles

### LINKED ISSUE #1900: [Bug]: NPE occurs when batch import user
### ODC version

4.2.4

### OB version

-

### What happened?

When batch import user from xlsx file and set roles, there will be an NPE . 
image

image

import should succeed

.

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

In `UserBatchImportPreviewer` , we used `Collections.emptyList()` incorrectly. It caused failures to import user.
Also optimized the error message of verification failure.

Fixes #1900

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/iam/UserBatchImportPreviewer.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/iam/UserBatchImportPreviewer.java
@@ -26,7 +26,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -43,7 +42,9 @@ import org.springframework.web.multipart.MultipartFile;
 import com.oceanbase.odc.common.util.StringUtils;
 import com.oceanbase.odc.core.authority.util.Authenticated;
 import com.oceanbase.odc.core.authority.util.PreAuthenticate;
+import com.oceanbase.odc.core.shared.Verify;
 import com.oceanbase.odc.core.shared.constant.ErrorCodes;
+import com.oceanbase.odc.core.shared.exception.VerifyException;
 import com.oceanbase.odc.service.common.util.FileConvertUtils;
 import com.oceanbase.odc.service.iam.auth.AuthenticationFacade;
 import com.oceanbase.odc.service.iam.model.BatchImportUser;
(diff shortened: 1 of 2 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/iam/UserBatchImportPreviewer.java
```diff
@@ -59,21 +60,12 @@
...
-    private static final Pattern REG_LOWERCASE = Pattern.compile(".*[a-z]+.*");
-    private static final Pattern REG_SYMBOL = Pattern.compile(".*[._+@#$%]+.*");
-    private static final Pattern SPACE_PATTERN = Pattern.compile("^(?=\\S).+(?<=[^.\\s])$");
-    public static final int PASSWORD_MAX_LENGTH = 32;
-    public static final int PASSWORD_MIN_LENGTH = 8;
-    public static final int NAME_MAX_LENGTH = 64;
-    public static final int ACCOUNTNAME_MAX_LENGTH = 64;
-    public static final int ACCOUNTNAME_MIN_LENGTH = 4;
+    private static final Pattern ACCOUNT_NAME_PATTERN = Pattern.compile("^[\\w.+@#$%]{4,64}$");
+    private static final Pattern PASSWORD_PATTERN = Pattern.compile(
```
smallsheeeep, line 64: Wow! How can you write so complex Regex expression? 👍🏻

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/UserBatchImportPreviewer.java
```diff
@@ -120,33 +118,33 @@ private boolean checkFileType(MultipartFile file) {
...
-        }
+        Verify.verify(SPACE_PATTERN.matcher(name).matches(),
+                "username length should not exceed 64 and not start or end with spaces");
+        batchImportUser.setName(name);
+
+        String accountName = map.get(USER_ACCOUNTNAME.getLocalizedMessage());
+        Verify.verify(!accountNames.contains(accountName), "account name already exists");
+        Verify.verify(ACCOUNT_NAME_PATTERN.matcher(accountName).matches(),
+                "account name must be 4 to 64 characters long "
+                        + "and consist of letters, numbers, and special characters in ._+@#$% only");
```
smallsheeeep, line 131: --> "Contains at least two lowercase letters, two uppercase letters, two numbers, and two special characters."

smallsheeeep (approved):
LGTM

yhilmare (approved):
lgtm

PR #1919, opened 2024-03-15
fix(stateful): max pool size may less than core pool size and cause error

ungreat:
#### What type of PR is this?
fix maxPoolSize may less than core

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #1923, opened 2024-03-15
fix(stateful): stateful interceptor will be npe when clientMode

ungreat:
#### What type of PR is this?
fix stateful interceptor will be npe when clientMode.
fix a multistate parameter passing error

Fixes #

```docs

```


server/odc-migrate/src/main/resources/migrate/web/V_4_2_4_16__add_stateful_route.sql
```diff
@@ -0,0 +1,21 @@
+/*
```
yhilmare, line 1: 42416 is occupied by this pr #1782

and why should we do this in a version script?

ungreat: Move script from data.sql to web version script, cause stateful should only be opened in web mode.

yhilmare (approved):
lgtm

PR #1938, opened 2024-03-18
fix(notification): NPE occurred when convert event to message

### LINKED ISSUE #1935: [Bug]: failed to send messge and NPE occurred
### ODC version

4.2.4

### OB version

-

### What happened?

image

send message successfully

-

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

When the channel is unbound from the policy , the relatiion will also be deleted.
So the `mappedRelations` could get a null, then NPE occurs.

Fixes #1935


MarkPotato777 (approved):
LGTM

PR #1939, opened 2024-03-18
fix(builtin-snippet): refine prefix for sys tenant, avoid duplicated name with mysql mode

yizhouxw:
#### What type of PR is this?

type-bug

refine prefix for sys tenant, avoid duplicated name with mysql mode.

all sys tenant query use`query_all` prefix:
image

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #1940, opened 2024-03-18
fix(ticket): cannot preview latest log and download complete log file

### LINKED ISSUE #1416: [Bug]: After the dlm log size exceeds the limit of 1M, the downloaded log is still 1M.
### ODC version

ODC423

### OB version

independent

### What happened?

After the dlm log size exceeds the limit of 1M, the downloaded log is still 1M.
image

Download all logs

chat privately

_No response_

### Cloud

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
module-ticket

In current version, user cannot preview latest log and download complete log file, beacuse the log will be truncated when out of limilt (10000 rows or 1MB).

This PR fix it by:
1. Use `ReversedLinesFileReader` to get the latest log;
2. Add a new API for downloading complete log file;
3. Set `fullLogDownloadUrl` for all downloadable task when the task is in final status. When OSS is enabled, then upload log file to OSS and return a OSS download URL. Otherwise, return a URL download from ODC Server.

Fixes #1416 

Self-test passed.

This PR does not deal with get latest log content and download log files when the task framework is enbaled. Please help complete these TODOs @krihy 

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowTaskInstanceService.java
```diff
@@ -192,33 +180,43 @@ public FlowInstanceDetailResp executeTask(@NotNull Long id) throws IOException {
...
-                        && instance.getTaskType() != TaskType.GENERATE_ROLLBACK);
+    public String getLog(@NotNull Long flowInstanceId, OdcTaskLogLevel level) throws IOException {
+        Optional<TaskEntity> taskEntityOptional = getLogDownloadableTaskEntity(flowInstanceId);
         if (!taskEntityOptional.isPresent()) {
             return null;
         }
         TaskEntity taskEntity = taskEntityOptional.get();
-        if (taskEntity.getResultJson() == null) {
-            return null;
+        if (!dispatchChecker.isTaskEntityOnThisMachine(taskEntity)) {
```
yhilmare, line 189: why should we check this first?
we may check if this task is impled by task framework, should it be better?

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/auth/OrganizationAuthenticationInterceptor.java
```diff
@@ -55,6 +55,7 @@ public class OrganizationAuthenticationInterceptor implements HandlerInterceptor
             "/api/v2/iam/users/me/organizations",
             "/api/v2/iam/login",
             "/api/v2/flow/flowInstances/*/tasks/download",
+            "/api/v2/flow/flowInstances/*/tasks/log/download",
```
yhilmare, line 58: why should we add a new api?

MarkPotato777 (dismissed):
LGTM

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowTaskInstanceService.java
smallsheeeep: Whether enabled task framework is not matter, it is just a different way for finding the log content. What we firstly need to do is determine whether the request is solved by the right meachine.

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/FlowInstanceController.java
```diff
@@ -176,6 +176,15 @@ public SuccessResponse<String> getLog(@PathVariable Long id, @RequestParam OdcTa
         return Responses.single(flowTaskInstanceService.getLog(id, logType));
     }
 
+    @ApiOperation(value = "downloadLog", notes = "下载任务完整日志")
+    @RequestMapping(value = "/{id:[\\d]+}/tasks/log/download", method = RequestMethod.GET)
+    public ResponseEntity<InputStreamResource> downloadLog(@PathVariable Long id) throws IOException {
```
guowl3, line 181: It may cause an OutOfMemoryError when the log file is too large, as the file is read into the JVM and then sent to the front-end in this manner.

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/auth/OrganizationAuthenticationInterceptor.java
smallsheeeep: Because I have not provide a API for the fron-end to download  full log file, and the existing APIs are not easily reusable. This issue (cannot download full log) exists since long time before.

guowl3 (dismissed):
LGTM

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/FlowInstanceController.java
smallsheeeep: We use `InputStreamResource ` and the inner inplementation is `FileInputStream`, so it will not occupied too much memory. I have test it by download three 1GB log files and the memory rate is only 2.8%:
image

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowTaskInstanceService.java
smallsheeeep: Yes you are right, done!

guowl3 (approved):
LGTM

krihy (dismissed):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/BaseODCFlowTaskDelegate.java
```diff
@@ -90,6 +112,24 @@ public abstract class BaseODCFlowTaskDelegate<T> extends BaseRuntimeFlowableDele
     private EventBuilder eventBuilder;
     @Autowired
     protected FlowTaskCallBackApprovalService flowTaskCallBackApprovalService;
+    @Autowired
+    protected CloudObjectStorageService cloudObjectStorageService;
+
+    private static final Set<TaskType> SUPPORT_DOWNLOAD_LOG_TASK_TYPES = new HashSet<TaskType>() {
```
yhilmare, line 118: this set may be useless, we can call `taskService.getLog` to get log content/file directly

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/BaseODCFlowTaskDelegate.java
```diff
@@ -310,4 +356,68 @@ public boolean cancel(boolean mayInterruptIfRunning) {
...
+                } catch (Exception e) {
+                    log.warn("Failed to upload task log file to OSS, taskId={}", taskId, e);
+                    throw new RuntimeException(
+                            String.format("Failed to upload task log file to OSS, file name: %s", fileName),
+                            e.getCause());
+                }
+            }
+            String resultJson = taskEntity.getResultJson();
+            AbstractFlowTaskResult result;
+            switch (taskType) {
```
yhilmare, line 381: we may use flowTaskInstance#getResult instead of this~

MarkPotato777 (dismissed):

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/BaseODCFlowTaskDelegate.java
smallsheeeep: Yes you are right, done!

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/BaseODCFlowTaskDelegate.java
smallsheeeep: Yes, done!

MarkPotato777 (approved):

yhilmare (approved):
lgtm

PR #1947, opened 2024-03-18
fix(flow): wrong total elements and total pages of flow instances while querying in page

MarkPotato777:
#### What type of PR is this?

type-bug

When querying the flow instances in pages via `flow_instance_view`, it will return the wrong `totalElements` and `pageSize` consequently.

The root cause is that Spring Data JPA gives a wrong SQL to count the total elements when the SQL has a **`GROUP BY`** statement. I assume it's a Spring's issue while they don't think so. Some other developers also encountered this problem. For more discussion, you can see at https://github.com/spring-projects/spring-data-jpa/issues/2361

Here is what happened in Spring Data JPA: When we use `GROUP BY` in our query sql, what we want to obtain is how many rows of the query sql will be returned by the query, but Spring just return the sum of each group's size. For example, let's say we have a table:
```
CREATE TABLE `test` (
  `a` bigint(20) DEFAULT NULL,
  `b` bigint(20) DEFAULT NULL
)
```

And we have such records in the table:
```
a b
1 1
1 2
2 3
```

If we create a query like `select * from test group by a` and we request for a `Page` result, then the Spring will use such sql to calculate the totalElements: `select count(a) from test group by a`, and sum all the result set as the `totalElements`. The result set of the above count sql will be like:
```
count(a)
2
1
```
And Spring will return 3(2 + 1) as the `totalElements` which is wrong. What we expect is 2, because the result set of the original query will be like this(given that we use a `GROUP BY` statement):
```
a b
1 1
2 3
```
The correct answer should be 2 because there are only 2 rows of the result set if we execute the original `GROUP BY` query.

The solution here is that if the query contains a `GROUP BY` statement, we just count distinct which will trickly address the issue. Thanks for the solution by https://github.com/spring-projects/spring-data-jpa/pull/2376

### DIFF
--- a/server/integration-test/src/test/java/com/oceanbase/odc/service/flow/FlowInstanceServiceTest.java
+++ b/server/integration-test/src/test/java/com/oceanbase/odc/service/flow/FlowInstanceServiceTest.java
@@ -46,6 +46,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.test.mock.mockito.MockBean;
 import org.springframework.data.domain.Page;
+import org.springframework.data.domain.PageRequest;
 import org.springframework.data.domain.Pageable;
 
 import com.oceanbase.odc.ServiceTestEnv;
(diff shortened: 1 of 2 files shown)


yizhouxw (approved):
lgtm

yhilmare (approved):
lgtm

PR #1948, opened 2024-03-18
fix(db-browser): the order of column names in Index and constraint are inconsistent

### LINKED ISSUE #1895: [Bug]: Display package body even when there is no package body in the native Oracle package
### ODC version

odc424

### OB version

*

### What happened?

image

image

Create a package with only a header in native Oracle

_No response_

### Cloud

_No response_

### LINKED ISSUE #1942: [Bug]: Index and constraint order are inconsistent
### ODC version

ODC424

### OB version

Oceanbase422

### What happened?

Index and constraint order are inconsistent
image
image

。

chat privately

_No response_

### Cloud

_No response_

PeachThinking:
#### What type of PR is this?
type-bug

1. fix the order of column names in Index and constraint are inconsistent
2. fix display package body even when there is no package body in the native Oracle package

Fixes #1942
Fixes #1895

```docs

```


yhilmare (approved):
lgtm

Issue #1943, opened 2024-03-18
[Bug]: Error when exporting result set in native oracle mode


tonghan-gsh:
### ODC version

ODC 4.2.4

### OB version

Oracle

### What happened?
![image](https://github.com/oceanbase/odc/assets/46598871/4594b929-7cd8-4128-969f-cba40d125ce6)
Error when exporting result set in native oracle mode

### What did you expect to happen?

Export result set without error

### How can we reproduce it (as minimally and precisely as possible)?

Export result set using native oracle

sl01388797:
![Image](https://github.com/oceanbase/odc/assets/143492234/b4ab989d-be9b-4772-8f3d-385044338521)

pass

PR #1956, opened 2024-03-19
fix(result-set): failed export oracle result set

### LINKED ISSUE #1931: [Bug]: Resultset export failed due to query timeout
### ODC version

4.2.4

### OB version

any

### What happened?

Resultset export failed due to query timeout when query SQL result set meta data by:
```java
SyncJdbcExecutor syncJdbcExecutor = session.getSyncJdbcExecutor(ConnectionSessionConstants.BACKEND_DS_KEY);
ResultSetMetaData rsMetaData =syncJdbcExecutor.query(parameter.getSql(), pss -> pss.setMaxRows(10),ResultSet::getMetaData);
```

Avoid timeout.

Export SQL result with long query time.

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

This PR fixes 2 issues:
1. We use ojdbc for oracle, and when a statement has been closed , the `ResultSetMetaData` could not be accessed any more. It has to be accessed in a statement.
2. If a query would be executed over default query timeout, the result set export task will fail. So we change this parameter before query.

Fixes #1931 #1943

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/resultset/ResultSetExportTask.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/resultset/ResultSetExportTask.java
@@ -19,7 +19,6 @@ import java.io.File;
 import java.io.IOException;
 import java.net.URL;
 import java.nio.file.Paths;
-import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -33,6 +32,7 @@ import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.io.FileUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.jdbc.core.StatementCallback;
 
 import com.oceanbase.odc.common.trace.TraceContextHolder;
 import com.oceanbase.odc.common.util.ExceptionUtils;
@@ -46,6 +46,7 @@ import com.oceanbase.odc.core.session.ConnectionSessionConstants;
 import com.oceanbase.odc.core.session.ConnectionSessionUtil;
 import com.oceanbase.odc.core.shared.PreConditions;
 import com.oceanbase.odc.core.shared.Verify;
+import com.oceanbase.odc.core.shared.constant.DialectType;
 import com.oceanbase.odc.core.shared.exception.OBException;
 import com.oceanbase.odc.core.shared.exception.UnexpectedException;
 import com.oceanbase.odc.core.shared.model.TableIdentity;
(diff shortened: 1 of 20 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/resultset/ResultSetExportTask.java
```diff
@@ -266,6 +276,21 @@ private void setColumnConfig(DataTransferConfig config, ResultSetExportTaskParam
         }
     }
 
+    private String getTimeoutSettingStatement(DialectType dialectType) {
```
yhilmare, line 279: you may use this for timeout setting
ConsoleTimeoutInitializer

server/odc-service/src/main/java/com/oceanbase/odc/service/resultset/ResultSetExportTask.java
```diff
@@ -266,6 +276,21 @@ private void setColumnConfig(DataTransferConfig config, ResultSetExportTaskParam
...
+        switch (dialectType) {
+            case OB_MYSQL:
+            case OB_ORACLE:
+            case ODP_SHARDING_OB_MYSQL:
+                return "set ob_query_timeout = " + DEFAULT_TIMEOUT_SECONDS * 1000000L;
```
yhilmare, line 284: we may not set this timeout settings by a static field, shall we add this parameter to the `ResultSetExportTaskParameter`?

we can give it a default value cause the front end does not set it~

LuckyPickleZZ: sure, I added a parameter in `ResultSetExportTaskParameter `

server/odc-service/src/main/java/com/oceanbase/odc/service/resultset/ResultSetExportTask.java
LuckyPickleZZ: thanks, updated.

server/odc-service/src/main/java/com/oceanbase/odc/service/resultset/ResultSetExportTaskParameter.java
```diff
@@ -49,6 +49,7 @@ public class ResultSetExportTaskParameter implements Serializable, TaskParameter
     private List<MaskingAlgorithm> rowDataMaskingAlgorithms;
     @JsonIgnore
     private String database;
+    private Integer executionTimeoutSeconds = 48 * 60 * 60;
```
smallsheeeep, line 52: 48h seems too long, we cannot accept that init a task needing 48 hours.

LuckyPickleZZ: Maybe we cannot avoid this. If user requests to export a query with `/*+ timeout */` hint, it would be executed for a long duration anyway.
And this timeout param does not take effect during task initialization, but rather during task execution. So it should not have a serious impact.

PeachThinking (approved):
LGTM

server/odc-service/src/main/java/com/oceanbase/odc/service/resultset/ResultSetExportTaskParameter.java
smallsheeeep: I got you!

PR #1961, opened 2024-03-19
fix(taskframework): fix cancel job and update executionTimes failed

krihy:
#### What type of PR is this?

type-bug
module-taskframework

1. if job status set to `CANCELING`, report result will be ignored
2.fix cancel request return `no content`
3.log query add default size when parameter is null
4.fix update execution_times failed

Fixes #

```docs

```


guowl3 (approved):
LGTM

yhilmare (approved):
lgtm

PR #1968, opened 2024-03-19
fix(parser): failed to parse insert statement with a table named 'json_table'

### LINKED ISSUE #1655: [Bug]: The instance stored procedure cannot be run by right-clicking and reports "Incorrect number of arguments"
### ODC version

ODC424

### OB version

Oceanbase421

### What happened?

The instance stored procedure cannot be run by right-clicking and reports "Incorrect number of arguments"
image
image

can execute it

```sql
CREATE TABLE json_table (
    id INT AUTO_INCREMENT PRIMARY KEY,
    data JSON
);

DELIMITER //

CREATE PROCEDURE InsertJsonData(IN num_rows INT)
BEGIN
    DECLARE i INT DEFAULT 1;
    DECLARE json_data JSON;
    WHILE i <= num_rows DO
        SET json_data = JSON_OBJECT(
            'name', CONCAT('Name', i),
            'age', FLOOR(RAND() * 30 + 20),
            'email', CONCAT('user', i, '@example.com')
        );
        INSERT INTO json_table (data) VALUES (json_data);
        SET i = i + 1;
    END WHILE;
END //

DELIMITER ;
```
Right click and run

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

"json_table" is a reserved key word and It should not be used as table name.

I checked the syntax file of ob and found that ob is allowed as a table name, so I modified it here to be consistent with OB.

Fixes #1655 

```docs

```

### DIFF
--- a/libs/ob-sql-parser/src/main/resources/obmysql/sql/OBParser.g4
+++ b/libs/ob-sql-parser/src/main/resources/obmysql/sql/OBParser.g4
@@ -4017,6 +4017,7 @@ unreserved_keyword_normal
     | JSON_VALUE
     | JSON_ARRAYAGG
     | JSON_OBJECTAGG
+    | JSON_TABLE
     | KEY_BLOCK_SIZE
     | KEY_VERSION
     | LAG


PeachThinking (approved):
LGTM

PR #1970, opened 2024-03-19
fix(sql-execute): the offset of the last sql goes wrong when it doesn't end with the delimiter

### LINKED ISSUE #1758: [Bug]: the location of a sql-check's violation is not correct
### ODC version

4.2.4

### OB version

not related

### What happened?

I input a statement and add some break lines before it
截屏2024-03-05 21 09 01

I click sql-check button, the location of these violations is not correct:
截屏2024-03-05 21 09 43

as you can see, the offset is zero:

截屏2024-03-05 21 10 15

location should be right

follow what happened

_No response_

### Cloud

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug

The issue is that when we execute sqls in sql console. If the last sql does NOT end with the delimiter, the allocating of violating sql-check rules is not accurate.

The root reason is that the splitter will add the last sql in the buffer rather than directly add it into spliited sqls. Then the caller wrongly calculates the offset of the last sql.

The solution is that if there is a buffer after splitting, we obtain the buffer's index in the original sql.

Fixes #1758 

```docs

```


yhilmare (approved):
lgtm

PR #1992, opened 2024-03-20
fix(partition-plan): failed to generate partition correctly when partition upper bound is not increased by 1 year/month/day

### LINKED ISSUE #1986: [Bug]: The first partition created by the partition plan should be larger than the existing partition or the partition start time
### ODC version

ODC 4.2.4

### OB version

Oracle430

### What happened?

The existing partitions are 20240401 and 20240701. The planned starting time for creating the partition is the current time and the time interval is 1 month. The created partitions are 20240501 and 20240601.
image
image
image

The created partition should be larger than the existing partition or the starting time. The first partition should be 20240801

CREATE TABLE oracle_range_date (
    id NUMBER,
    date_column DATE
)
PARTITION BY RANGE (date_column)
(
    PARTITION p1 VALUES LESS THAN (TO_DATE('2022-01-01', 'YYYY-MM-DD')),
    PARTITION p2 VALUES LESS THAN (TO_DATE('2022-04-01', 'YYYY-MM-DD')),
    PARTITION p3 VALUES LESS THAN (TO_DATE('2022-07-01', 'YYYY-MM-DD')),
    PARTITION p4 VALUES LESS THAN (TO_DATE('2022-10-01', 'YYYY-MM-DD')),
    PARTITION p5 VALUES LESS THAN (TO_DATE('2024-04-01', 'YYYY-MM-DD')),
    PARTITION p6 VALUES LESS THAN (TO_DATE('2024-07-01', 'YYYY-MM-DD'))
);

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

the partition plan can work normally when the partitioned table is like this:
```sql
CREATE TABLE datetime_range_parti_tbl (
  id integer DEFAULT NULL,
  parti_key_0 integer DEFAULT NULL,
  parti_key date DEFAULT NULL
) partition by range (parti_key)
(partition p202402 values less than (to_date('2024-02-01 00:00:00', 'YYYY-MM-DD HH24:MI:SS')),
partition p202403 values less than (to_date('2024-03-01 00:00:00', 'YYYY-MM-DD HH24:MI:SS')),
partition p202404 values less than (to_date('2024-04-01 00:00:00', 'YYYY-MM-DD HH24:MI:SS')));
```

as you can see, the partition is increased by 1 month, but the partition plan can not work when the partition is not increased by 1 month, eg. 3 month:

```sql
CREATE TABLE datetime_range_parti_tbl (
  id integer DEFAULT NULL,
  parti_key_0 integer DEFAULT NULL,
  parti_key date DEFAULT NULL
) partition by range (parti_key)
(partition p202402 values less than (to_date('2024-02-01 00:00:00', 'YYYY-MM-DD HH24:MI:SS')),
partition p202405 values less than (to_date('2024-05-01 00:00:00', 'YYYY-MM-DD HH24:MI:SS')),
partition p202408 values less than (to_date('2024-08-01 00:00:00', 'YYYY-MM-DD HH24:MI:SS')));
```

if we run partition plan at this table
before:
截屏2024-03-20 22 22 49

we expect that we will get "2024-11-01" next but we get "2024-09-01" actually cause that we are based on the wrong start time.

after:
截屏2024-03-20 22 31 43

we get the right partition~ why is there five partitions? cause that the "2024-08-01" has been existed~

you may noticed that the partition name is not expected and this is a design error and we can fix it at next version~ 

Fixes #1986 

```docs

```

### DIFF
--- a/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoLessThan5700SchemaAccessor.java
+++ b/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoLessThan5700SchemaAccessor.java
@@ -856,7 +856,15 @@ public class MySQLNoLessThan5700SchemaAccessor implements DBSchemaAccessor {
         return tableName2Res.entrySet().stream().collect(Collectors.toMap(Entry::getKey, e -> {
             Map<String, List<Map<String, Object>>> partiName2Res = e.getValue().stream()
                     .collect(Collectors.groupingBy(m -> (String) m.get("PARTITION_NAME")));
-            List<Map<String, Object>> group = partiName2Res.keySet().stream().map(partiName -> {
+            List<String> orderedPartiName = new ArrayList<>();
+            for (Map<String, Object> row : e.getValue()) {
+                String partiName = (String) row.get("PARTITION_NAME");
+                if (orderedPartiName.contains(partiName)) {
+                    continue;
+                }
+                orderedPartiName.add(partiName);
+            }
+            List<Map<String, Object>> group = orderedPartiName.stream().map(partiName -> {
                 List<Map<String, Object>> rows = partiName2Res.get(partiName);
                 Map<String, Object> row = rows.get(0);
                 row.put("SUB_NUM", rows.size() == 1 ? 0 : rows.size());
--- a/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoLessThan5700SchemaAccessor.java
+++ b/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoLessThan5700SchemaAccessor.java
@@ -855,16 +855,9 @@ public class MySQLNoLessThan5700SchemaAccessor implements DBSchemaAccessor {
                 Collectors.groupingBy(m -> (String) m.get("TABLE_NAME")));
         return tableName2Res.entrySet().stream().collect(Collectors.toMap(Entry::getKey, e -> {
             Map<String, List<Map<String, Object>>> partiName2Res = e.getValue().stream()
-                    .collect(Collectors.groupingBy(m -> (String) m.get("PARTITION_NAME")));
-            List<String> orderedPartiName = new ArrayList<>();
-            for (Map<String, Object> row : e.getValue()) {
-                String partiName = (String) row.get("PARTITION_NAME");
-                if (orderedPartiName.contains(partiName)) {
-                    continue;
-                }
-                orderedPartiName.add(partiName);
-            }
-            List<Map<String, Object>> group = orderedPartiName.stream().map(partiName -> {
+                    .collect(Collectors.groupingBy(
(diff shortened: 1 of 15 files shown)


libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoLessThan5700SchemaAccessor.java
```diff
@@ -856,7 +856,15 @@ public Map<String, DBTablePartition> listTablePartitions(@NonNull String schemaN
         return tableName2Res.entrySet().stream().collect(Collectors.toMap(Entry::getKey, e -> {
             Map<String, List<Map<String, Object>>> partiName2Res = e.getValue().stream()
                     .collect(Collectors.groupingBy(m -> (String) m.get("PARTITION_NAME")));
-            List<Map<String, Object>> group = partiName2Res.keySet().stream().map(partiName -> {
+            List<String> orderedPartiName = new ArrayList<>();
```
yizhouxw, line 859: may use linkedhashmap, for avoid possible performance risk while there exists too much partition for the table

server/integration-test/src/test/java/com/oceanbase/odc/service/partitionplan/PartitionPlanServiceTest.java
```diff
@@ -198,11 +198,11 @@ public void generatePartitionDdl_mysqlOnlyCreate_generateSucceed() throws Except
             Map<PartitionPlanStrategy, List<String>> expect = new HashMap<>();
             expect.put(PartitionPlanStrategy.CREATE, Collections.singletonList(String.format(
                     "ALTER TABLE %s.%s ADD PARTITION (\n"
-                            + "\tPARTITION `p20240125` VALUES LESS THAN (20220801,'2024-01-27'),\n"
-                            + "\tPARTITION `p20240125` VALUES LESS THAN (20220802,'2024-01-28'),\n"
-                            + "\tPARTITION `p20240125` VALUES LESS THAN (20220803,'2024-01-29'),\n"
-                            + "\tPARTITION `p20240125` VALUES LESS THAN (20220804,'2024-01-30'),\n"
-                            + "\tPARTITION `p20240125` VALUES LESS THAN (20220805,'2024-01-31'));\n",
+                            + "\tPARTITION `p20240125` VALUES LESS THAN (20220801,'2024-01-26'),\n"
```
smallsheeeep, line 201: Why should 27～31  change to 26~30?

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoLessThan5700SchemaAccessor.java
yhilmare: I use LinkedhashMap instead~

yizhouxw (approved):
lgtm

PeachThinking (approved):
LGTM

smallsheeeep (approved):
LGTM

LuckyPickleZZ (approved):

guowl3 (approved):
LGTM

PR #1996, opened 2024-03-21
fix(sql): precision is lost when displaying timestamps

### LINKED ISSUE #1671: [Bug]: precision is lost when displaying timestamps

zijiacj:
#### What type of PR is this?
type-bug

Fixes #1671

```docs

```

### DIFF
--- a/server/odc-core/src/main/java/com/oceanbase/odc/core/sql/execute/mapper/MySQLTimestampMapper.java
+++ b/server/odc-core/src/main/java/com/oceanbase/odc/core/sql/execute/mapper/MySQLTimestampMapper.java
@@ -42,10 +42,10 @@ public class MySQLTimestampMapper implements JdbcColumnMapper {
             return "0000-00-00 00:00:00";
         }
 
-        String returnValue = new String (data.getBytes()) ;
-       /* if (returnValue.endsWith(".0")) {
-            return returnValue.substring(0, returnValue.length() - 2);
-        }*/
+        String returnValue = new String(data.getBytes());
+        /*
+         * if (returnValue.endsWith(".0")) { return returnValue.substring(0, returnValue.length() - 2); }
+         */
         return returnValue;
     }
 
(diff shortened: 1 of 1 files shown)


server/odc-core/src/main/java/com/oceanbase/odc/core/sql/execute/mapper/MySQLTimestampMapper.java
```diff
@@ -41,10 +41,11 @@ public Object mapCell(@NonNull CellData data) throws SQLException {
         if (!isValidTimestamp(data)) {
             return "0000-00-00 00:00:00";
         }
-        String returnValue = timestamp.toString();
-        if (returnValue.endsWith(".0")) {
-            return returnValue.substring(0, returnValue.length() - 2);
-        }
+
+        String returnValue = new String(data.getBytes());
```
yhilmare, line 45: remove useless code, you may not put them in a comment~

and you can return it directly.

server/odc-core/src/main/java/com/oceanbase/odc/core/sql/execute/mapper/MySQLTimestampMapper.java
```diff
@@ -41,10 +41,11 @@ public Object mapCell(@NonNull CellData data) throws SQLException {
...
+        /*
```
yhilmare, line 46: btw, you may test this for two situation:
1. query data(you may have done this already)
2. try to modify a timestamp in table data viewing panel or result set viewing panel~

yhilmare (approved):
lgtm

yhilmare (approved):
lgtm

LuckyPickleZZ (approved):
lgtm

PR #2001, opened 2024-03-21
fix(notification): fix some security vulnerabilities

### LINKED ISSUE #2000: [Bug]: the sign key of the notification channel is not encrypted
### ODC version

ODC 4.2.4

### OB version

none

### What happened?

The sign key of the notification channel is not encrypted

it should be encrypted

create a channel with sign

_No response_

### Cloud

_No response_

LuckyPickleZZ:

1. The sign key of the notification channel is not encrypted. This PR fixed it, when front-end send `null` value as sign, meaning it's already in metadb, we won't update this key.
2. Now only project's owner could test channel.
3. Support `ticketUrl` as event label key

Fixes #2000

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/notification/helper/EventBuilder.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/notification/helper/EventBuilder.java
@@ -331,7 +331,11 @@ public class EventBuilder {
                 log.warn("failed to query project info.", e);
             }
         }
-        labels.putIfNonNull(TICKET_URL, getTicketUrl(labels, task));
+        try {
+            labels.putIfNonNull(TICKET_URL, getTicketUrl(labels, task));
+        } catch (Exception e) {
+            log.warn("failed to get ticket url.", e);
+        }
     }
 
     private <T> String getTicketUrl(EventLabels labels, T task) {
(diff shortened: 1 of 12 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/notification/helper/EventBuilder.java
```diff
@@ -321,6 +331,34 @@ private <T> void resolveLabels(EventLabels labels, T task) {
...
+        String taskType = labels.get(TASK_TYPE);
+        String taskId = labels.get(TASK_ID);
+        Long organizationId;
+        if (task instanceof TaskEntity) {
+            organizationId = ((TaskEntity) task).getOrganizationId();
+        } else {
+            organizationId = ((ScheduleEntity) task).getOrganizationId();
+        }
+
+        String host = Optional.ofNullable(hostProperties.getOdcHost()).orElse(SystemUtils.getLocalIpAddress());
```
yizhouxw, line 351: 1. lets' extract siteUrl cacluation logic to a separated component for reuse.

2. systemConfigService.queryByKeyPrefix, SystemConfigService may add a new method for queryByKey.

yizhouxw (approved):

server/odc-service/src/main/java/com/oceanbase/odc/service/notification/helper/EventBuilder.java
LuckyPickleZZ: I'll do these in the next PR~

MarkPotato777 (approved):
LGTM

yhilmare (approved):
lgtm

PR #2002, opened 2024-03-21
fix(taskframework): fix cancel result when status is done  

krihy:
#### What type of PR is this?

type-bug
module-taskframework

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/server/ThreadPoolTaskExecutor.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/server/ThreadPoolTaskExecutor.java
@@ -28,7 +28,6 @@ import java.util.concurrent.TimeoutException;
 import com.oceanbase.odc.common.concurrent.ExecutorUtils;
 import com.oceanbase.odc.core.task.TaskThreadFactory;
 import com.oceanbase.odc.service.task.caller.JobContext;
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/server/ThreadPoolTaskExecutor.java
```diff
@@ -66,28 +66,32 @@ public void execute(Task<?> task, JobContext jc) {
...
+            log.warn("Stop task time out, taskId={} .", ji.getId(), e);
+        }
+        if (!result) {
+            // if task is terminated, this method should return true,
+            // current status is CANCELING must push to CANCELED
+            result = getTask(ji).getStatus().isTerminated();
         }
-        return result || startFuture.cancel(true);
+        ExecutorUtils.gracefulShutdown(executor, "Task-Executor", result ? 1 : 5);
+        log.info("Task is canceled succeed, taskId={}, status={}.", ji.getId(), getTask(ji).getStatus());
```
yizhouxw, line 87: may add result into log output

krihy: okey, done

yhilmare (approved):
lgtm

PR #2007, opened 2024-03-21
fix(resultset-edit): bad performance when modifing query result set

### LINKED ISSUE #2005: [Bug]: Bad performance when batch generate UPDATE modifying SQL
### ODC version

4.2.4

### OB version

~

### What happened?

I preview table data, and update 1 column for about 100 rows, then generate modifying SQL cost about 16 seconds.

Complete above operation in 5 seconds.

Modify query result set, update 100+ rows.

_No response_

### Cloud

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
module-resultset

When I update 100 rows and 1 column in the result set, then ODC will cost about 16.5 second. The root cause is that the `UpdateGenerator` has repeated SQL query. This PR fix it and the time elapsed decrease to about 700ms.

Fixes #2005 

Self-test passed

```docs

```

### DIFF
--- a/server/integration-test/src/test/java/com/oceanbase/odc/service/dml/UpdateGeneratorTest.java
+++ b/server/integration-test/src/test/java/com/oceanbase/odc/service/dml/UpdateGeneratorTest.java
@@ -18,6 +18,10 @@ package com.oceanbase.odc.service.dml;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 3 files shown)


yizhouxw (approved):
👍👍

yhilmare (approved):
lgtm

LuckyPickleZZ (approved):

PeachThinking (approved):
LGTM

PR #2027, opened 2024-03-22
enhancement(connection): print exception stack when test connection failed

LuckyPickleZZ:
#### What type of PR is this?
type-enhancement

When user test connection failed, the exceptio would be eaten. Make some modifications here and print the complete exception stack.

### DIFF
--- a/server/plugins/connect-plugin-ob-mysql/src/main/java/com/oceanbase/odc/plugin/connect/obmysql/OBMySQLConnectionExtension.java
+++ b/server/plugins/connect-plugin-ob-mysql/src/main/java/com/oceanbase/odc/plugin/connect/obmysql/OBMySQLConnectionExtension.java
@@ -142,10 +142,11 @@ public class OBMySQLConnectionExtension implements ConnectionExtensionPoint {
         } catch (Exception e) {
             Throwable rootCause = ExceptionUtils.getRootCause(e);
             if (rootCause == null) {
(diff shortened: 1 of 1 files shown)


server/plugins/connect-plugin-ob-mysql/src/main/java/com/oceanbase/odc/plugin/connect/obmysql/OBMySQLConnectionExtension.java
```diff
@@ -142,10 +142,10 @@ protected TestResult internalTest(String jdbcUrl, Properties properties, int que
         } catch (Exception e) {
             Throwable rootCause = ExceptionUtils.getRootCause(e);
             if (rootCause == null) {
-                log.warn("Failed to get connection, errMsg={}", e.getLocalizedMessage());
+                log.warn("Failed to get connection, errMsg={}", e.getLocalizedMessage(), e);
```
yhilmare, line 145: there may be a lot of log printing

yizhouxw: may refer com.oceanbase.odc.common.util.ExceptionUtils.getRootCauseReason, which record 5 depth stack as root cause

yhilmare (approved):
lgtm

PR #2031, opened 2024-03-25
fix(sql-check): avoid npe when virtual column exists

### LINKED ISSUE #2028: [Bug]: Tables with virtual columns do not match sqlcheck detection
### ODC version

ODC 4.2.4

### OB version

OBOracle430

### What happened?

I created a partitioned table with virtual columns, but no sqlcheck detection was found.
image

image

```sql
CREATE TABLE ORACLE_RANGE_VARCHAR_2(
    id INT,
    char_column VARCHAR(64),
    char_column_number GENERATED ALWAYS AS (TO_NUMBER(char_column)) VIRTUAL
) PARTITION BY RANGE (char_column_number) (
    PARTITION p20240101 VALUES LESS THAN (20240101),
    PARTITION p20240201 VALUES LESS THAN (20240201),
    PARTITION p20240301 VALUES LESS THAN (20240301),
    PARTITION p20240401 VALUES LESS THAN (20240401)
);
```

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

the dataType is null when a column is a virtual column and it will lead to null pointer exception, this pr fix this problem.

Fixes #2028

```docs

```


PeachThinking (approved):
LGTM

PR #2033, opened 2024-03-25
fix(result-set-export): fix export result-set for oracle failed

### LINKED ISSUE #2017: [Bug]: Native Oracle export result set query result restriction cannot take effect
### ODC version

odc424

### OB version

*

### What happened?

image

Only two pieces of data

1. Create Table
2. Add 3 rows of data
3. Execute export result set query result restriction fill in 2

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

1. The `fetch first x rows only` syntax is only available on oracle after 12c. So we change it into `where rownum<=x`.
2. When export result-set, the `maxDumpSizeBytes` was not set. Fixed this.

Fixes #2017


yhilmare (approved):
lgtm

PeachThinking (approved):
LGTM

PR #2051, opened 2024-03-26
fix(task): try to upload log file even through the task has no log file

smallsheeeep:
#### What type of PR is this?
type-bug

This PR refine the code logic for uploading log files.

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/BaseODCFlowTaskDelegate.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/BaseODCFlowTaskDelegate.java
@@ -39,8 +39,8 @@ import com.oceanbase.odc.core.flow.exception.BaseFlowException;
 import com.oceanbase.odc.core.flow.model.AbstractFlowTaskResult;
 import com.oceanbase.odc.core.flow.model.FlowTaskResult;
 import com.oceanbase.odc.core.shared.Verify;
-import com.oceanbase.odc.core.shared.constant.TaskType;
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/BaseODCFlowTaskDelegate.java
```diff
@@ -336,6 +337,11 @@ public boolean cancel(boolean mayInterruptIfRunning) {
 
     private void setDownloadLogUrl(@NonNull Long flowInstanceId) throws IOException, NotFoundException {
         TaskEntity taskEntity = taskService.detail(taskId);
+        TaskType taskType = taskEntity.getTaskType();
```
yhilmare, line 340: we may use try-catch block to wrap this calling:
```java
File logFile = taskService.getLogFile(taskEntity.getCreatorId(), taskId + "", taskEntity.getTaskType(), OdcTaskLogLevel.ALL);
```

we may upload nothing when this calling throw any exceptions

smallsheeeep: The whole method `setDownloadLogUrl` is in a try-catch block.

smallsheeeep: Yes u are right, done!

yhilmare (approved):
lgtm

PR #2057, opened 2024-03-26
fix(sql-check): fix npe caused by offline statement detection

### LINKED ISSUE #2054: [Bug]: Delete SQL for primary key and perform SQL check with an error message
### ODC version

odc424

### OB version

*

### What happened?

image

No error reported

1. Create Table
CREATE TABLE "SHILIN"."A1" (
"A" INTEGER NULL  DEFAULT NULL,
"B" INTEGER NULL  DEFAULT NULL,
CONSTRAINT "F" PRIMARY KEY ("A")
) ;
2. Enter in SQL window   ALTER TABLE A1 DROP CONSTRAINT F;
3. Perform SQL check

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

there will be a npe, I avoid it~

Fixes #2054 

```docs

```


LuckyPickleZZ (approved):
lgtm

PeachThinking (approved):
LGTM

PR #2063, opened 2024-03-26
fix(db-browser): create table statement containing index is not recognized as a create type

### LINKED ISSUE #1568: [Bug]: The create table statement containing index is not recognized as a create type
### ODC version

ODC 4.2.3

### OB version

-

### What happened?

image
The create table statement containing index is not recognized as a create type,even if create is selected as the SQL type that is allowed to be executed, it will be intercepted during execution.

Executing a create table statement containing index is recognized as a create type.

1.Select create for the sql type allowed to be executed in the sql check specification.
2.
```sql
CREATE TABLE user_fav_app_0 (
    `id` BIGINT AUTO_INCREMENT PRIMARY KEY,
    `department_id` INT,
    `user_id` INT,
    `app_id` INT,
    `create_time` DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
    `cancel_time` DATETIME,
    `state` TINYINT DEFAULT 0 COMMENT '1代表收藏，0代表取消收藏',
    INDEX (`user_id`, `app_id`)
) COMMENT='记录每个用户的收藏应用';
```

_No response_

### Cloud

_No response_

zijiacj:
…te type

#### What type of PR is this?
type-bug
module-sql execution

Fixes #1568

```docs

```


yhilmare (approved):
lgtm

Issue #1998, opened 2024-03-21
[Bug]: When the OB version is 3. x, the execution of the code fragment query_user_list failed


sl01388797:
### ODC version

odc424

### OB version

*

### What happened?

![image](https://github.com/oceanbase/odc/assets/143492234/3561c14d-3aa9-474c-bcfd-cbd632d2bc85)

### What did you expect to happen?

*

### How can we reproduce it (as minimally and precisely as possible)?

NySQL code snippet search query_user_list

sl01388797:
![Image](https://github.com/oceanbase/odc/assets/143492234/93c2d673-7fdd-42d1-a1ea-8449ebfe639f)

pass

PR #2077, opened 2024-03-27
fix(builtin-snippets): duplicated prefix and wrong column for ob 3.x

### LINKED ISSUE #1997: [Bug]: MySQL has two revoke_all_privilages_from_user
### ODC version

odc424

### OB version

*

### What happened?

image

The names should be inconsistent

search for revoke_all_privilages_from_user

_No response_

### Cloud

_No response_

yizhouxw:
#### What type of PR is this?

type-bug

Fixes #1997 #1998 

```docs

```

### DIFF
--- a/server/odc-service/src/main/resources/builtin-snippet/mysql-snippet.yml
+++ b/server/odc-service/src/main/resources/builtin-snippet/mysql-snippet.yml
@@ -84,7 +84,7 @@
   body: |
     SELECT
       `user`,
-      `account_locked`,
+      `password_expired`,
       `select_priv`,
       `insert_priv`,
       `update_priv`,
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/resources/builtin-snippet/mysql-snippet.yml
```diff
@@ -84,7 +84,7 @@
   body: |
     SELECT
       `user`,
-      `account_locked`,
```
yizhouxw, line 87: account_locked invovled from mysql 5.7, however ob 3.x has no account_locked column.
here use password_expired instead, which is involved from mysql 5.6.

yhilmare (approved):
lgtm

PR #2078, opened 2024-03-27
enhancement: filter out data directory schema while checking the database permission

### LINKED ISSUE #2048: [Bug]: There is no need to restrict users’ permissions to access to schemas like  sys, information_schema
### ODC version

4.2.3

### OB version

all

### What happened?

If the sys schema is not added to the current project, you need to add it to the project to access it.
image

There is no need to restrict users’ permissions to access schema like sys、 information_schema and mysql.

like above

_No response_

### Cloud

_No response_

MarkPotato777:
#### What type of PR is this?

type-enhancement

When users execute SQLs, ODC will check if they have the associated database permissions. However, some data directory schema is not filtered out which will cause a bad user experience, as these data directory schemas are default granted to every database user.
Therefore, in this PR, I filter out the `SYS` in the Oracle dialect type and `information_schema` in the MySQL dialect type no matter if these two schemas are placed in the project.

Fixes #2048 

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/PreCheckRuntimeFlowableTask.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/PreCheckRuntimeFlowableTask.java
@@ -299,7 +299,8 @@ public class PreCheckRuntimeFlowableTask extends BaseODCFlowTaskDelegate<Void> {
                 }
             }
             unauthorizedDatabases =
-                    databaseService.filterUnauthorizedDatabases(schemaName2PermissionTypes, connectionConfig.getId());
+                    databaseService.filterUnauthorizedDatabases(schemaName2PermissionTypes, connectionConfig.getId(),
+                            true);
         }
         this.permissionCheckResult = new DatabasePermissionCheckResult(unauthorizedDatabases);
         this.sqlCheckResult = SqlCheckTaskResult.success(violations);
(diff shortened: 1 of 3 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/PreCheckRuntimeFlowableTask.java
```diff
@@ -299,7 +299,8 @@ private void doSqlCheckAndDatabasePermissionCheck(TaskEntity preCheckTaskEntity,
                 }
             }
             unauthorizedDatabases =
-                    databaseService.filterUnauthorizedDatabases(schemaName2PermissionTypes, connectionConfig.getId());
+                    databaseService.filterUnauthorizedDatabases(schemaName2PermissionTypes, connectionConfig.getId(),
```
yhilmare, line 302: There is also a class named PreCheckRuntimeFlowableTaskCopied. Does it also need to be modified?

MarkPotato777: Confirmed with @smallsheeeep , `PreCheckRuntimeFlowableTaskCopied` is not in use currently.

yizhouxw (approved):
lgtm

yhilmare (approved):
lgtm

PR #2081, opened 2024-03-27
fix (sql): Using AGGREGATE KEY in table building DDL resulted in failure to …

### LINKED ISSUE #2037: [Bug]: Using AGGREGATE KEY in table building DDL resulted in failure to view DDL
### ODC version

odc424

### OB version

*

### What happened?

image

No error will be reported

1. Create Table
```sql
CREATE TABLE sales (
    sales_id INT NOT NULL COMMENT "Unique sales identifier",
    product_id INT NOT NULL COMMENT "Identifier of the product sold",
    customer_id INT NOT NULL COMMENT "Identifier of the customer",
    sales_date DATEV2 NOT NULL COMMENT "Date of the sale",
    quantity BIGINT SUM COMMENT "Quantity sold",
    total_cost DECIMAL(9, 2) SUM COMMENT "Total cost of sales"
)
ENGINE=olap
AGGREGATE KEY (sales_id, product_id, customer_id, sales_date)
DISTRIBUTED BY HASH(sales_id) BUCKETS 10
PROPERTIES (
    "storage_type" = "COLUMN",
    "replication_num" = "1"
);
```
3. Detail

_No response_

### Cloud

_No response_

zijiacj:
…view DDL

#### What type of PR is this?
type-bug

Fixes #2037

```docs

```

### DIFF
--- a/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/model/DBTableColumn.java
+++ b/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/model/DBTableColumn.java
@@ -207,7 +207,8 @@ public class DBTableColumn implements DBObject, DBObjectWarningDescriptor {
          * If COLUMN_KEY is MUL, the column is the first column of a nonunique index in which multiple
          * occurrences of a given value are permitted within the column.
          */
-        MUL
+        MUL,
+        AGG
     }
 
     public enum CharUnit {


libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/model/DBTableColumn.java
```diff
@@ -207,7 +207,8 @@ public enum KeyType {
          * If COLUMN_KEY is MUL, the column is the first column of a nonunique index in which multiple
          * occurrences of a given value are permitted within the column.
          */
-        MUL
+        MUL,
+        AGG
```
PeachThinking, line 211: what dose AGG mean？Could you please add some comments like the other values above?

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/model/DBTableColumn.java
```diff
@@ -207,7 +207,12 @@ public enum KeyType {
          * If COLUMN_KEY is MUL, the column is the first column of a nonunique index in which multiple
          * occurrences of a given value are permitted within the column.
          */
-        MUL
+        MUL,
+
+        /**
+         * If COLUMN_KEY is AGG, the colum is an aggregate key in an Aggregate Model table.
```
PeachThinking, line 213: colum -> column

PeachThinking (approved):
LGTM

Issue #2053, opened 2024-03-26
[Bug]: failed to execute a sql with a single comment


yhilmare:
### ODC version

4.2.4

### OB version

not related

### What happened?

I write a sql with a single comment and an syntax exception will be thrown.

```sql
select
  -- this is a comment
1 from dual;
```
<img width="637" alt="截屏2024-03-26 14 10 16" src="https://github.com/oceanbase/odc/assets/22992296/6ddffb28-01eb-443a-a516-9f47a2c12ea8">

### What did you expect to happen?

execute it successfully~

### How can we reproduce it (as minimally and precisely as possible)?

execute the sql above, remember add two white space at the beginning of the single comment.

sl01388797:
![Image](https://github.com/oceanbase/odc/assets/143492234/2276780f-4c65-4b1f-bc4c-51314d9474d2)

pass

PR #2085, opened 2024-03-27
fix(sql-execute): cannot execute SQLs with single-line comments

### LINKED ISSUE #1758: [Bug]: the location of a sql-check's violation is not correct

MarkPotato777:
#### What type of PR is this?

type-bug

This PR addressed 2 issues:
- Cannot execute SQLs with single-line comments. The root cause is that it unexpectedly modified the Object of `OrderChar` when splitting SQLs with offset.
- The offset of SQLs is not accurate on **Windows** OS. The root cause is that while splitting SQLs line by line, it used `\r` as the regex pattern but on Windows OS, the line separator is `\r\n`.

Fixes #1758 #2053 

```docs

```


yhilmare (approved):
lgtm

PR #2087, opened 2024-03-27
fix(shadowtable): it occurs NPE while creating shadow table comparing tasks

### LINKED ISSUE #2039: [Bug]: Create shadow table synchronization failed
### ODC version

odc424

### OB version

*

### What happened?

image

Created successfully

Create shadow table synchronization

_No response_

### Cloud

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug

It occurs NPE because before the shadow table tasks have been bound to a flow instance, the `ShadowTableComparingTaskEntity#getId` will be null. If this is the case, we just need to compare the task's creatorId and currentUserId.

Fixes #2039 

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/structurecompare/StructureComparisonService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/structurecompare/StructureComparisonService.java
@@ -35,6 +35,7 @@ import com.oceanbase.odc.core.authority.util.SkipAuthorize;
 import com.oceanbase.odc.core.shared.constant.DialectType;
 import com.oceanbase.odc.core.shared.constant.ErrorCodes;
 import com.oceanbase.odc.core.shared.constant.ResourceType;
+import com.oceanbase.odc.core.shared.exception.AccessDeniedException;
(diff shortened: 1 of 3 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/structurecompare/StructureComparisonService.java
```diff
@@ -147,6 +148,12 @@ public List<StructureComparisonTaskResultEntity> batchCreateTaskResults(List<DBO
     }
 
     private void checkPermission(@NonNull StructureComparisonTaskEntity taskEntity) {
+        if (taskEntity.getFlowInstanceId() == null) {
```
PeachThinking, line 151: structure compare task is bind to flow instance id, the FlowInstanceId would not be null

MarkPotato777: okay~ I removed it.

yhilmare (approved):
lgtm

PeachThinking (approved):
LGTM

PR #2090, opened 2024-03-27
feat(tag): add tag service

ungreat:
#### What type of PR is this?
add tag service  and provide default implement

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/tag/TagService.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.oceanbase.odc.service.tag;
+
+public interface TagService {
```
yizhouxw, line 18: it is not a general design, may

```java
public interface AccountTags {
  List<Tag> queryTags(String acountName, List<String> labels);
}
```

server/odc-service/src/main/java/com/oceanbase/odc/service/config/SystemConfigService.java
```diff
@@ -57,6 +66,10 @@ public class SystemConfigService {
     @Autowired
     private List<SystemConfigRefreshMatcher> systemConfigRefreshMatchers;
 
+    private final LoadingCache<String, List<Configuration>> configCache = Caffeine.newBuilder().maximumSize(1)
```
yizhouxw, line 69: tag service should responsible for performance, so the cache may invovled in tag service

yizhouxw (commented):
tag is a basic infra, it may or may not usef for feature swith

server/odc-service/src/main/java/com/oceanbase/odc/service/config/SystemConfigService.java
ungreat: `configCache` works for the tag service as well.

server/odc-service/src/main/java/com/oceanbase/odc/service/tag/TagService.java
ungreat: Replace the original `TagService` with a `TagServiceFacade`. Now TagService` only focuses on providing tag service.

yizhouxw (approved):

yhilmare (approved):
lgtm

MarkPotato777 (approved):
LGTM

PR #2099, opened 2024-03-27
fix(partition-plan): failed to view partition plan tables when partition is not active

### LINKED ISSUE #2089: [Bug]: Partition plans requiring approval do not display partition strategy details
### ODC version

ODC4.2.4

### OB version

-

### What happened?

image
Partition plans requiring approval do not display partition strategy details.

Partition plans requiring approval need to display partition strategy details.

Add partition planning tasks to high-risk tasks and initiate tasks. Approvers view the details of tasks to be approved.

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

The partition plan configuration of a specific table cannot be viewed when the partition plan is not in effect. This is because the relevant records have been inserted into the corresponding table when the partition plan is started (usually in the antenna), so they cannot be found.

This PR fixes this problem. When the relevant records cannot be found, they are obtained from the parameters of the task.

Fixes #2089 

```docs

```


yizhouxw (approved):

smallsheeeep (approved):

smallsheeeep (approved):
LGTM

guowl3 (approved):
LGTM

PR #2104, opened 2024-03-28
fix(doris): failed to view active db sessions in doris

### LINKED ISSUE #2034: [Bug]: Doris, check session error
### ODC version

odc424

### OB version

*

### What happened?

image

No error reported

View sessions within Doris link

_No response_

### Cloud

_No response_

yhilmare:
#### What type of PR is this?

type-bug

doris doesn't have `information_schema.PROCESSLIST` and our session management will throw an exception when we attempt to query it. I use `show full processlist` instead:

截屏2024-03-28 12 09 17

there is some fields missing such as `status` and so on~

I have tried kill query and kill session and these are normal.

Fixes #2034 

```docs

```


PeachThinking (approved):
LGTM

MarkPotato777 (approved):
LGTM

yizhouxw (approved):

PR #2114, opened 2024-03-28
fix(data-masking): cannot masking result set when using Oracle

smallsheeeep:
#### What type of PR is this?
type-bug
module-data-masking

Fix cannot masking result set when using Oracle.

Fixes #

Self-test passed.

```docs

```


yhilmare (approved):
lgtm

PR #2119, opened 2024-03-29
fix(taskframework): handle all preparing and running task to failed when task-framework be set disabled

krihy:
#### What type of PR is this?

type-bug
module-taskframework
1. we should handle all running and failed jobs to failed when task-framework be set disabled

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/task/service/StdTaskFrameworkService.java
```diff
@@ -154,6 +154,14 @@ public Page<JobEntity> findHeartTimeTimeoutJobs(int timeoutSeconds, int page, in
         return page(condition, page, size);
     }
 
+    @Override
+    public Page<JobEntity> findNotTerminalJobs(int page, int size) {
+        Specification<JobEntity> condition = Specification.where(getRecentDaySpec(RECENT_DAY))
```
guowl3, line 159: Only querying the data of the past 30 days may miss published tasks, which may be optimized in the future.

guowl3 (approved):

server/odc-service/src/main/java/com/oceanbase/odc/service/task/service/StdTaskFrameworkService.java
krihy: ok, this wil be optimized by archive history data

yhilmare (dismissed):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/task/service/StdTaskFrameworkService.java
```diff
@@ -154,6 +154,14 @@ public Page<JobEntity> findHeartTimeTimeoutJobs(int timeoutSeconds, int page, in
         return page(condition, page, size);
     }
 
+    @Override
+    public Page<JobEntity> findNotTerminalJobs(int page, int size) {
```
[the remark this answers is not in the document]
krihy: okey,  looks more reasonable

guowl3 (approved):
LGTM

yhilmare (approved):
lgtm

MarkPotato777 (approved):
LGTM

PR #2129, opened 2024-04-01
fix(taskframework): fix decrypt meta db password

krihy:
#### What type of PR is this?

module-taskframework
type-bug
1.decrypt meta db password in pod
2.single fetch `PREPARING` job to 1
3.single fetch `RUNNING` job to 100

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/task/constants/JobConstants.java
```diff
@@ -59,5 +59,10 @@ public class JobConstants {
 
     public static final String ODC_JOB_MONITORING = "odcJobMonitoring";
 
+    public static final String ODC_PROPERTY_ENCRYPTION_ALGORITHM_NAME = "PBEWithHmacSHA512AndAES_128";
```
yhilmare, line 62: magic value

server/odc-service/src/main/java/com/oceanbase/odc/service/task/constants/JobConstants.java
```diff
@@ -59,5 +59,10 @@ public class JobConstants {
...
+
+    public static final String ODC_PROPERTY_ENCRYPTION_ALGORITHM_PREFIX = "ENC(";
```
yhilmare, line 64: odc has de/encryptor, why shouldn't we use them directly?

yhilmare (approved):
lgtm

PR #2146, opened 2024-04-02
fix(environment): wrong default environment description

MarkPotato777:
#### What type of PR is this?

type-bug

The former default environment's description is 'default environment which is not editable' while it is actually editable. This PR correct the description as just 'default environment'.

Fixes #

```docs

```


yizhouxw (approved):
lgtm

yhilmare (approved):
lgtm

PR #2147, opened 2024-04-02
enhancement(notification): add white list check for webhook to prevent SSRF

LuckyPickleZZ:
#### What type of PR is this?
type-enhancement

The  webhook url was not validated, this may cause SSRF attack.
So we added a system configuration `odc.notification.host-white-list` , only allow to create channel with webhook in this list.

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/notification/helper/ChannelConfigValidator.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/notification/helper/ChannelConfigValidator.java
@@ -15,15 +15,19 @@
  */
 package com.oceanbase.odc.service.notification.helper;
 
+import java.util.List;
 import java.util.Map;
 
+import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.collections4.MapUtils;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 import com.oceanbase.odc.common.json.JsonUtils;
 import com.oceanbase.odc.common.util.StringUtils;
 import com.oceanbase.odc.core.shared.Verify;
 import com.oceanbase.odc.core.shared.exception.NotImplementedException;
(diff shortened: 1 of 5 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/notification/helper/ChannelConfigValidator.java
```diff
@@ -103,4 +116,16 @@ private void validateWebhookChannelConfig(WebhookChannelConfig channelConfig) {
 
     }
 
+    private boolean checkHostInWhiteList(String webhook) {
```
MarkPotato777, line 119: ODC already existed an SSRF Checker, reference to `PreConditions#validInHostWhiteList`

LuckyPickleZZ: > seems we can use PreConditions.validInUrlWhiteList straightly. which is configured by system configuration `odc.integration.url-white-list straightly`

Should we not mix the white lists of these two modules?

yizhouxw: > > seems we can use PreConditions.validInUrlWhiteList straightly. which is configured by system configuration `odc.integration.url-white-list straightly`
> 
> Should we separate the white lists of these two modules?

I'm okay ,as it is security part, not business feature

LuckyPickleZZ: sure, I've fixed this.
Now we use the `SSRFChecker` straightly and the config `odc.integration.url-white-list `

yizhouxw (approved):

ungreat (approved):

yhilmare (approved):
lgtm

smallsheeeep (approved):
lgtm

PR #2158, opened 2024-04-07
fix(partition-plan): failed to migrate historical partition plan data

### LINKED ISSUE #2156: [Bug]: partition plan data migration may be failed
### ODC version

4.2.4

### OB version

not related

### What happened?

user do not set any partition plan or make all partition plan disable.
odc will startup failed.

odc can start up succeed~

follow what happened

_No response_

yhilmare:
#### What type of PR is this?

type-bug

odc will startup failed when all partition plan configs is disabled~ we may delete these data to work around this issue.

Fixes #2156 

```docs

```

### DIFF
--- a/server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/V42417HistoricalPartitionPlanMigrate.java
+++ b/server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/V42417HistoricalPartitionPlanMigrate.java
@@ -70,7 +70,9 @@ public class V42417HistoricalPartitionPlanMigrate implements JdbcMigratable {
     @Override
     public void migrate(DataSource dataSource) {
         JdbcTemplate jdbcTemplate = new JdbcTemplate(dataSource);
-        List<PartitionPlanEntityWrapper> partitionPlanWrappers = generatePartitionPlanEntities(jdbcTemplate);
+        List<PartitionPlanEntityWrapper> partitionPlanWrappers = generatePartitionPlanEntities(jdbcTemplate).stream()
+                .filter(w -> CollectionUtils.isNotEmpty(w.getPartitionPlanTableEntityWrappers()))
+                .collect(Collectors.toList());
         if (CollectionUtils.isEmpty(partitionPlanWrappers)) {
             log.info("No historical partition plan exists and the migration is complete");
             return;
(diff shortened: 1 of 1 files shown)


server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/V42417HistoricalPartitionPlanMigrate.java
```diff
@@ -70,7 +70,9 @@ public class V42417HistoricalPartitionPlanMigrate implements JdbcMigratable {
     @Override
     public void migrate(DataSource dataSource) {
         JdbcTemplate jdbcTemplate = new JdbcTemplate(dataSource);
-        List<PartitionPlanEntityWrapper> partitionPlanWrappers = generatePartitionPlanEntities(jdbcTemplate);
+        List<PartitionPlanEntityWrapper> partitionPlanWrappers = generatePartitionPlanEntities(jdbcTemplate).stream()
+                .filter(w -> CollectionUtils.isNotEmpty(w.getPartitionPlanTableEntityWrappers()))
+                .collect(Collectors.toList());
```
MarkPotato777, line 75: What's the impact of filtering out these disabled plans? Will these plans still exist in ODC 4.2.4?

yhilmare: > What's the impact of filtering out these disabled plans?

nothing, the way we cancel a partition plan is to startup a new partition plan and this new partition plan will contain no table or contain tables which is all disabled~

we will encounter a npe if we do not filter out these partition plans.

> Will these plans still exist in ODC 4.2.4

no, we will filter out these partition plan cause that these partition plans will do nothing and there is no necessary to migrate them~

MarkPotato777: So if users disabled a partition plan before V4.2.4, he/she cannot see that disabled plan any more after V4.2.4? Is that what we expected?

yhilmare: no, he/she will will can see these disabled partition plans but he/she will not see the detail of these partition plan cause that we do not migrate these data. ODC will return empty response when he/she attempt to view these data which is expected~

MarkPotato777: okay, that makes sense to me~

MarkPotato777 (approved):
LGTM

PR #2160, opened 2024-04-07
fix(web-framework): swagger-ui.html page 404 notfound

yizhouxw:
swagger-ui.html page 404 notfound

rootcause was springfox-wagger-ui dependency missed.


yhilmare (approved):
lgtm

PR #2166, opened 2024-04-07
feat(logicaldatabase): supports automatic recognition of logical tables and logical table expression generation 

MarkPotato777:
#### What type of PR is this?

type-feature

This PR implements automatic recognition of logical tables and generation of logical table expressions. The entry point is the `LogicalTableFinder#find` static method. Overall, the steps of the algorithm are as follows:
1. Based on the input` List<Database> databases`, find out all the table names under these databases. (Optimization has been made here, and the query will be grouped according to the `DataSource`, and the database will not be traversed directly to avoid excessive sessions)
2. Use `LogicalTableRecognitionUtils#recognizeLogicalTables` to detect all table names obtained in the first step, and return potential logical table candidates from the perspective of table name as input to the next step. We use an example to describe this algorithm. For example, the input is `"tb2_0, tb2_1, tb2_2, tb2_3, tb_0_t, tb_1_t, tb_2_t, tb_3_t"`.
   2.1 First replace the numbers in all table names with `[#]` to get a set of patterns. The patterns of all table names are consistently grouped together.
The output of this step is `"tb[#]_[#] ---> tb2_0, tb2_1, tb2_2, tb2_3"` and `"tb_[#]_t ---> tb_0_t, tb_1_t, tb_2_t, tb_3_t"`
   2.2 For each group of table names, since the fixed part of the table name may also contain numbers, we need to traverse the table names in the group. If the number at `[#]` is unchanged, replace `[#]` with the number.  This step can get the correct pattern, where the `[#]` part is the sequence number of the change.
The output of this step is `"tb2_[#] ---> tb2_0, tb2_1, tb2_2, tb2_3"` and `"tb_[#]_t ---> tb_0_t, tb_1_t, tb_2_t, tb_3_t"`
3. Traverse the `List<LigicalTable> candidates` obtained in the previous step, compare whether the structure of all physical tables under each LogicalTable is consistent, and move inconsistent physical tables out of the logical table. At this point, the logic table has gone through another round of screening. The way to determine whether the physical table structure is consistent is to obtain the complete `DBTable` object, and then calculate a sha1 for the attribute we are interested in in the `DBTable` as the unique identifier of the table structure. One detail here is that a physical table under `LogicalTbale` may have multiple table structures (although this is unreasonable, we need to take this abnormal scenario into account). At this time, we assume this logical table is composed of those most It consists of physical tables with the same structure. For example, there are 3 table structures of ddl1, 2 of ddl2, and 1 of ddl3. The three tables of ddl1 will jointly form this logical table, and the other tables will be removed.
4. Use the `LogicalTableRecognitionUtils#recognizeLogicalTablesWithExpression` method to identify the `List<LogicalTable>` obtained in the previous step that has been filtered according to the table structure again. This time, the identification will eliminate those logical tables with only one physical table and generate a logical table expression. . The specific algorithm steps are relatively complicated, and are briefly explained here:
- 4.1 First call `LogicalTableRecognitionUtils#recognizeLogicalTables` to obtain the logical table list
- 4.2 Classification discussion, depending on whether `databaseNamePattern` and `tableNamePattern` contain `[#]`, determine whether it is divided into databases only but not tables, only tables but not databases, or both databases and tables:
  - 4.2.1 Only divide the database and not the table. You only need to calculate the expression for the table part.
  - 4.2.2 Only divide the table and not the database. You only need to calculate the expression for the database part.
  - 4.2.3 Divide into databases and tables, continue the classification discussion
       - 4.2.3.1 If the physical table names under each physical library are exactly the same, they can be expressed in the form of `[[]]`. At this time, the expressions are calculated for `databaseNames` and `tableNames` respectively, and then the last pair of single brackets in the `tableNames` expression (`[]`) can be replaced with double brackets (`[[]]`)
       - 4.2.3.2 If the number of physical tables under each physical library is the same, it means that they can be divided equally. At this time, you need to judge whether it can be written into an expression. If it cannot be written into an expression, group it according to the database and calculate the values under each database separately. The expression of tableName is then joined with a `,`; if it can be written as an expression, calculate the expressions for `databaseNames` and `tableNames` respectively, and join them together. As for how to judge whether it can be written as an expression, it actually depends on the serial number of the table and whether it is assigned to each database in **"order"**. The algorithm here is actually the semantics of `order by` multiple columns in SQL, because the expression could be cascaded, such as `table_[#]_[#]`, see `LogicalTableRecognitionUtils#isOrderByEachElement`.
       - 4.2.3.3 In other cases, if it cannot be written as an expression, group it according to the database, calculate the expression of `tableName` under each database separately, and then join them with `,`.
5. The calculation expressions for "database" and "table" in step 4.2 above are implemented by the `LogicalTableRecognitionUtils#generateRangeOrList` method. This method has two overloaded methods, which are used to process a single number `(tb_[#]`) and two Scenarios of concatenated numbers (`tb_[#]_[#]`).
5.1 For the scenario of a single number: Classification discussion: whether it is an arithmetic sequence (that is, whether it can be expressed as a step-length scenario such as` [start-end:step]`), if so, whether the tolerance is 1 (that is, whether it can be expressed in the form `[start-end]`). All other scenarios are enumerated directly.
5.2 For the scenario of two levels of concatenated numbers: determine whether the two sets of numbers are a Cartesian product (`LogicalTableRecognitionUtils#isCartesianProduct`). If it is a Cartesian product, it can be expressed as `tb_[xxx]_[xxx]`. At this time, only You need to use a single number method for the two sets of numbers, obtain the expressions separately, and then splice them together. If it is not a Cartesian product, for the stability of the algorithm, the first group of numbers is first grouped, and the expression of the second group of numbers is calculated within the group. The calculation method is also processed according to the method of a single number.

The logic of the above algorithm is relatively complex and requires a large number of single tests to cover it. I divided the single test into two parts. LogicalTableRecognitionUtilsTest is a separate test for LogicalTableRecognitionUtils. This single test only focuses on the correctness of logical table recognition and expression generation. LogicalTableFinderTest will test the correctness after adding DDL judgment logic. In fact, LogicalTableFinder strongly relies on LogicalTableRecognitionUtils, but only adds DDL consistency detection logic. This part of the single test mainly covers the correctness of DDL consistency judgment.

The line coverage of a single test is 96%, and the remaining 4% cannot logically be constructed as a case but only exists for the robustness and completeness of the code. Although row coverage is not our primary concern, what we care about most is coverage of the core scenario. The positive examples I used in the single test have covered almost all legal expression types, and the counterexamples have selected several typical scenarios, such as inconsistent index column order, inconsistent column types, default values, and not null constraints, etc.

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/connection/logicaldatabase/LogicalTableFinder.java
```diff
@@ -0,0 +1,161 @@
...
+
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * @Author: Lebie
+ * @Date: 2024/3/22 17:20
+ * @Description: []
+ */
+@Slf4j
+@Validated
```
MarkPotato777, line 51: `@Valiedated` won't work

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/logicaldatabase/LogicalTableRecognitionUtils.java
```diff
@@ -0,0 +1,519 @@
...
+    private static final String DIGIT_REGEX_CAPTURING_GROUP_REPLACEMENT = "(\\\\d+)";
+    private static final String DOT_DELIMITER = ".";
+    private static final String COMMA_DELIMITER = ",";
+    private static final String LEFT_SQUARE_BRACKET = "[";
+    private static final String RIGHT_SQUARE_BRACKET = "]";
+    private static final String DOUBLE_LEFT_SQUARE_BRACKET = "[[";
+    private static final String DOUBLE_RIGHT_SQUARE_BRACKET = "]]";
+    private static final Pattern DIGIT_PATTERN = Pattern.compile("\\d+");
+
+    public static List<LogicalTable> recognizeLogicalTablesWithExpression(@NotEmpty List<DataNode> dataNodes) {
```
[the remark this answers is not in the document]
MarkPotato777: I've replaced with `Verify`~

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/logicaldatabase/model/DataNode.java
```diff
@@ -0,0 +1,155 @@
...
+
+    @JsonIgnore
+    private ConnectionConfig dataSourceConfig;
+
+    private String schemaName;
+
+    private String tableName;
+
+    @JsonIgnore
+    private DBTable table;
```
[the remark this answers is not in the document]
MarkPotato777: I've deleted the field `table`, and directly calculated the structure signature by passing the table as a parameter.

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/logicaldatabase/LogicalTableFinder.java
MarkPotato777: I've replaced with Verify~

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/logicaldatabase/LogicalTableFinder.java
```diff
@@ -0,0 +1,159 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * @Author: Lebie
+ * @Date: 2024/3/22 17:20
+ * @Description: []
+ */
+@Slf4j
+public class LogicalTableFinder {
+    public static List<LogicalTable> find(@NotEmpty List<Database> databases) {
```
yhilmare, line 51: `@NotEmpty` is still here~~

smallsheeeep (commented):
Great job 👏🏿     Several proposals.

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/logicaldatabase/LogicalTableFinder.java
```diff
@@ -0,0 +1,159 @@
...
+
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * @Author: Lebie
+ * @Date: 2024/3/22 17:20
+ * @Description: []
+ */
+@Slf4j
+public class LogicalTableFinder {
```
yizhouxw, line 48: oftenly, we use static method for simple utils, or something like a function.

this scenario, many outside resource is required, non-static method may easy for coding.
for example, you can involve some privite fields as calulate context.

for caller, juse `new LogicalTable(databases).find()`

yizhouxw, line 48: oftenly, we use static method for simple utils, or something like a function.

this scenario, many outside resource is required, non-static method may easy for coding.
for example, you can involve some privite fields as calulate context.

for caller, juse `new LogicalTable(databases).find()`

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/logicaldatabase/LogicalTableRecognitionUtils.java
```diff
@@ -0,0 +1,518 @@
...
+import com.oceanbase.odc.core.shared.exception.UnexpectedException;
+import com.oceanbase.odc.service.connection.logicaldatabase.model.DataNode;
+import com.oceanbase.odc.service.connection.logicaldatabase.model.LogicalTable;
+
+/**
+ * @Author: Lebie
+ * @Date: 2024/3/26 14:17
+ * @Description: []
+ */
+public class LogicalTableRecognitionUtils {
```
yizhouxw, line 49: if LogicalTableRecognitionUtils only called from LogicalTableFinder, may use package access privilege

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/logicaldatabase/LogicalTableFinder.java
```diff
@@ -0,0 +1,159 @@
...
+    public static List<LogicalTable> find(@NotEmpty List<Database> databases) {
+        Verify.notEmpty(databases, "LogicalTableFinder#find.databases");
```
yizhouxw, line 50: PreConditions for parameter validaton, Verify for internal state verification

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/logicaldatabase/LogicalTableFinder.java
MarkPotato777: Deleted~

server/integration-test/src/test/java/com/oceanbase/odc/service/connection/logicaldatabase/LogicalTableFinderTest.java
```diff
@@ -0,0 +1,133 @@
...
+ * @Date: 2024/4/7 16:06
+ * @Description: []
+ */
+public class LogicalTableFinderTest {
+    private static final String OBMYSQL_BASE_PATH = "src/test/resources/connection/obmysql/";
+    private static final JdbcTemplate OBMYSQL_JDBC_TEMPLATE =
+            new JdbcTemplate(TestDBConfigurations.getInstance().getTestOBMysqlConfiguration().getDataSource());
+
+
+    private static final LogicalTableFinder logicalTableFinder = new LogicalTableFinder();
```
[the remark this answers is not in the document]
MarkPotato777: Sure~

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/logicaldatabase/model/LogicalDatabase.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * @Author: Lebie
+ * @Date: 2024/3/22 11:35
+ * @Description: []
+ */
+@Data
+@Builder
+@NoArgsConstructor
+@AllArgsConstructor
+public class LogicalDatabase {
```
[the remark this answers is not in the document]
MarkPotato777: No, it's used in this PR, but will be used in the next PR. I added it for integrity~

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/logicaldatabase/LogicalTableFinder.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+
+        List<LogicalTable> finalLogicalTables =
+                LogicalTableRecognitionUtils.recognizeLogicalTablesWithExpression(logicalTableCandidates.stream()
+                        .map(LogicalTable::getActualDataNodes).flatMap(List::stream).collect(Collectors.toList()));
+
+        return finalLogicalTables;
+    }
+
+    private static Map<String, List<String>> getSchemaName2TableNames(ConnectionConfig dataSource,
```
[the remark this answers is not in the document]
MarkPotato777: I think `ConnectionSession` is bound to a `ConnectionConfig`(`DataSource`),  and I already group schemas and tables by data sources, so I assume it already made the most of `ConnectionSession`?

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/logicaldatabase/LogicalTableFinder.java
```diff
@@ -0,0 +1,159 @@
...
+        } finally {
+            try {
+                connectionSession.expire();
+            } catch (Exception ex) {
+                // eat exception
+            }
+        }
+    }
+
+    private static Map<String, DBTable> getTableName2Tables(ConnectionConfig dataSource, String schemaName,
```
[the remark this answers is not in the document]
MarkPotato777: Also see my responses to `getSchemaName2TableNames`

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/logicaldatabase/LogicalTableFinder.java
```diff
@@ -0,0 +1,159 @@
...
+                        .map(LogicalTable::getActualDataNodes).flatMap(List::stream).collect(Collectors.toList()));
+
+        return finalLogicalTables;
+    }
+
+    private static Map<String, List<String>> getSchemaName2TableNames(ConnectionConfig dataSource,
+            List<Database> groupedDatabases) {
+        ConnectionSessionFactory connectionSessionFactory = new DefaultConnectSessionFactory(dataSource);
+        ConnectionSession connectionSession = connectionSessionFactory.generateSession();
+        DBSchemaAccessor schemaAccessor = DBSchemaAccessors.create(connectionSession);
```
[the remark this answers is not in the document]
MarkPotato777: Sure~

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/logicaldatabase/LogicalTableFinder.java
MarkPotato777: Deleted~

yizhouxw (commented):
we may set a peformance taget,  for example:
how long and how many memory usage for  100 phisical databases scenario, with 1000 logic tables, each table with average 20 columns, 3 indexes, 3 constraints.

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/logicaldatabase/LogicalTableRecognitionUtils.java
MarkPotato777: Sure, I changed it to the default access privilege.

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/logicaldatabase/LogicalTableFinder.java
MarkPotato777: Okay, done~

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/logicaldatabase/LogicalTableFinder.java
MarkPotato777: Got it, updated~

MarkPotato777:
> we may set a peformance taget, for example: how long and how many memory usage for 100 phisical databases scenario, with 1000 logic tables, each table with average 20 columns, 3 indexes, 3 constraints.

Sure, I will provide a performance test report after I completed the main process~~

smallsheeeep (approved):
LGTM

yizhouxw (approved):
lgtm

PR #2169, opened 2024-04-08
fix(taskframework): dlm task show log not exists

krihy:
#### What type of PR is this?

type-bug
module-taskframework

1. throw exception when  destroy process failed
2. do final work before stop task 
3. get log by forward to remote odc server with try catch, avoid remote server is not reachable

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/BaseTask.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/BaseTask.java
@@ -78,13 +78,13 @@ public abstract class BaseTask<RESULT> implements Task<RESULT> {
                 doStop();
                 updateStatus(JobStatus.CANCELED);
             }
+            return true;
         } catch (Throwable e) {
             log.warn("Stop task failed, id={}", getJobId(), e);
             return false;
         } finally {
             close();
         }
-        return true;
     }
 
     @Override
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/BaseTask.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/BaseTask.java
@@ -116,7 +116,7 @@ public abstract class BaseTask<RESULT> implements Task<RESULT> {
             } catch (Throwable e) {
                 // do nothing
             }
-            log.info("Task be completed, id={}, status={}.", getJobId(), getStatus());
+            log.info("Task completed, id={}, status={}.", getJobId(), getStatus());
             taskMonitor.finalWork();
         }
     }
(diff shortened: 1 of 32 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/BaseTask.java
```diff
@@ -111,6 +109,18 @@ private void initCloudObjectStorageService() {
         storageConfig.ifPresent(osc -> this.cloudObjectStorageService = CloudObjectStorageServiceBuilder.build(osc));
     }
 
+    private void close() {
+        if (closed.compareAndSet(false, true)) {
```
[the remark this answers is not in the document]
krihy: ok, this only node can stop this task at one time

guowl3 (dismissed):
LGTM

server/odc-service/src/main/java/com/oceanbase/odc/service/task/schedule/daemon/CheckRunningJob.java
```diff
@@ -72,30 +72,16 @@ private void handleJobRetryingOrFailed(JobEntity jobEntity) {
...
+        if (isNeedRetry) {
             log.info("Need to restart job, destroy old executor completed, jobId={}.", a.getId());
             int rows = getConfiguration().getTaskFrameworkService()
-                    .updateStatusDescriptionByIdOldStatusAndExecutorDestroyed(a.getId(), JobStatus.RUNNING,
+                    .updateStatusDescriptionByIdOldStatus(a.getId(), JobStatus.RUNNING,
                             JobStatus.RETRYING, "Heart timeout and retrying job");
             if (rows > 0) {
                 log.info("Job {} set status to RETRYING.", a.getId());
+            } else {
+                throw new TaskRuntimeException("Set job status to FAILED failed, jobId=" + jobEntity.getId());
```
yizhouxw, line 84: FAILED --> RETRYING ?

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/BaseTask.java
```diff
@@ -111,6 +109,18 @@ private void initCloudObjectStorageService() {
...
+            try {
+                doClose();
+            } catch (Throwable e) {
+                // do nothing
+            }
```
yizhouxw, line 118: Task completed

server/odc-service/src/main/java/com/oceanbase/odc/service/task/schedule/daemon/CheckRunningJob.java
krihy: yes, that's right , i fix it

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/BaseTask.java
krihy: done

yizhouxw (approved):

PR #2174, opened 2024-04-08
fix(taskframework):  calculate capacity of task in  single node by node memory

krihy:
#### What type of PR is this?

 calculate capacity of task size in  single node by node memory

Fixes #

```docs

```

### DIFF
--- a/server/odc-server/src/main/resources/data.sql
+++ b/server/odc-server/src/main/resources/data.sql
@@ -753,8 +753,8 @@ INSERT INTO config_system_configuration ( `key`, `value`, `description` ) VALUES
 INSERT INTO config_system_configuration ( `key`, `value`, `description` ) VALUES( 'odc.task-framework.k8s-properties.limit-cpu', '2', 'k8s pod limit cpu' ) ON DUPLICATE KEY UPDATE `id` = `id`;
 INSERT INTO config_system_configuration ( `key`, `value`, `description` ) VALUES( 'odc.task-framework.k8s-properties.limit-mem', '4096', 'k8s pod limit memory,unit is MB' ) ON DUPLICATE KEY UPDATE `id` = `id`;
 INSERT INTO config_system_configuration ( `key`, `value`, `description` ) VALUES( 'odc.task-framework.k8s-properties.max-node-count', '', 'k8s max node count' ) ON DUPLICATE KEY UPDATE `id` = `id`;
-INSERT INTO config_system_configuration ( `key`, `value`, `description` ) VALUES( 'odc.task-framework.k8s-properties.node-cpu', '', 'single k8s node cpu count' ) ON DUPLICATE KEY UPDATE `id` = `id`;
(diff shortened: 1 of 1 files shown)


guowl3 (approved):
I think the best solution is to obtain the specifications of the expansion node through the API.
ok, i will discuss with other team

server/odc-server/src/main/resources/data.sql
```diff
@@ -753,6 +753,8 @@ INSERT INTO config_system_configuration ( `key`, `value`, `description` ) VALUES
 INSERT INTO config_system_configuration ( `key`, `value`, `description` ) VALUES( 'odc.task-framework.k8s-properties.limit-cpu', '2', 'k8s pod limit cpu' ) ON DUPLICATE KEY UPDATE `id` = `id`;
 INSERT INTO config_system_configuration ( `key`, `value`, `description` ) VALUES( 'odc.task-framework.k8s-properties.limit-mem', '4096', 'k8s pod limit memory,unit is MB' ) ON DUPLICATE KEY UPDATE `id` = `id`;
 INSERT INTO config_system_configuration ( `key`, `value`, `description` ) VALUES( 'odc.task-framework.k8s-properties.max-node-count', '', 'k8s max node count' ) ON DUPLICATE KEY UPDATE `id` = `id`;
+INSERT INTO config_system_configuration ( `key`, `value`, `description` ) VALUES( 'odc.task-framework.k8s-properties.node-cpu', '', 'k8s node cpu' ) ON DUPLICATE KEY UPDATE `id` = `id`;
```
yizhouxw, line 756: description missed

server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/PodConfig.java
```diff
@@ -59,4 +59,8 @@ public class PodConfig {
     private Long mountDiskSize;
 
     private Long maxNodeCount;
+
+    private Long nodeCpu;
```
yizhouxw, line 63: may double, for example cpu can be 0.5 core

server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/PodConfig.java
```diff
@@ -59,4 +59,8 @@ public class PodConfig {
...
+
+    private Long nodeMem;
```
yizhouxw, line 65: if hardcode unit, filed name may end with unit name.

better follow k8s configuration design.

server/odc-server/src/main/resources/data.sql
krihy: okey, done

server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/PodConfig.java
krihy: done

server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/PodConfig.java
krihy: okey, i rename to `nodeMemInMb` which follow other field

yizhouxw (approved):

PR #2179, opened 2024-04-09
fix(pl-debug): make timeout settings can be overwritten by session init script

### LINKED ISSUE #2172: [Bug]: failed to debug a pl that contains a sql costing more than 10 minutes
### ODC version

4.2.4

### OB version

not related

### What happened?

1. create a pl that contains a sql costing more that 10 mintues
2. try to debug this pl
3. debug will timeout at this sql

we can increase the timeout setting using session init script

follow 'What happened'

_No response_

yhilmare:
#### What type of PR is this?

type-bug

the pldebug's session timeout settings is hard code and can not be modified.
I run these timeout settings after session init script to avoid this issue~

Fixes #2172 

```docs

```


krihy (approved):
lgtm

yizhouxw (approved):

PR #2181, opened 2024-04-09
fix(data-transfer): obloader fail to load MANIFEST.bin

### LINKED ISSUE #2176: [Bug]: fail to import a zip exported by ODC
### ODC version

4.2.4

### OB version

-

### What happened?

Fail to import a zip exported by ODC.

-

1. Export some data in CSV format. 
2. Download zip file.
3. Import.

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

In the earlier version, ob-loader-dumper would save binary files like `MANIFEST.bin` in data directory. 
But in obloader-2.4.8.2, it has changed the file path into root directory. So ODC has adapted this feature.

FROM
```
root
└── data
    ├── TABLE
    │      └── xxx.sql
    └── MANIFEST.bin
```
TO
```
root
├── data
│     └── TABLE
│          └── xxx.sql
└── MANIFEST.bin
```

Fixes #2176

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/datatransfer/task/DataTransferTask.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/datatransfer/task/DataTransferTask.java
@@ -273,14 +273,14 @@ public class DataTransferTask implements Callable<DataTransferTaskResult> {
 
         if (dataPath.exists()) {
             boolean deleteRes = FileUtils.deleteQuietly(dataPath);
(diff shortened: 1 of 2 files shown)


yhilmare (approved):
could the object provided by earlier ob-loader-dumper be imported?

LuckyPickleZZ:
> could the object provided by earlier ob-loader-dumper be imported?

yes, ODC finds MANIFEST.bin by scanning directories. As long as ODC can find the manifest file, it can be output to the working directory of ob-loader

PR #2185, opened 2024-04-09
fix(notification): failure to enqueue event will result in failure of ticket rejection

### LINKED ISSUE #2183: [Bug]: fail to reject ticket because of notification exception
### ODC version

4.2.4

### OB version

-

### What happened?

When a datasource was configured without cluster, it would lead to failure of enqueue event.
And reject ticket would fail.

1. The event should be enqueued normally.
2. It should not affect original transaction.

1. Create a datasource without cluster. 
2. Configure notification policy.
3. Reject a ticket.

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

1. There maybe some null value of event labels, and these null would leads to failure of enqueue.
2. In `FlowInstanceService#reject`, although the enqueue exception has been successfully captured, the hibernate mechanism has caused the session to fail. And subsequent database operations cannot proceed, resulting in rejection failure.
So I set the propegation of `Broker#enqueueEvent` to REQUIRES_NEW. Start a new transaction to insert an event.

Fixes #2183

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/notification/Broker.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/notification/Broker.java
@@ -20,8 +20,6 @@ import java.util.Optional;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
-import org.springframework.transaction.annotation.Propagation;
-import org.springframework.transaction.annotation.Transactional;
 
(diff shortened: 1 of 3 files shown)


MarkPotato777 (dismissed):
LGTM

yhilmare (approved):
lgtm

yizhouxw (approved):
lgtm

MarkPotato777 (approved):

PR #2192, opened 2024-04-11
feat(connect): support connect backup instance

MarkPotato777:
#### What type of PR is this?

type-feature

If DataSource is an OB backup instance, then we automatically add 'set ob_read_consistency='WEAK';' when the session was created.

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionTesting.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionTesting.java
@@ -256,8 +256,8 @@ public class ConnectionTesting {
 
         List<ConnectionInitializer> initializers = new ArrayList<>();
         initializers.addAll(
-                Arrays.asList(new DataSourceInitScriptInitializer(config, false),
-                        new BackupInstanceInitializer(config)));
+                Arrays.asList(new BackupInstanceInitializer(config),
+                        new DataSourceInitScriptInitializer(config, false)));
         try (Connection connection = DriverManager.getConnection(jdbcUrl, properties);
                 Statement statement = connection.createStatement()) {
             if (queryTimeoutSeconds >= 0) {
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionTesting.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionTesting.java
@@ -19,6 +19,9 @@ import java.sql.Connection;
 import java.sql.DriverManager;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Objects;
 import java.util.Properties;
 
@@ -53,7 +56,8 @@ import com.oceanbase.odc.service.connection.ssl.ConnectionSSLAdaptor;
 import com.oceanbase.odc.service.connection.util.ConnectTypeUtil;
 import com.oceanbase.odc.service.plugin.ConnectionPluginUtil;
 import com.oceanbase.odc.service.session.factory.OBConsoleDataSourceFactory;
-import com.oceanbase.odc.service.session.initializer.SessionCreatedInitializer;
+import com.oceanbase.odc.service.session.initializer.BackupInstanceInitializer;
+import com.oceanbase.odc.service.session.initializer.DataSourceInitScriptInitializer;
 
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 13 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/ConnectionConfig.java
```diff
@@ -426,6 +429,14 @@ public int queryTimeoutSeconds() {
         return queryTimeoutSeconds;
     }
 
+    @JsonIgnore
```
yhilmare, line 432: It's too customized. This class also needs to be oriented to other data sources. This way it will be too closely bound to OB.

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/ConnectionConfig.java
```diff
@@ -302,6 +302,9 @@ public class ConnectionConfig
     @JsonIgnore
     private OBInstanceType instanceType;
 
+    @JsonIgnore
+    private OBInstanceRoleType instanceRoleType;
```
yhilmare, line 306: It's too customized. This class also needs to be oriented to other data sources. This way it will be too closely bound to OB.

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/ConnectionConfig.java
```diff
@@ -426,6 +429,14 @@ public int queryTimeoutSeconds() {
         return queryTimeoutSeconds;
     }
 
+    @JsonIgnore
+    public String getInternalSessionInitScript() {
```
yhilmare, line 433: how about this?

`ConnectionAdapter` gets different implementation in different env, we may provide a special implementation and we can add `set ob_read_consistency='WEAK'` to session init script.

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/ConnectionConfig.java
MarkPotato777: I don't think so. Think about the `clusterName`, `tenantName`, `sysTenantUsername`, and `sysTenantPassword` in `ConnectionConfig`, they are also OceanBase-specific fields. Additionally, the method `getDefaultSchema` is also bound to `DialectType`. 

On the other hand, I don't think that checking if the dialect is OceanBase within this `getInternalSessionInitScript` contradicts what you've said about this class being oriented to other data sources.

It's unlikely that you would use only one class to represent all data sources without introducing features specific to different types of data sources within this class.

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/ConnectionConfig.java
MarkPotato777: I don't think it's called 'customized'. Yes, the class `ConnectionConfig` is designed to be oriented to different types of data sources, but it doesn't mean we should only involve the common part of each data source. Think about the design of `DBTable` in schema accessors.

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/ConnectionConfig.java
MarkPotato777: `ConnectionAdapter`'s design is a kind of wield btw. I think `ConnectionAdapter` is not oriented to `Environment` but to `Connection` itself. What I expect if that this interface would have implementations of different dialect types, but the current implementation of `ConnectionAdapter` seems to be oriented to deploy environments.

MarkPotato777: I created a new implementation of `ConnectionInitializer` called `BackupInstanceInitializer` and moved the script into it.

yhilmare, line 433: this method should be deleted?

server/odc-service/src/main/java/com/oceanbase/odc/service/session/factory/DruidDataSourceFactory.java
```diff
@@ -59,7 +62,9 @@ public DataSource getDataSource() {
         String jdbcUrl = getJdbcUrl();
         String username = getUsername();
         String password = getPassword();
-        DruidDataSource dataSource = new InnerDataSource(new SessionCreatedInitializer(connectionConfig));
+        DruidDataSource dataSource =
+                new InnerDataSource(Arrays.asList(new DataSourceInitScriptInitializer(connectionConfig),
```
yhilmare, line 66: we may add BackupInstanceInitializer first and add DataSourceInitScriptInitializer second

server/odc-service/src/main/java/com/oceanbase/odc/service/pldebug/session/AbstractDebugSession.java
```diff
@@ -190,11 +193,12 @@ protected void enableDbmsOutput(Statement statement) {
     static class DebugDataSource extends SingleConnectionDataSource {
 
         private final List<String> initSqls;
-        private final SessionCreatedInitializer initializer;
+        private final List<ConnectionInitializer> initializers;
 
         public DebugDataSource(@NonNull ConnectionConfig connectionConfig, List<String> initSqls) {
             this.initSqls = initSqls;
-            this.initializer = new SessionCreatedInitializer(connectionConfig, true);
+            this.initializers = Arrays.asList(new DataSourceInitScriptInitializer(connectionConfig, true),
```
yhilmare, line 200: we may add BackupInstanceInitializer first and add DataSourceInitScriptInitializer second

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionTesting.java
```diff
@@ -243,23 +247,25 @@ private ConnectionConfig reqToConnectionConfig(TestConnectionReq req) {
...
         String jdbcUrl =
                 extensionPoint.generateJdbcUrl(getJdbcUrlProperties(config, schema));
 
         Properties properties = getTestConnectionProperties(config);
         properties.setProperty("socketTimeout", ConnectTypeUtil.REACHABLE_TIMEOUT_MILLIS + "");
         properties.setProperty("connectTimeout", ConnectTypeUtil.REACHABLE_TIMEOUT_MILLIS + "");
 
-        ConnectionInitializer initializer = new SessionCreatedInitializer(config, false);
+        List<ConnectionInitializer> initializers = new ArrayList<>();
+        initializers.addAll(
```
yhilmare, line 258: we may add BackupInstanceInitializer first and add DataSourceInitScriptInitializer second

server/odc-service/src/main/java/com/oceanbase/odc/service/pldebug/session/AbstractDebugSession.java
MarkPotato777: Sure, Nice suggestion! I have moved all BackupInstanceInitializer before DataSourceInitScriptInitializer~

server/odc-service/src/main/java/com/oceanbase/odc/service/session/factory/DruidDataSourceFactory.java
MarkPotato777: Sure~

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/ConnectionConfig.java
MarkPotato777: Yes~ I deleted it.

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionTesting.java
MarkPotato777: Sure~

yhilmare (approved):
lgtm

PR #2196, opened 2024-04-12
fix(connection): add some log about datasource lock

ungreat:

Fixes #

```docs

```

### DIFF
--- a/server/odc-core/src/main/java/com/oceanbase/odc/core/datasource/SingleConnectionDataSource.java
+++ b/server/odc-core/src/main/java/com/oceanbase/odc/core/datasource/SingleConnectionDataSource.java
@@ -131,7 +131,9 @@ public class SingleConnectionDataSource extends BaseClassBasedDataSource impleme
     private boolean tryLock() {
         try {
             boolean locked = this.lock.tryLock(timeOutMillis, TimeUnit.MILLISECONDS);
-            log.info("get connection lock success, locked={}, lock={}", locked, this.lock.hashCode());
+            if (locked) {
+                log.info("Get connection lock success, lock={}", this.lock.hashCode());
+            }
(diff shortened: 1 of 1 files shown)


server/odc-core/src/main/java/com/oceanbase/odc/core/datasource/SingleConnectionDataSource.java
```diff
@@ -146,6 +147,7 @@ private Connection getConnectionProxy(@NonNull Connection connection) {
                     new Class[] {Connection.class},
                     new CloseIgnoreInvocationHandler(connection, this.lock));
         } catch (Exception e) {
+            log.info("get connection error unlock, hashcode=" + this.lock.hashCode());
```
yhilmare, line 150: log.warn

server/odc-core/src/main/java/com/oceanbase/odc/core/datasource/SingleConnectionDataSource.java
```diff
@@ -130,6 +130,7 @@ protected void prepareConnection(Connection con) throws SQLException {
 
     private boolean tryLock() {
         try {
+            log.info("get connection try lock, hashcode=" + this.lock.hashCode());
```
yhilmare, line 133: we'd better get tryLock's result, if we obtain the lock and then we print this lock log~

yhilmare (approved):
lgtm

server/odc-core/src/main/java/com/oceanbase/odc/core/datasource/SingleConnectionDataSource.java
ungreat: That makes a lot of sense~

PR #2197, opened 2024-04-12
enhancement(security): add public webhook host into security ignore file

LuckyPickleZZ:
add public webhook host into security ignore file


yhilmare (approved):
lgtm

PR #2199, opened 2024-04-12
enhancement(log):opt scheduled task log

guowl3:
#### What type of PR is this?
type-feature

The scheduled task logs are difficult to review. We are now adding a unique ID to locate task logs more easily.

Fixes #

```docs

```

### DIFF
--- a/server/odc-server/src/main/resources/log4j2.xml
+++ b/server/odc-server/src/main/resources/log4j2.xml
@@ -47,6 +47,20 @@
             </DefaultRolloverStrategy>
         </RollingFile>
 
+        <RollingFile name="SchedulerFileAppender" fileName="${LOG_DIRECTORY}/scheduler.log"
+                     filePattern="${LOG_DIRECTORY}/odc-%d{yyyy-MM-dd}-%i.log">
+            <PatternLayout pattern="[%date{yyyy-MM-dd HH:mm:ss.SSS}][%thread][%level][%class][%line][ScheduledTask-%X{scheduleId}-%X{scheduleTaskId}]: %message%n"/>
+            <Policies>
+                <SizeBasedTriggeringPolicy size="100MB"/>
+            </Policies>
+            <DefaultRolloverStrategy max="20">
+                <Delete basePath="${LOG_DIRECTORY}" maxDepth="2">
+                    <IfFileName glob="*.log*"/>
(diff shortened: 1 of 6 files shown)


server/odc-server/src/main/resources/log4j2.xml
```diff
@@ -47,6 +47,20 @@
             </DefaultRolloverStrategy>
         </RollingFile>
 
+        <RollingFile name="SchedulerFileAppender" fileName="${LOG_DIRECTORY}/scheduler.log"
+                     filePattern="${LOG_DIRECTORY}/odc-%d{yyyy-MM-dd}-%i.log">
```
ungreat, line 51: filePattern should have the same prefix with filePattern.

server/odc-server/src/main/resources/log4j2.xml
```diff
@@ -752,6 +766,14 @@
             <AppenderRef ref="ScheduleTaskWarnRoutingAppender"/>
         </Logger>
 
+        <Logger name="com.oceanbase.odc.service.quartz" level="INFO" additivity="true">
+            <AppenderRef ref="SchedulerFileAppender"/>
+        </Logger>
+
+        <Logger name="com.oceanbase.odc.service.schedule.job" level="INFO" additivity="true">
```
yhilmare, line 773: partitionplan job is using this appender 'AllRoutingPartitionPlanAppender'

server/odc-service/src/main/java/com/oceanbase/odc/service/quartz/OdcJobListener.java
```diff
@@ -92,6 +93,10 @@ public void jobToBeExecuted(JobExecutionContext context) {
         ScheduleEntity scheduleEntity =
                 scheduleRepository.findById(scheduleId)
                         .orElseThrow(() -> new NotFoundException(ResourceType.ODC_SCHEDULE, "id", scheduleId));
+        ScheduleTaskContextHolder.trace(scheduleEntity.getId(), scheduleEntity.getJobType().name(), null);
```
yhilmare, line 96: why should we set these thread local? 'SchedulerFileAppender' seems like don't need any variables

yhilmare (approved):
lgtm, btw, we may test for partition plan job, cause that there  has been an appender for it~

PR #2211, opened 2024-04-15
enhancement(jpa): add json type

ungreat:
#### What type of PR is this?

add json type.

eg.
First at all, you have to def custom type .

```java
@TypeDefs({@TypeDef(name = "Json", typeClass = JsonType.class)})
@Entity
public class Example {
}

```

Then annotate filed

```java
  @Type(type = "Json")
  private ExampleObject example;
```

The `example` filed will  persisted as a string type, such as varchar.

Fixes #

```docs

```


yizhouxw:
may same feature as Conveter:
image

ungreat:
> may same feature as Conveter: image

A fairly obvious advantage is that, by using `JsonType`, the type required for deserialization can be dynamically obtained based on the type of the field currently being annotated.

yizhouxw (approved):
awesome generic json column solution!

PR #2214, opened 2024-04-15
fix(connection): add version check before set weak read consistency

MarkPotato777:
Before OceanBase 4.x, we need to execute `set ob_read_consistency = weak ` to connect to a backup cluster; but after OceanBase 4.x, we do not need to(and not recommended to) do this. So this PR add a OceanBase version check.

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #2219, opened 2024-04-15
fix(sql-execute): disable full link trace  on OB 4.1.x

### LINKED ISSUE #2218: [Bug]: full link trace is unavailable on OB 4.1.x
### ODC version

4.2.4

### OB version

4.1.x

### What happened?

Full link trace is unavailable on OB 4.1.x, sometimes it would return empty set.
And this would cause empty trace id.
image

This should not be enabled on OB 4.1.x

Execute `show trace` on OB 4.1.x .

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

Full link trace is unavailable on OB 4.1.x, sometimes it would return empty set.
And this would cause empty trace id.
So we should disable this feature on OB 4.1.x

Fixes #2218 

```docs

```


yizhouxw (approved):
lgtm

yhilmare (approved):
lgtm

PR #2220, opened 2024-04-15
enhancement(connection): optimize error message when testing the connection to an OceanBase backup instance

MarkPotato777:
#### What type of PR is this?

type-enhancement
module-connection
This PR refined the error message when testing the connection to an OceanBase instance.
If it is OBOracle mode and executes `SELECT USER_ID FROM ALL_USERS WHERE ROWNUM < 1`, if it's a backup instance, then it will cause an error message 'ErrorCode = 16000, SQLState = HY000, Details = ORA-16000: standby cluster support weak read only'. We catch this message and then inform users to set ob_read_consistency to WEAK in Advanced Settings.

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #2224, opened 2024-04-15
enhancement(connection): stop synchronizing databases whose cluster no longer exists

MarkPotato777:
#### What type of PR is this?

type-enhancement
module-connection

Current database synchronization involves all databases, including those with no longer existing clusters. In this case, we are not supposed to sync them.

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #2227, opened 2024-04-16
fix(permission): permission error message is not readable

### LINKED ISSUE #2226: [Bug]: error message of resource role permission is not readable
### ODC version

4.2.4

### OB version

not related

### What happened?

1. create a new odc user
2. login odc based on this user
3. attempt visit a project that isn't belong to him/her
截屏2024-04-16 10 11 33

as you can see, the error message is not readable.

the error message of resource role permission should be readable~

follow 'what happened'

_No response_

yhilmare:
#### What type of PR is this?

type-bug

error message of permission is not readable and this pr fix this issue~

Fixes #2226 

```docs

```

### DIFF
--- a/server/odc-core/src/main/java/com/oceanbase/odc/core/authority/permission/ConnectionPermission.java
+++ b/server/odc-core/src/main/java/com/oceanbase/odc/core/authority/permission/ConnectionPermission.java
@@ -119,7 +119,7 @@ public class ConnectionPermission extends ResourcePermission {
 
     @Override
     public String toString() {
-        return "ConnectionPermission " + this.resourceType + ":" + this.resourceId + " " + getActions(this.mask);
+        return ResourceType.ODC_CONNECTION.getLocalizedMessage() + ":" + this.resourceId + ": " + getActions(this.mask);
     }
 
 }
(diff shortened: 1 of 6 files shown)


server/odc-core/src/main/java/com/oceanbase/odc/core/authority/permission/ConnectionPermission.java
```diff
@@ -119,7 +119,7 @@ protected void initPermissionMask(int mask) {
 
     @Override
     public String toString() {
-        return "ConnectionPermission " + this.resourceType + ":" + this.resourceId + " " + getActions(this.mask);
+        return ResourceType.ODC_CONNECTION.getLocalizedMessage() + ":" + this.resourceId + ": " + getActions(this.mask);
```
smallsheeeep, line 122: Where does this copywriting come into play?

yhilmare: `com.oceanbase.odc.core.authority.exception.AccessDeniedException`

smallsheeeep (approved):

PR #2233, opened 2024-04-16
enhancement(schedule): terminate the schedule if the database does not exist.

guowl3:
#### What type of PR is this?
type-feature

This PR introduces a check for the database status to prevent unnecessary resource consumption. Previously, scheduled tasks would continue to be scheduled even when the database was nonexistent, consuming resources. With this update, once the database is deleted, the scheduling of tasks ceases.

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduleService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduleService.java
@@ -228,6 +228,34 @@ public class ScheduleService {
         scheduleRepository.updateStatusById(scheduleConfig.getId(), ScheduleStatus.TERMINATION);
     }
 
+    /**
+     * The method detects whether the database required for scheduled task operation exists. It returns
+     * true and terminates the scheduled task if the database does not exist. If the database exists, it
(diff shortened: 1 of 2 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduleService.java
```diff
@@ -228,6 +228,36 @@ public void terminate(ScheduleEntity scheduleConfig) throws SchedulerException {
...
+            Database database;
+            try {
+                database = databaseService.getBasicSkipPermissionCheck(scheduleEntityOptional.get().getDatabaseId());
+            } catch (NotFoundException e) {
+                database = null;
+            }
+            if (database == null || !database.getExisted()) {
+                try {
+                    log.info(
+                            "The database for scheduled task operation does not exist, and the schedule is being terminated.scheduleId={}",
```
yizhouxw, line 248: `.` --> `,`
use comma separated style for log output

MarkPotato777 (approved):
LGTM

yhilmare (approved):
lgtm

PR #2242, opened 2024-04-17
fix(taskframework): alarm notification is not  effective

krihy:

fix alarm notification is not effective, AlarmUtils.warn is only print log but not send notification, so  call `AlarmUtils.alarm` to  send notification

```docs

```


ungreat (approved):
lgtm

yhilmare (approved):
lgtm

PR #2245, opened 2024-04-17
fix(task): precheck query flow task instance not found when transaction uncommited

krihy:

we  found FlowTaskInstance is not exists when flowableAdaptor#getTaskInstanceByActivityId, so use `RetryExecutor` 5 times until transaction commit

```docs

```


yizhouxw:
new solution : https://github.com/oceanbase/odc/pull/2255

PR #2249, opened 2024-04-18
fix(notification): failed to query connection info

LuckyPickleZZ:
#### What type of PR is this?
type-bug

When query connection info , we would call `AuthenticationFacade#currentUser` for uid on alibaba cloud.
So we set context here.

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/notification/helper/EventBuilder.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/notification/helper/EventBuilder.java
@@ -264,78 +264,81 @@ public class EventBuilder {
                 log.warn("failed to query creator info.", e);
             }
         }
-        if (labels.containsKey(CONNECTION_ID)) {
-            try {
-                ConnectionConfig connectionConfig = connectionService.getForConnectionSkipPermissionCheck(
-                        labels.getLongFromString(CONNECTION_ID));
-                labels.put(CLUSTER_NAME, connectionConfig.getClusterName());
-                labels.put(TENANT_NAME, connectionConfig.getTenantName());
-                Environment environment =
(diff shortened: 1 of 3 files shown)


yizhouxw:
if the whitelist is empty, the checker will always thinks that this URL is legal
> seems make sense

LuckyPickleZZ:
> if the whitelist is empty, the checker will always thinks that this URL is legal
> 
> > seems make sense

Removed this.

yhilmare (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/notification/helper/EventBuilder.java
```diff
@@ -254,85 +255,89 @@ private Event ofSchedule(ScheduleEntity schedule, TaskEvent status) {
...
-                labels.put(CLUSTER_NAME, connectionConfig.getClusterName());
-                labels.put(TENANT_NAME, connectionConfig.getTenantName());
-                Environment environment =
-                        environmentService.detailSkipPermissionCheck(connectionConfig.getEnvironmentId());
-                labels.put(ENVIRONMENT, environment.getName());
-            } catch (Exception e) {
-                log.warn("failed to query connection info.", e);
-            }
-        }
         if (labels.containsKey(CREATOR_ID)) {
```
MarkPotato777, line 258: Why don't we move this if-brach to the try... finally... block below?

LuckyPickleZZ: because the context is set in this brach. It is the same whether it is placed in finally or not.

MarkPotato777 (approved):
LGTM

PR #2250, opened 2024-04-18
fix(taskframework): add pod config from properties

krihy:

set podPendingTimeoutSeconds in podConfig

```docs

```


yhilmare (approved):
lgtm

PR #2253, opened 2024-04-18
fix(bastion): could not access SQL console when using OAuth for account integration

smallsheeeep:
#### What type of PR is this?
type-bug
module-bastion integration

When using Oauth for account integration, then can not using bastion integration for accessing a temp SQL console.

The root cause is that ODC must create a built-in project for user and then the bastion integration will work. But if user use Oauth for login, then the project will not be create.

In This PR, I fix it by create user related built-in project once the bastion integration is enbled. 

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/iam/UserService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/iam/UserService.java
@@ -42,6 +42,7 @@ import javax.validation.constraints.NotNull;
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.lang.Validate;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.data.domain.Page;
 import org.springframework.data.domain.PageRequest;
 import org.springframework.data.domain.Pageable;
@@ -60,7 +61,6 @@ import org.springframework.validation.annotation.Validated;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.google.common.base.MoreObjects;
-import com.oceanbase.odc.common.security.PasswordUtils;
 import com.oceanbase.odc.common.util.StringUtils;
 import com.oceanbase.odc.core.authority.permission.Permission;
 import com.oceanbase.odc.core.authority.permission.ResourcePermission;
@@ -71,7 +71,6 @@ import com.oceanbase.odc.core.shared.PreConditions;
 import com.oceanbase.odc.core.shared.Verify;
 import com.oceanbase.odc.core.shared.constant.Cipher;
 import com.oceanbase.odc.core.shared.constant.ErrorCodes;
-import com.oceanbase.odc.core.shared.constant.OdcConstants;
 import com.oceanbase.odc.core.shared.constant.PermissionType;
 import com.oceanbase.odc.core.shared.constant.ResourceType;
 import com.oceanbase.odc.core.shared.constant.RoleType;
@@ -96,6 +95,7 @@ import com.oceanbase.odc.metadb.iam.UserRoleEntity;
 import com.oceanbase.odc.metadb.iam.UserRoleRepository;
 import com.oceanbase.odc.metadb.iam.UserSpecs;
 import com.oceanbase.odc.service.automation.model.TriggerEvent;
+import com.oceanbase.odc.service.collaboration.project.ProjectService;
 import com.oceanbase.odc.service.common.response.CustomPage;
 import com.oceanbase.odc.service.common.response.PaginatedData;
 import com.oceanbase.odc.service.common.util.SpringContextUtil;
@@ -176,6 +176,12 @@ public class UserService {
     @Autowired
     private UserOrganizationService userOrganizationService;
 
+    @Autowired
+    private ProjectService projectService;
+
+    @Value("${odc.integration.bastion.enabled:false}")
+    private boolean bastionEnabled;
+
     private final PasswordEncoder passwordEncoder = new BCryptPasswordEncoder();
     private final List<Consumer<PasswordChangeEvent>> postPasswordChangeHooks = new ArrayList<>();
     private final List<Consumer<UserDeleteEvent>> postUserDeleteHooks = new ArrayList<>();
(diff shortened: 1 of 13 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/iam/model/User.java
```diff
@@ -147,6 +154,23 @@ public User(UserEntity userEntity) {
         this.extraProperties = userEntity.getExtraPropertiesJson();
     }
 
+    public static UserEntity innerEntity(@NonNull String account, @NonNull String name, @NonNull Long organizationId) {
```
yhilmare, line 157: what does 'inner' mean?

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/auth/CustomAuthenticationSuccessHandler.java
```diff
@@ -74,6 +76,12 @@ public class CustomAuthenticationSuccessHandler extends SavedRequestAwareAuthent
     @Qualifier("organizationResourceMigrator")
     private OrganizationResourceMigrator organizationResourceMigrator;
 
+    @Autowired
+    private ProjectService projectService;
+
+    @Value("${odc.integration.bastion.enabled:false}")
```
yhilmare, line 82: we may user another flag to control this behavior

yhilmare (dismissed):
lgtm

ungreat (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/UserService.java
```diff
@@ -192,44 +190,55 @@ public class UserService {
...
+        } else {
+            user = save(userEntity);
+            if (CollectionUtils.isNotEmpty(roleNames)) {
+                List<Long> roleIds = new ArrayList<>();
+                List<Role> roles = new ArrayList<>();
+                for (String roleName : roleNames) {
+                    RoleEntity roleEntity =
+                            roleRepository.findByNameAndOrganizationId(roleName, user.getOrganizationId())
+                                    .orElseThrow(() -> new NotFoundException(ResourceType.ODC_ROLE, "name", roleName));
+                    PreConditions.validArgumentState(roleEntity.getType() != RoleType.INTERNAL, ErrorCodes.BadArgument,
```
MarkPotato777, line 224: Prefer use `Verify` instead of `PreConditions`

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/UserService.java
```diff
@@ -192,44 +190,55 @@ public class UserService {
...
-                UserRoleEntity userRoleEntity = new UserRoleEntity();
-                userRoleEntity.setUserId(userId);
-                userRoleEntity.setRoleId(roleId);
-                userRoleEntity.setOrganizationId(organizationId);
-                userRoleEntity.setCreatorId(OdcConstants.DEFAULT_ADMIN_USER_ID);
-                userRoleRepository.saveAndFlush(userRoleEntity);
-
-                roleIds.add(roleId);
-                roles.add(new Role(roleEntity.get()));
+    public User createIfNotExistsOrUpdate(UserEntity userEntity, List<String> roleNames) {
```
MarkPotato777, line 202: method name suggestion: createIfNotExistsOrUpdate --> upsert

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/model/User.java
```diff
@@ -147,6 +154,24 @@ public User(UserEntity userEntity) {
         this.extraProperties = userEntity.getExtraPropertiesJson();
     }
 
+    public static UserEntity autoCreatedUserEntity(@NonNull String account, @NonNull String name,
```
MarkPotato777, line 157: why do we put this method into `User`? It seems like it is not related to `User` at all.

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/UserService.java
smallsheeeep: It is no matter, `PreConditions` also throw exception with i18n message.

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/model/User.java
smallsheeeep: Avoid the same codes appear several times. Build user entity in User class is readable.

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/UserService.java
smallsheeeep: Good suggestion, I will modify it in the future.

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/model/User.java
MarkPotato777: Moving this method into the `UserEntity` makes more sense intuitively.

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/UserService.java
MarkPotato777: Since you renamed the method name in this PR, so why not just rename it to a more understandable name directly in this PR?

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/UserService.java
MarkPotato777: It's not about the effect, it's about the usage.
We generally use `PreConditions` at the very beginning of a method for parameter validation, while `Verify` is typically used for logical validation.

smallsheeeep: Okay, done!

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/UserService.java
smallsheeeep: Done.

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/model/User.java
smallsheeeep: Okay~

MarkPotato777 (approved):
LGTM

yhilmare (approved):
lgtm

PR #2259, opened 2024-04-18
fix(sql-execute): failed to kill query

yhilmare:
#### What type of PR is this?

type-bug

User will not kill query when he/she switch on the feature 'continue when sql executed failed' cause that the interrupt exception is also considered as an exception which is ignored.

I add an flag to indicate this situation

Fixes #

```docs

```


LuckyPickleZZ (approved):
lgtm

PR #2267, opened 2024-04-22
fix(notification): use blacklist to check webhook instead of whitelist

LuckyPickleZZ:
#### What type of PR is this?
type-enhancement

We used whitelist to prevent SSRF from notification webhook, but if so, a large number of IPs need to be added to the whitelist.
Therefore, we switched to using blacklist for verification and only filtered out the IPs in the blacklist


yizhouxw (commented):
we may use both whitelist and blacklist, all list works while value is not empty.

while whitelist is set, blacklist will not works.

server/odc-service/src/main/java/com/oceanbase/odc/service/notification/helper/ChannelConfigValidator.java
```diff
@@ -90,10 +92,12 @@ private void validateWebhookChannelConfig(WebhookChannelConfig channelConfig) {
         Verify.verify(
                 channelConfig.getWebhook().startsWith("http://") || channelConfig.getWebhook().startsWith("https://"),
                 "Webhook should start with 'http://' or 'https://'");
-        Verify.verify(SSRFChecker.checkUrlInWhiteList(channelConfig.getWebhook(),
-                integrationConfigProperties.getUrlWhiteList()),
-                "The webhook is not in white list, please add it into system configuration with the key 'odc.integration.url-white-list'");
-
+        try {
```
yizhouxw, line 95: we may use both whitelist and blacklist, the mechanism enabled while value is not empty. and if whitelist is set, blacklist will not works.

LuckyPickleZZ: done. Only check blacklist when whitelist is empty.

yizhouxw (approved):
lgtm

MarkPotato777 (approved):
LGTM

PR #2271, opened 2024-04-22
security: test masking using rounding algorithm may cause denial of service (DoS) risk

### LINKED ISSUE #2270: [Bug]: Test masking using rounding algorithm may cause denial of service (DoS) risk
### ODC version

4.2.4

### OB version

any

### What happened?

When I test masking using rounding algorithm, and input a big decimal such as 100e999999, then it occupied large CPU usage. And if I mock many these request at same time ,then DoS may occur.

Defend these risk.

Ditto.

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug

When I test masking using rounding algorithm, and input a big decimal such as 100e999999, then it occupied large CPU usage. And if I mock many these request at same time ,then DoS may occur.
This PR fix it by adding a `OdcBigDecimalChecker` to check if the BigDecimal value passed in has a denial of service risk.

Fixes #2270 

```docs

```

### DIFF
--- a/server/odc-common/src/main/java/com/oceanbase/odc/common/security/OdcBigDecimalChecker.java
+++ b/server/odc-common/src/main/java/com/oceanbase/odc/common/security/OdcBigDecimalChecker.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright (c) 2023 OceanBase.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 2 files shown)


yizhouxw:
title may start with `security:`

server/odc-common/src/main/java/com/oceanbase/odc/common/security/OdcBigDecimalChecker.java
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * @author gaoda.xy
+ * @date 2024/4/22 15:31
+ */
+public class OdcBigDecimalChecker {
+
+    private static final Pattern DIGITS_PATTERN = Pattern.compile("[+-]?[0-9.]+");
+
+    private static final int MAX_EXPONENT_LENGTH = 4;
```
yhilmare, line 30: is it better that this value can be configurable~

smallsheeeep:
> title may start with `security:`

okay, done!

server/odc-common/src/main/java/com/oceanbase/odc/common/security/OdcBigDecimalChecker.java
smallsheeeep: There is a public method that can specify this value. 
For ODC, 4 (meansing 1e9999) is enough for almost all user usage, and is low cost for CPU processing.

PR #2272, opened 2024-04-22
fix(flow): cancel pending task cause "Some tasks is not found"

krihy:
#### What type of PR is this?

type-bug
module-flow
flowableAdaptor#getFlowableElementByType will return `CALL_BACK` task can  cause` Verify.verify(tasks.size() == elements.size())`  exception, so  filter `CALL_BACK` user task in  getFlowableElementByType result

cancel pending task which is timer execute cause "Some tasks is not found"

test passed

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/model/CallBackUserTaskConstants.java
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+package com.oceanbase.odc.service.flow.task.model;
+
+/**
+ * @author yaobin
+ * @date 2024-04-22
+ * @since 4.2.4
+ */
+public class CallBackUserTaskConstants {
```
yhilmare, line 23: RuntimeTaskConstants may meet your needs

krihy: looks more better

krihy: done

yhilmare (approved):
lgtm

PR #2276, opened 2024-04-23
fix(dlm): get log failed when the task framework was switched during task execution.

guowl3:
…ion caused log query failure.

#### What type of PR is this?
type-bug

Previously, we determined whether a task was executed within the task framework by checking for the presence of a JobId. However, if the task framework was switched during task execution, this would lead to an erroneous judgment.

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #2277, opened 2024-04-23
fix(web-framework): csrfToken API does not return token

### LINKED ISSUE #2263: [Bug]: csrf token returns everywhere
### ODC version

4.2.4

### OB version

not related

### What happened?

clear all cookies and visit any api that need login 

截屏2024-04-19 15 30 40

as you can see, the api returns a xsrf token which is not expected~

we expect that only login and csrf token api can return this token

follow 'what happened'

_No response_

yizhouxw:
#### What type of PR is this?
type-bug

in application integration sceanario, client need fetch csrfToken before call POST/PUT/PATCH/DELETE method.

/api/v1/user/csrfToken API does not return xsrf-token in response cookie due it was set in authenticaton whitelist.

Fixes #2263


smallsheeeep (approved):
LGTM

yhilmare (approved):
lgtm

PR #2287, opened 2024-04-24
fix(partition-plan): failed to start up a partition plan on oceanbase before 3.x

### LINKED ISSUE #2285: [Bug]: An error occurs when the partition plan is delivered using a version earlier than ob3.x
### ODC version

ODC 4.2.4-bp1

### OB version

OB 1479，ob2.x

### What happened?

An error occurs when the partition plan is delivered using a version earlier than ob3.x
image

The partition plan task is successfully delivered. 

Use OB1479 and OB2.x versions to deliver partition plan tasks.

_No response_

yhilmare:
#### What type of PR is this?

type-bug

the query statement before oceanbase 3.x is invalid. I fix them in this pr~

Fixes #2285 

```docs

```


PeachThinking (approved):
LGTM

PR #2289, opened 2024-04-24
fix(objectstorage): wrong log/database change/rollback plan download URLs if cloud storage is enabled

MarkPotato777:
#### What type of PR is this?

type-bug

This PR fixed wrong log/database change/rollback plan download URLs if cloud storage is enabled.

The root cause is that the so-called `downloadUrl` is not an actual download link. Instead, it is composed of `bucketName/objectName` following a convention that not everyone might be aware of. Then, at some point, this `downloadUrl` is parsed to generate the real download link. To ensure minimal changes to the codebase, this PR awkwardly adapts to this poor design in an ugly way.

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/util/TaskId2DownloadUrlsCache.java
```diff
@@ -0,0 +1,122 @@
...
+
+/**
+ * @Author: Lebie
+ * @Date: 2021/10/20 下午2:07
+ * @Description: [This class is responsible for maintaining relationships from async task id to its
+ *               result set temp OSS download url.]
+ */
+@Service
+@Slf4j
+public class TaskId2DownloadUrlsCache {
```
MarkPotato777, line 49: rename class

MarkPotato777: renamed to `TaskDownloadUrlsProvider`

yhilmare (approved):
lgtm

PR #2292, opened 2024-04-25
fix(taskframework): column 'job_id' cannot be null

krihy:
fix job attribute Column 'job_id' is null 

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #2307, opened 2024-04-29
fix(connection): connection may be blocked

yhilmare:
#### What type of PR is this?

type-bug

`Connection#isValid` may be blocked when connection is occupied cause that driver execute a sql to verify a connect

Fixes #

```docs

```


MarkPotato777 (approved):
LGTM

yizhouxw (approved):
lgtm

PR #2313, opened 2024-05-06
fix(dlm): the target table does not exist

guowl3:
#### What type of PR is this?
type-bug

When closing the task framework, the automatic table creation function did not adapt to the custom target table name, resulting in an error of the target table not existing.

Fixes #

```docs

```


yhilmare (commented):
lgtm

yhilmare (approved):
lgtm

PR #2314, opened 2024-05-06
fix(notification): lost sign when updating channel

### LINKED ISSUE #2305: [Bug]: Edit and modify the message channel. The signature key will be deleted after saving.
### ODC version

ODC424bp2

### OB version

independent

### What happened?

Edit and modify the message channel. The signature key will be deleted after saving.
image

。

Edit and modify the message channel. The signature key will be deleted after saving.

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

When user edit and update channel with sign, the sign would be lost after updating.
This is due to the annotation of `@OneToMany`, which caused the new `ChannelPropertyEntity` to overwrite the old ones when editing `ChannelEntity`.
This PR modifies the `save` to the `update` method of `ChannelRepository`. And abandoned the cascading updates of JPA, instead manually deleting ChannelProperty and inserting them again.

Fixes #2305

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/notification/NotificationService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/notification/NotificationService.java
@@ -174,17 +174,18 @@ public class NotificationService {
 
         BaseChannelConfig channelConfig = channel.getChannelConfig();
         if (channelConfig instanceof WebhookChannelConfig && ((WebhookChannelConfig) channelConfig).getSign() == null) {
-            channelPropertyRepository.deleteByChannelIdAndKeyNotEquals(entity.getId(), "sign");
-        } else {
-            channelPropertyRepository.deleteByChannelId(entity.getId());
+            String sign = ((WebhookChannelConfig) channelMapper.fromEntityWithConfig(entity).getChannelConfig())
+                    .getSign();
+            ((WebhookChannelConfig) channelConfig).setSign((sign));
         }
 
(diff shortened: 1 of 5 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/notification/NotificationService.java
```diff
@@ -174,17 +174,18 @@ public Channel updateChannel(@NotNull Long projectId, @NotNull Channel channel)
...
         }
 
         ChannelEntity toBeSaved = channelMapper.toEntity(channel);
-        toBeSaved.setCreatorId(authenticationFacade.currentUserId());
-        toBeSaved.setOrganizationId(authenticationFacade.currentOrganizationId());
-        toBeSaved.setProjectId(projectId);
+        channelRepository.update(toBeSaved);
+
+        channelPropertyRepository.deleteByChannelId(channel.getId());
+        toBeSaved.getProperties().forEach(property -> channelPropertyRepository.save(property));
```
yizhouxw, line 186: may involve performace risk ?

LuckyPickleZZ: Compared to the original @onetonay cascading update, there is no performance rollback. And since the number of properties is in single digits, there should be no performance issues.

yizhouxw (approved):
lgtm

smallsheeeep (approved):
LGTM

MarkPotato777 (approved):
LGTM

PR #2315, opened 2024-05-06
fix(result-set-export): use console datasource to avoid socket timeout

### LINKED ISSUE #2115: [Bug]: JDBC socket timeout is 60 seconds in Export data
### ODC version

424

### OB version

all

### What happened?

image

Make the socketTimeout of the export task a configurable item

create new Export tasks

_No response_

### Cloud

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

Now ODC use backend datasource of session to export result set. But it's socket timeout is 60 seconds and cannot be configured.
If a query is executed for more than 60 seconds, such as `select sleep (70)`, the socket will time out with an error: Read timed out.
So we use console datasource , which socket timeout could be configured.

Fixes #2115

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/resultset/ResultSetExportTask.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/resultset/ResultSetExportTask.java
@@ -220,7 +220,7 @@ public class ResultSetExportTask implements Callable<ResultSetExportResult> {
         List<MaskingAlgorithm> algorithms = parameter.getRowDataMaskingAlgorithms();
         Map<String, Map<String, List<OrdinalColumn>>> catalog2TableColumns = new HashMap<>();
         try {
-            SyncJdbcExecutor syncJdbcExecutor = session.getSyncJdbcExecutor(ConnectionSessionConstants.BACKEND_DS_KEY);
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/resultset/ResultSetExportTask.java
```diff
@@ -220,7 +220,7 @@ private void setColumnConfig(DataTransferConfig config, ResultSetExportTaskParam
         List<MaskingAlgorithm> algorithms = parameter.getRowDataMaskingAlgorithms();
         Map<String, Map<String, List<OrdinalColumn>>> catalog2TableColumns = new HashMap<>();
         try {
-            SyncJdbcExecutor syncJdbcExecutor = session.getSyncJdbcExecutor(ConnectionSessionConstants.BACKEND_DS_KEY);
+            SyncJdbcExecutor syncJdbcExecutor = session.getSyncJdbcExecutor(ConnectionSessionConstants.CONSOLE_DS_KEY);
```
yhilmare, line 223: this datasource may be occupied~

LuckyPickleZZ: This session was constructed by result export task, and there is no reuse. So there will be no conflict.

yhilmare (approved):
lgtm

PeachThinking (approved):
LGTM

PR #2318, opened 2024-05-06
fix(connection): get connection not in lock

ungreat:
#### What type of PR is this?
The connection.isValid() is a blocking method, and since no timeout has been set, it will block until the currently executing SQL statement is finished.The current implementation of the reset method requires acquiring a lock with a timeout of 10 seconds, which prevents the method from being blocked indefinitely.

Fixes #

```docs

```


yhilmare:
repeat with #2307

PR #2320, opened 2024-05-07
fix(osc): osc job is terminated when database id is not exist

krihy:
#### What type of PR is this?

type-bug
module-schedule
because schedule task input param from osc do not contain project id, so quartz job will be deleted by  `ScheduleService#terminateIfDatabaseNotExisted`

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/OnlineSchemaChangeFlowableTask.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/OnlineSchemaChangeFlowableTask.java
@@ -35,6 +35,8 @@ import com.oceanbase.odc.metadb.schedule.ScheduleEntity;
 import com.oceanbase.odc.metadb.schedule.ScheduleTaskEntity;
 import com.oceanbase.odc.metadb.schedule.ScheduleTaskRepository;
 import com.oceanbase.odc.metadb.task.TaskEntity;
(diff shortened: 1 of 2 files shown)


guowl3 (dismissed):
LGTM

LuckyPickleZZ (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/OnlineSchemaChangeFlowableTask.java
```diff
@@ -267,9 +269,9 @@ private ScheduleEntity createScheduleEntity(ConnectionConfig connectionConfig,
         scheduleEntity.setAllowConcurrent(false);
         scheduleEntity.setCreatorId(creatorId);
         scheduleEntity.setOrganizationId(organizationId);
-        // todo project id database id
+        // todo project id
         scheduleEntity.setProjectId(1L);
```
yizhouxw, line 273: what is projectId = 1 means ?

krihy: project id is not used for osc , but i also query by `databaseService.getBasicSkipPermissionCheck` and and set in scheduleEntity

yhilmare (approved):
lgtm

LuckyPickleZZ (approved):

PR #2328, opened 2024-05-08
fix(db-browser): partition definitions is not ordered

yhilmare:
#### What type of PR is this?

type-bug

the partition definitions is not ordered on OceanBase 4.x and MySQL

Fixes #

```docs

```


PeachThinking (approved):
LGTM

PR #2330, opened 2024-05-08
fix(dlm): correct the task status when the task exits abnormally

guowl3:
#### What type of PR is this?
type-bug

When a node crashes causing an abnormal exit and the task status is not updated correctly, this PR automatically corrects the historical task status upon the next scheduled execution of the planned task.

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #2336, opened 2024-05-08
fix(account-integration): failed to login after updating user extra properties

### LINKED ISSUE #2326: [Bug]: existed user fails to login on play site
### ODC version

4.2.4-bp2

### OB version

-

### What happened?

After upgrading 4.2.4, when a existed user login on play site, there would be an error.
image

This error occurred by chance.

From the code, it can be seen that the UserEntity that reported the error was created by the `UserEntity.autoCreatedEntity` method. In fact, it should be created using the `createIfNotExists` method,

login with diffrent `extra_properties_json`

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
module-account integration (SSO, Bastion, PlayStation)

When updating the extra properties of existed user, then the user cannot login. The root cause is due to bad codes and this PR fix it.

Fixes #2326 

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/iam/UserService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/iam/UserService.java
@@ -203,14 +203,14 @@ public class UserService {
         User user;
         Optional<UserEntity> optionalUser = userRepository.findByAccountName(userEntity.getAccountName());
         if (optionalUser.isPresent()) {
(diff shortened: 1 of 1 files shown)


LuckyPickleZZ (approved):
LGTM

yhilmare (approved):
lgtm

yizhouxw (approved):
lgtm

MarkPotato777 (approved):
LGTM

PR #2338, opened 2024-05-08
build: upgrade workflow action version to v4

smallsheeeep:
#### What type of PR is this?
type-build

Upgrade workflow action version to v4. 
This is because that some old version actions will not be available soon in the future, and v4 improve the performance of upload/download artifact.

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #2339, opened 2024-05-08
fix(odp-sharding): cannot connect to odp sharding

yizhouxw:

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionValidator.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionValidator.java
@@ -50,6 +50,12 @@ public class ConnectionValidator {
         PreConditions.validInHostWhiteList(connection.getHost(), connectProperties.getHostWhiteList());
         PreConditions.validArgumentState(environmentService.exists(connection.getEnvironmentId()),
                 ErrorCodes.BadRequest, null, "invalid environment id");
+        if (connection.getType().isDefaultSchemaRequired()) {
+            PreConditions.notBlank(connection.getDefaultSchema(), "connection.defaultSchema");
+        }
+        if (connection.getDialectType().isOracle()) {
+            connection.setDefaultSchema(null);
+        }
     }
 
     void validateForUpdate(ConnectionConfig connection, ConnectionConfig saved) {
@@ -58,11 +64,6 @@ public class ConnectionValidator {
                         || Objects.equals(connection.getType(), saved.getType()),
                 ErrorCodes.FieldUpdateNotSupported, new Object[] {"connection.type"},
                 "Cannot change field 'connection.type'");
-        PreConditions.validRequestState(
-                Objects.isNull(connection.getDialectType())
-                        || Objects.equals(connection.getDialectType(), saved.getDialectType()),
-                ErrorCodes.FieldUpdateNotSupported, new Object[] {"connection.dialectType"},
-                "Cannot change field 'connection.dialectType'");
     }
 
     void validatePrivateConnectionTempOnly(Boolean temp) {
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/session/ConnectConsoleService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/session/ConnectConsoleService.java
@@ -44,6 +44,7 @@ import org.springframework.validation.annotation.Validated;
 import com.oceanbase.odc.common.lang.Holder;
 import com.oceanbase.odc.common.unit.BinarySize;
 import com.oceanbase.odc.common.unit.BinarySizeUnit;
+import com.oceanbase.odc.common.util.ExceptionUtils;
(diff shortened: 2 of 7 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionValidator.java
```diff
@@ -58,11 +64,6 @@ void validateForUpdate(ConnectionConfig connection, ConnectionConfig saved) {
                         || Objects.equals(connection.getType(), saved.getType()),
                 ErrorCodes.FieldUpdateNotSupported, new Object[] {"connection.type"},
                 "Cannot change field 'connection.type'");
-        PreConditions.validRequestState(
```
yizhouxw, line 61: there is no dialectType filed now, it is includes in connectType

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionValidator.java
```diff
@@ -50,6 +50,12 @@ void validateForUpsert(ConnectionConfig connection) {
         PreConditions.validInHostWhiteList(connection.getHost(), connectProperties.getHostWhiteList());
         PreConditions.validArgumentState(environmentService.exists(connection.getEnvironmentId()),
                 ErrorCodes.BadRequest, null, "invalid environment id");
+        if (connection.getType().isDefaultSchemaRequired()) {
+            PreConditions.notBlank(connection.getDefaultSchema(), "connection.defaultSchema");
+        }
+        if (connection.getDialectType().isOracle()) {
+            connection.setDefaultSchema(null);
```
yizhouxw, line 57: skip defaultSchema for oracle

server/odc-service/src/main/java/com/oceanbase/odc/service/session/ConnectConsoleService.java
```diff
@@ -551,7 +552,7 @@ private SqlExecuteResult generateResult(@NonNull ConnectionSession connectionSes
             try (TraceStage s = watch.start(SqlExecuteStages.INIT_COLUMN_INFO)) {
                 result.initColumnInfo(connectionSession, resultTable, schemaAccessor);
             } catch (Exception e) {
-                log.warn("Failed to init column comment", e);
+                log.warn("Failed to init column comment, reason={}", ExceptionUtils.getSimpleReason(e));
```
yizhouxw, line 555: for avoid too much exception stack trace in log output

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionValidator.java
yhilmare: It seems like not useful, we will set default schema when creating session:

```java
ConnectSessionService#create

@SkipAuthorize("check permission internally")
    public ConnectionSession create(@NotNull CreateSessionReq req) {
        ...
        if (StringUtils.isNotEmpty(schemaName)) {
            connection.setDefaultSchema(schemaName);
        }
        ...
    }
```

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionValidator.java
yhilmare: so, should we check if the connection's connectType is equals to the saved's connectType

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionValidator.java
yhilmare: this pr #1797 involved this problem. let us check this problem in detail:
1.  defaultSchema should not be set by the front end at any time except for odp-sharding
2. we will ignore the default schema when connect type is odp-sharding if we add `JsonIgnore` at the filed which cause this issue

so, can we do this to fix this issue:
1. remove the `JsonIgnore` annotation from the `defaultschema` field
2. make odp-sharding as a special case, that is to say we should do this in two method `ConnectionService#innerCreate` and `ConnectionService#update`, like this:
```java
if (connection.getDialectType != DialectType.ODP_SHARDING) {
    connection.setDefaultSchema(null);
}
```

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionValidator.java
yizhouxw: > so, should we check if the connection's connectType is equals to the saved's connectType

yes, the connectType exists

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionValidator.java
yizhouxw: `setDefaultSchema(null)` move to ConnectionService now

MarkPotato777 (approved):
LGTM

yhilmare (approved):
lgtm

PR #2340, opened 2024-05-08
fix(dlm):  the task status was set to completed when the task was canceled

guowl3:
…e task was cancelled.

#### What type of PR is this?
type-bug

Fixed the issue where the task status was set to completed when the task was canceled.

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #2341, opened 2024-05-08
fix(taskframework): oss log info is ignored when cancel job

krihy:
#### What type of PR is this?

type-bug
module-taskframework
log info is  missing when cancel job or close task-framework, so we should update oss `LOG_STORAGE_ALL_OBJECT_ID` and `LOG_STORAGE_BUCKET_NAME` in job_attribute in report result

- case 1: 

1. start job, job is running and can show log
2. cancel job 
result: log can be show normally

- case 2:

1. start job, job is running and can show log
2. close task-framework and restart
result: job is failed and log can be show normally

above 2 cases has test passed

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/task/service/StdTaskFrameworkService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/task/service/StdTaskFrameworkService.java
@@ -310,8 +310,13 @@ public class StdTaskFrameworkService implements TaskFrameworkService {
             log.warn("Job identity is not exists by id {}", taskResult.getJobIdentity().getId());
             return;
         }
-        if (je.getStatus().isTerminated() || je.getStatus() == JobStatus.CANCELING) {
-            log.warn("Job {} is finished, ignore result, currentStatus={}", je.getId(), je.getStatus());
+        if (je.getStatus() == JobStatus.CANCELING) {
+            saveOrUpdateLogMetadata(taskResult, je.getId(), je.getStatus());
+            return;
+        }
+
+        if (je.getStatus().isTerminated()) {
+            log.warn("Job is finished, ignore result,jobId={}, currentStatus={}", je.getId(), je.getStatus());
             return;
         }
         int rows = updateJobScheduleEntity(taskResult, je);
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/task/service/StdTaskFrameworkService.java
```diff
@@ -310,8 +313,11 @@ public void handleResult(TaskResult taskResult) {
             log.warn("Job identity is not exists by id {}", taskResult.getJobIdentity().getId());
             return;
         }
-        if (je.getStatus().isTerminated() || je.getStatus() == JobStatus.CANCELING) {
-            log.warn("Job {} is finished, ignore result, currentStatus={}", je.getId(), je.getStatus());
+        if (je.getStatus() == JobStatus.CANCELING || !taskFrameworkEnabledProperties.isEnabled()) {
+            saveOrUpdateLogMetadata(taskResult, je.getId(), je.getStatus());
+            return;
+        } else if (je.getStatus().isTerminated()) {
+            log.warn("Job is finished, ignore result, jobId={}, currentStatus={}", je.getId(), je.getStatus());
```
yizhouxw, line 320: why ignore result while job is finieshed ?

server/odc-service/src/main/java/com/oceanbase/odc/service/task/service/StdTaskFrameworkService.java
```diff
@@ -310,8 +313,11 @@ public void handleResult(TaskResult taskResult) {
             log.warn("Job identity is not exists by id {}", taskResult.getJobIdentity().getId());
             return;
         }
-        if (je.getStatus().isTerminated() || je.getStatus() == JobStatus.CANCELING) {
-            log.warn("Job {} is finished, ignore result, currentStatus={}", je.getId(), je.getStatus());
+        if (je.getStatus() == JobStatus.CANCELING || !taskFrameworkEnabledProperties.isEnabled()) {
```
yizhouxw, line 316: why save log meta while  task framework not enabled

krihy: task is running before close task framework,   when close task framework we send stop command to pod, then task will be stopped and upload log to oss,  we save the report result  about oss info

yizhouxw (approved):
merge first, may refactor later

PR #2347, opened 2024-05-09
fix(full-link-trace): enable trace by default on OB 4.x

LuckyPickleZZ:
#### What type of PR is this?
type-bug

Sometimes, due to performance issues, some tenants may default to disabling full link traces. This will result in the trace function of ODC becoming unavailable.
To avoid this issue, we enabled full link trace by default on version 4. x.
Please note that `DBMS_MONITOR.OB_SESSION_TRACE_ENABLE` called here  is session level and does not affect the overall tenant. Therefore, the impact on tenant performance should be controllable


yizhouxw (approved):
lgtm

yhilmare (approved):
lgtm

PR #2352, opened 2024-05-10
feat(dlm):data clearing  tasks support data check before delete

kiko-art:
#### What type of PR is this?
type-feature

Data cleansing previously did not support data validation; now, a new feature for validation has been added.

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/dlm/model/DataDeleteParameters.java
```diff
@@ -33,6 +33,14 @@ public class DataDeleteParameters implements TaskParameters {
 
     private String databaseName;
 
+    private Long targetDatabaseId;
+
+    private String targetDatabaseName;
+
+    private String sourceDataSourceName;
```
guowl3, line 40: What does this field do?

kiko-art: Because data clearing needs to stay synchronized with data archiving, when there are two data sources, it is necessary to display information about each data source.

guowl3 (approved):
LGTM

smallsheeeep (dismissed):
LGTM

PR #2353, opened 2024-05-10
fix(obclient): remove unexpected characters to avoid obclient display incorrectly

### LINKED ISSUE #2301: [Bug]: The command line window displays an incorrect sql file that is too long
### ODC version

ODC 4.2.4-bp

### OB version

-

### What happened?

Typing too long sql in the command-line window does not display correctly.
image

Typing long sql in a command-line window displays correctly.

Enter the following sql in the command line window.
```
SELECT TENANT_ID, TENANT_NAME, TENANT_ROLE, STATUS, SWITCHOVER_STATUS, SYNC_SCN, REPLAYABLE_SCN,READABLE_SCN, RECOVERY_UNTIL_SCN FROM oceanbase.DBA_OB_TENANTS where tenant_type<>'META';
```

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

When input a long sequence in obclient, it would return a sequence with ` \r` in it. And it would cause the cursor going back to head of the line . 
So we replace these unexpected characters manually.

Fixes #2301

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/websocket/OBClientReadThread.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/websocket/OBClientReadThread.java
@@ -18,6 +18,7 @@ package com.oceanbase.odc.service.websocket;
 import java.io.InputStream;
 import java.util.Arrays;
 import java.util.function.Consumer;
+import java.util.regex.Pattern;
 
 import lombok.extern.slf4j.Slf4j;
 
@@ -28,6 +29,8 @@ import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 public class OBClientReadThread extends Thread {
+    private static final Pattern UNEXPECTED_SEQUENCE_PATTERN = Pattern.compile(" \\r");
+
     private Consumer<String> messageConsumer;
     private InputStream inputStream;
     private volatile Boolean stop = false;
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/websocket/OBClientReadThread.java
```diff
@@ -28,6 +29,8 @@
 
 @Slf4j
 public class OBClientReadThread extends Thread {
+    private static final Pattern UNEXPECTED_SEQUENCE_PATTERN = Pattern.compile(" \\r");
```
yhilmare, line 32: what is the white space of this pattern used for? `\\s`?

yizhouxw (dismissed):
lgtm,

current only windows/linux support command line windows,
windows user `\r\n`, linux use `\n` as line separator, but windows is compatible with `\n` also.

server/odc-service/src/main/java/com/oceanbase/odc/service/websocket/OBClientReadThread.java
yizhouxw: white space may not  affect line breaking behavior

yizhouxw: ` \\r` may change to `\\r`

LuckyPickleZZ: We should not change this , because the obclient error sequences are always with a blank space~
image

yhilmare (approved):
lgtm

yizhouxw (approved):
lgtm

PR #2356, opened 2024-05-10
feature(security): auth whitelist support configurable

ungreat:
#### What type of PR is this?
auth whitelist support configurable

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/config/CommonSecurityProperties.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/config/CommonSecurityProperties.java
@@ -123,7 +123,7 @@ public class CommonSecurityProperties {
     @Value("${odc.web.security.cors.allowedOrigins:*}")
     private List<String> corsAllowedOrigins;
 
-    @Value("${odc.web.security.auth-white_list:}")
+    @Value("${odc.web.security.auth-white-list:}")
     private List<String> securityAuthWhiteList;
 
     @PostConstruct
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/config/CommonSecurityProperties.java
```diff
@@ -121,6 +123,9 @@ public class CommonSecurityProperties {
     @Value("${odc.web.security.cors.allowedOrigins:*}")
     private List<String> corsAllowedOrigins;
 
+    @Value("${odc.web.security.auth-white_list:}")
```
yhilmare, line 126: odc.web.security.auth-whitelist？

ungreat: done!

yhilmare (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/config/CommonSecurityProperties.java
```diff
@@ -121,6 +123,9 @@ public class CommonSecurityProperties {
     @Value("${odc.web.security.cors.allowedOrigins:*}")
     private List<String> corsAllowedOrigins;
 
+    @Value("${odc.web.security.auth-whitelist:}")
+    private List<String> securityAuthWhiteList;
```
yizhouxw, line 127: 1. we may distinguish buitin whitelist and self defined whitelist, or the concept is confused due same configruation name and field name 'auth-whitelist'.

2. WhiteList or Whitelist will be okay either,  but if filed name 'WhiteList', the configuraiton item key may 'white-list'.

3. if the new configration is for cloud scenario only, we may not involved here

ungreat: Agree with points 1 and 2！For the point 3, it can indeed be done in other ways，i will close this pr.

PR #2365, opened 2024-05-11
fix(taskframework): job is not started  in resource idle node when deployed multi-nodes

krihy:
#### What type of PR is this?

type-bug

job is not started  when multi nodes has enough resource,  when JobStore acquire trigger but resource is not available , we should release it  and give a chance to other node acquire trigger.

test case with two nodes, task be started in equalization in two nodes.
image

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/task/config/ResourceDetectJobStore.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/task/config/ResourceDetectJobStore.java
@@ -59,7 +59,7 @@ public class ResourceDetectJobStore extends LocalDataSourceJobStore {
 
     private boolean shouldAcquire(OperableTrigger trigger) {
         TaskFrameworkProperties taskFrameworkProperties = TaskFrameworkPropertiesSupplier.getSupplier().get();
-        // If resource is not available return null and give a chance to other node acquire trigger
+        // If resource is not available return false and give a chance to other node acquire trigger
         if (!ResourceDetectUtil.isResourceAvailable(taskFrameworkProperties)) {
             if (TriggerKey.triggerKey("startPreparingJob", JobConstants.ODC_JOB_MONITORING)
                     .equals(trigger.getKey())) {
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/task/config/ResourceDetectJobStore.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/task/config/ResourceDetectJobStore.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024 OceanBase.
+ * Copyright (c) 2023 OceanBase.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -13,7 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package com.oceanbase.odc.service.task.config;
 
 import java.util.LinkedList;
@@ -28,12 +27,15 @@ import com.oceanbase.odc.service.common.util.SpringContextUtil;
 import com.oceanbase.odc.service.task.constants.JobConstants;
 import com.oceanbase.odc.service.task.schedule.ResourceDetectUtil;
 
+import lombok.extern.slf4j.Slf4j;
+
 /**
  * @author yaobin
  * @date 2024-05-10
  * @since 4.2.4
  */
-public class JobStoreSupportDelegate extends LocalDataSourceJobStore {
+@Slf4j
+public class ResourceDetectJobStore extends LocalDataSourceJobStore {
 
     @Override
     public List<OperableTrigger> acquireNextTriggers(long noLaterThan, int maxCount, long timeWindow)
@@ -61,6 +63,7 @@ public class JobStoreSupportDelegate extends LocalDataSourceJobStore {
         if (!ResourceDetectUtil.isResourceAvailable(taskFrameworkProperties)) {
             if (TriggerKey.triggerKey("startPreparingJob", JobConstants.ODC_JOB_MONITORING)
                     .equals(trigger.getKey())) {
+                log.debug("StartPreparingJob trigger is discarded because no resource.");
                 return false;
             }
         }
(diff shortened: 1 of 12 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/task/config/TaskFrameworkConfiguration.java
```diff
@@ -84,6 +86,9 @@ public SchedulerFactoryBean taskFrameworkSchedulerFactoryBean(DataSource dataSou
         schedulerFactoryBean.setSchedulerName(taskFrameworkSchedulerName);
         schedulerFactoryBean.setStartupDelay(taskFrameworkProperties.getQuartzStartDelaySeconds());
         schedulerFactoryBean.setTaskExecutor(executor);
+        Properties properties = new Properties();
+        properties.put("org.quartz.jobStore.class", ResourceDetectJobStore.class.getName());
```
yhilmare, line 90: magic value

yhilmare: is it a standard method to set custom JobStore?

server/odc-service/src/main/java/com/oceanbase/odc/service/task/config/ResourceDetectJobStore.java
```diff
@@ -0,0 +1,73 @@
...
+            }
+        }
+        return filteredTriggers;
+    }
+
+    private boolean shouldAcquire(OperableTrigger trigger) {
+        TaskFrameworkProperties taskFrameworkProperties = TaskFrameworkPropertiesSupplier.getSupplier().get();
+        // If resource is not available return false and give a chance to other node acquire trigger
+        if (!ResourceDetectUtil.isResourceAvailable(taskFrameworkProperties)) {
+            if (TriggerKey.triggerKey("startPreparingJob", JobConstants.ODC_JOB_MONITORING)
```
yhilmare, line 64: what does "startPreparingJob" means? why is it this value?

yhilmare:
I can not accept your solution:

1. StartPreparingJob is a daemon job on every node, we'd better not disable it on a specific node
2. we have already got rateLimiter `MonitorProcessRateLimiter`, why does it work?
3. Will other nodes be affected if you `releaseAcquiredTrigger` in `ResourceDetectJobStore`?

server/odc-service/src/main/java/com/oceanbase/odc/service/task/config/TaskFrameworkConfiguration.java
krihy: yes,  it's standard method,  reference `SchedulerFactoryBean` line 575 ,
`mergedProps.putIfAbsent(StdSchedulerFactory.PROP_JOB_STORE_CLASS, LocalDataSourceJobStore.class.getName());`
but the magic value can be replaced by constants

krihy:
> I can not accept your solution:
> 
> 1. StartPreparingJob is a daemon job on every node, we'd better not disable it on a specific node
> 2. we have already got rateLimiter `MonitorProcessRateLimiter`, why does it work?
> 3. Will other nodes be affected if you `releaseAcquiredTrigger` in `ResourceDetectJobStore`?

in cluster model,  all quartz node seize trigger, and be fired exact only once,  if one node  with no resource acquire trigger and fired it, the job will not be started in `StartPreparingJob` in `QuartzSchedulerThread` while loop  and  other node with enough resource will not acquire trigger

server/odc-service/src/main/java/com/oceanbase/odc/service/task/config/ResourceDetectJobStore.java
krihy: `JobStore` acquire one trigger at one time, so we should identify the trigger name, and detect current node resource is enough

yizhouxw:
> > I can not accept your solution:
> > 
> > 1. StartPreparingJob is a daemon job on every node, we'd better not disable it on a specific node
> > 2. we have already got rateLimiter `MonitorProcessRateLimiter`, why does it work?
> > 3. Will other nodes be affected if you `releaseAcquiredTrigger` in `ResourceDetectJobStore`?
> 
> in cluster model, all quartz node seize trigger, and be fired exact only once, if one node with no resource acquire trigger and fired it, the job will not be started in `StartPreparingJob` in `QuartzSchedulerThread` while loop and other node with enough resource will not acquire trigger

as we discused before, in cluster model, each node should do StartPreparingJob

yhilmare:
> > I can not accept your solution:
> > 
> > 1. StartPreparingJob is a daemon job on every node, we'd better not disable it on a specific node
> > 2. we have already got rateLimiter `MonitorProcessRateLimiter`, why does it work?
> > 3. Will other nodes be affected if you `releaseAcquiredTrigger` in `ResourceDetectJobStore`?
> 
> in cluster model, all quartz node seize trigger, and be fired exact only once, if one node with no resource acquire trigger and fired it, the job will not be started in `StartPreparingJob` in `QuartzSchedulerThread` while loop and other node with enough resource will not acquire trigger



> > I can not accept your solution:
> > 
> > 1. StartPreparingJob is a daemon job on every node, we'd better not disable it on a specific node
> > 2. we have already got rateLimiter `MonitorProcessRateLimiter`, why does it work?
> > 3. Will other nodes be affected if you `releaseAcquiredTrigger` in `ResourceDetectJobStore`?
> 
> in cluster model, all quartz node seize trigger, and be fired exact only once, if one node with no resource acquire trigger and fired it, the job will not be started in `StartPreparingJob` in `QuartzSchedulerThread` while loop and other node with enough resource will not acquire trigger

so, why doesn't we rely on `MonitorProcessRateLimiter` but disable the `StartPreparingJob`?

server/odc-service/src/main/java/com/oceanbase/odc/service/task/constants/JobConstants.java
```diff
@@ -59,6 +59,8 @@ public class JobConstants {
 
     public static final String ODC_JOB_MONITORING = "odcJobMonitoring";
 
+    public static final String START_PREPARING_JOB = "startPreparingJob";
```
yhilmare, line 62: StdJobScheduler#initStartPreparingJob may use this constants

PR #2371, opened 2024-05-13
fix(encryption): exception occurs when invoking RSA decryption concurrently

### LINKED ISSUE #2368: [Bug]: fail to decrypt password
### ODC version

4.2.4

### OB version

no

### What happened?

fail to decrypt password,log:
[com.oceanbase.odc.service.encryption.SensitivePropertyHandlerImpl][137]: Failed to decrypt sensitive property, [org.springframework.web.servlet.handler.AbstractHandlerExceptionResolver][208]: Resolved [org.springframework.http.converter.HttpMessageNotReadableException: JSON parse error: Unable to invoke Cipher due to illegal block size; nested exception is com.fasterxml.jackson.databind.JsonMappingException: Unable to invoke Cipher due to illegal block size (through reference chain: 

decrypt password success

Concurrent decryption password

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
module-RSA encryption

There exist concurrent issue in RSA encryption and this PR fix it by add `sychronized` keyword for the `encrypt` and `decrypt` methods.

Fixes #2368 

Self-test passed.

```docs

```

### DIFF
--- a/server/odc-common/src/main/java/com/oceanbase/odc/common/crypto/RsaBytesEncryptor.java
+++ b/server/odc-common/src/main/java/com/oceanbase/odc/common/crypto/RsaBytesEncryptor.java
@@ -74,7 +74,7 @@ public class RsaBytesEncryptor implements BytesEncryptor {
     }
 
     @Override
-    public byte[] encrypt(byte[] origin) {
+    public synchronized byte[] encrypt(byte[] origin) {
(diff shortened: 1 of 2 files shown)


yhilmare (approved):
lgtm

PeachThinking (approved):
lgtm

PR #2382, opened 2024-05-14
fix(database): database sync failed after updated an invalid datasource to a valid datasource

### LINKED ISSUE #2376: [Bug]: The data source did not set a connection timeout so that the database could not be synchronized after editing
### ODC version

ODC 4.2.4-bp2

### OB version

~

### What happened?

image
Create a data source with a connection timeout, then edit it as a normal data source, and synchronize the database with an error.

Synchronize database successfully.

Follow what happened.

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug

If the data source was previously failing to connect, changing it to a connectable data source will result in a database synchronization failure with the error message: "Data Source is synchronizing". The reason for this is that the JDBC lock from the last data source synchronization was never released, preventing other tasks from acquiring the lock. The lock was not released because` DataSource.getConnection` got stuck, even though we set the `connectTimeout` to `5000 ms`. It may be related to our ` DataSource.getLoginTimeout` always returning 0. The specific mechanism needs to be further discussed with our JDBC colleagues. 

This PR has just implemented a timeout mechanism at the service layer to ensure that deadlocks do not occur.

Fixes #2376 

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
@@ -440,7 +440,8 @@ public class DatabaseService {
         List<String> blockedDatabaseNames = listBlockedDatabaseNames(connection.getDialectType());
         DataSource teamDataSource = new OBConsoleDataSourceFactory(connection, true, false).getDataSource();
         ExecutorService executorService = Executors.newFixedThreadPool(1);
-        Future<Connection> connectionFuture = executorService.submit(() -> teamDataSource.getConnection());
+        Future<Connection> connectionFuture = executorService.submit(
+                (Callable<Connection>) teamDataSource::getConnection);
         Connection conn = null;
         try {
             conn = connectionFuture.get(5, TimeUnit.SECONDS);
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
```diff
@@ -512,11 +519,20 @@ private void syncTeamDataSources(ConnectionConfig connection) {
...
             }
-        } catch (SQLException e) {
+        } catch (ExecutionException e) {
             deleteDatabaseIfClusterNotExists(e, connection.getId(),
                     "update connect_database set is_existed = 0 where connection_id=?");
             throw new IllegalStateException(e);
+        } catch (Exception e) {
+            throw new IllegalStateException(e);
         } finally {
+            if (conn != null) {
```
yhilmare, line 529: we may shutdown the thread pool

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
```diff
@@ -512,11 +519,20 @@ private void syncTeamDataSources(ConnectionConfig connection) {
                         "update connect_database set table_count=?, collation_name=?, charset_name=?, project_id=? where id = ?";
                 jdbcTemplate.batchUpdate(update, toUpdate);
             }
-        } catch (SQLException e) {
+        } catch (ExecutionException e) {
```
yhilmare, line 522: we may get rootCause and judge if it is an instance of `SQLException`

yizhouxw (dismissed):

yizhouxw (approved):

PR #1401, opened 2024-01-16
fix(db-browser): fix the visualization of mysql table structure design and supports strings enclosed in single quotes

ISADBA:
#### What type of PR is this?
type-bug
dbbrowser

#### What this PR does / why we need it:
fix the visualization of MySQL table structure design, supporting strings enclosed in single quotes ('').

#### Which issue(s) this PR fixes:

Fixes #1317

#### Special notes for your reviewer:
have not frontend change need。

#### Additional documentation e.g., usage docs, etc.:
none

PR #2388, opened 2024-05-14
fix(shadowtable): bad sql grammer when table column's default value is a string type in MySQL

### LINKED ISSUE #2379: [Bug]: The sql statement generated by the shadow table task does not put quotes on the default field
### ODC version

ODC424bp1

### OB version

~

### What happened?

```
create table test_char(`col12` char(10) NOT NULL DEFAULT '中文字段' COMMENT '第十二列');
```
sql statements generated by the shadow table task:
```
create table test_char(`col12` char(10) NOT NULL DEFAULT  中文字段  COMMENT '第十二列');
```

Execution report syntax error.

Follow what happened.

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug

When the original table column's default is a string type, the generated ddl does not quote the default value which causes to sql execution failed.

Fixes #2379 

This issue was introduced by https://github.com/oceanbase/odc/pull/1401

```docs

```

### DIFF
--- a/server/odc-common/src/main/java/com/oceanbase/odc/common/util/StringUtils.java
+++ b/server/odc-common/src/main/java/com/oceanbase/odc/common/util/StringUtils.java
@@ -18,6 +18,7 @@ package com.oceanbase.odc.common.util;
 import java.net.URLDecoder;
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.UUID;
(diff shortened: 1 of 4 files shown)


yhilmare (dismissed):
lgtm

PeachThinking (dismissed):
LGTM

server/odc-common/src/main/java/com/oceanbase/odc/common/util/StringUtils.java
```diff
@@ -163,6 +168,31 @@ public static String unquoteSqlValue(final String str, final char wrapChar, fina
         return unescapeUseDouble(unwrap, escapeChars);
     }
 
+    public static void quoteColumnDefaultValues(DBTable table) {
```
yhilmare, line 171: we may rename it: quoteColumnDefaultValuesForMysql

yhilmare (approved):
lgtm

yizhouxw (approved):
lgtm

PeachThinking (approved):
LGTM

PR #2389, opened 2024-05-14
fix(encryption): RSA decrypting failed if already decrypted a invalid input string

### LINKED ISSUE #2368: [Bug]: fail to decrypt password

smallsheeeep:
#### What type of PR is this?
type-bug

This PR #2371 does not really fix the bug #2368 , and this PR fix it. 

Fixes #2368 

Unit test case added.

```docs

```


yhilmare (approved):
lgtm

PR #2403, opened 2024-05-15
fix(data-transfer): fix incorrect task result update

LuckyPickleZZ:
#### What type of PR is this?
type-bug

In DataTransferRuntimeFlowableTask, we use TaskService#update to upgrade task result, then use TaskService#success to finish task. 
If the `success` is called before `update`, the real result woule be overwriten.

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/FlowTaskCallBackApprovalService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/FlowTaskCallBackApprovalService.java
@@ -103,7 +103,7 @@ public class FlowTaskCallBackApprovalService {
         FlowInstanceEntity flowInstance = getFlowInstance(flowInstanceId);
         FlowableElement flowableElement = getFlowableElementOfUserTask(flowTaskInstanceId);
 
-        Await.await().timeout(60).timeUnit(TimeUnit.SECONDS)
+        Await.await().timeout(60).timeUnit(TimeUnit.SECONDS).period(1).periodTimeUnit(TimeUnit.SECONDS)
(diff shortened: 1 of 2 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/FlowTaskCallBackApprovalService.java
```diff
@@ -103,7 +103,7 @@ private void completeTask(long flowInstanceId, long flowTaskInstanceId, FlowNode
         FlowInstanceEntity flowInstance = getFlowInstance(flowInstanceId);
         FlowableElement flowableElement = getFlowableElementOfUserTask(flowTaskInstanceId);
 
-        Await.await().timeout(60).timeUnit(TimeUnit.SECONDS)
+        Await.await().timeout(60).timeUnit(TimeUnit.SECONDS).period(1).periodTimeUnit(TimeUnit.SECONDS)
```
MarkPotato777, line 106: Could you please explain why we need this line of code?

LuckyPickleZZ: This code is for task framework by @krihy . It would block 5s to wait approval node to be created. I changed this to wait only 1s .
But this modification is only for optimizing the code. The fix that truly works is another in  `BaseODCFlowTaskDelegate` .

MarkPotato777: Yes I know that your fix is in `BaseODCFlowTaskDelegate`, so I wonder what's the effect of this line of code.
Is this change necessary? Or will it affect the task framework? @krihy

yhilmare (approved):
lgtm

PR #2405, opened 2024-05-16
fix(ticket): failed to set download log file url

smallsheeeep:
#### What type of PR is this?
type-bug

When invoking `BaseODCFlowTaskDelegate#callback()`, the actual task may not executed adn has no log file. In this case, odc will throw exception. This  PR fix it.

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #2408, opened 2024-05-16
fix(taskframework): daemon job be fired at one time in cluster model

krihy:
#### What type of PR is this?

daemon job running be fired at one time in cluster model, so we remove datasource and quartz will use RAMJobStore as non-cluster model

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/task/config/TaskFrameworkConfiguration.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/task/config/TaskFrameworkConfiguration.java
@@ -75,11 +75,10 @@ public class TaskFrameworkConfiguration {
 
     @Lazy
     @Bean("taskFrameworkSchedulerFactoryBean")
-    public SchedulerFactoryBean taskFrameworkSchedulerFactoryBean(DataSource dataSource,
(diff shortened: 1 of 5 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/task/config/TaskFrameworkConfiguration.java
```diff
@@ -75,11 +73,10 @@ public JasyptEncryptorConfigProperties JasyptEncryptorConfigProperties(
 
     @Lazy
     @Bean("taskFrameworkSchedulerFactoryBean")
-    public SchedulerFactoryBean taskFrameworkSchedulerFactoryBean(DataSource dataSource,
```
guowl3, line 78: After switching to RAM mode, are there no concurrency issues with the existing tasks? I don't seem to see any locking logic implemented.

krihy: daemon job fetch every task record by lock, see `StartPreparingJob#startJob`

guowl3 (approved):
LGTM

yizhouxw (approved):
lgtm,
may improve performance later, currently multiple nodes may fetch same tasks, aquire conflict lock, which may cause unneccessary lock wait

PR #2423, opened 2024-05-17
feat(sql-execute): support service layer for query profile

LuckyPickleZZ:
#### What type of PR is this?
type-feature

This is the 3rd PR for query profile.

You should review api in `SqlDiagnoseController` `SqlDiagnoseExtensionPoint`. And `OBQueryProfileManager` in service layer alse should be cared.


server/odc-service/src/main/java/com/oceanbase/odc/service/session/ConnectConsoleService.java
```diff
@@ -698,4 +604,18 @@ private Integer checkQueryLimit(Integer queryLimit) {
         return queryLimit;
     }
 
+    private boolean await(@NotNull AsyncExecuteContext context, long timeout, TimeUnit unit) {
```
yhilmare, line 607: name suggestion: isSqlExecutionDone

server/odc-service/src/main/java/com/oceanbase/odc/service/session/ConnectConsoleService.java
```diff
@@ -401,48 +329,19 @@ public SqlAsyncExecuteResp streamExecute(@NotNull String sessionId,
...
+    public AsyncExecuteResultResp getMoreResults(@NotNull String sessionId, String requestId, Integer timeoutSeconds) {
         PreConditions.validArgumentState(Objects.nonNull(requestId), ErrorCodes.SqlRegulationRuleBlocked, null, null);
         ConnectionSession connectionSession = sessionService.nullSafeGet(sessionId);
         AsyncExecuteContext context =
                 (AsyncExecuteContext) ConnectionSessionUtil.getExecuteContext(connectionSession, requestId);
-        boolean shouldRemoveContext = context.isFinished();
+        int timeout = Objects.isNull(timeoutSeconds) ? DEFAULT_GET_RESULT_TIMEOUT_SECONDS : timeoutSeconds;
+        boolean shouldRemoveContext = false;
         try {
+            shouldRemoveContext = await(context, timeout, TimeUnit.SECONDS);
```
yhilmare, line 344: I think that we should not do this, assume that I get two sqls:
1. select 1 from dual;
2. select sleep(200) from dual;

the first one needs less time to execute but we await for both two sql executed completely for 1 second, It makes nonsense~

server/odc-core/src/main/resources/i18n/ErrorMessages.properties
```diff
@@ -195,3 +195,4 @@ com.oceanbase.odc.ErrorCodes.InvalidSqlExpression=An error occurred during the e
 com.oceanbase.odc.ErrorCodes.PartitionKeyDataTypeMismatch=The data type does not match. The expected data type for partition key {0} is {1}, but the actual data type is {2}.
 com.oceanbase.odc.ErrorCodes.TimeDataTypePrecisionMismatch=The time type precision does not match. The configured time interval precision is greater than the maximum precision of the partition key {0} data type.
 com.oceanbase.odc.ErrorCodes.DatabaseAccessDenied=Database access is denied because the current user does not have {0} permissions for the database.
+com.oceanbase.odc.ErrorCodes.ObQueryProfileNotSupported=Query profile is only available for OceanBase Database with versions equal to or higher than 4.2.0.
```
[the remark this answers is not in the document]
LuckyPickleZZ: It's 4.2.4.
Sure, I us variable instead of hard-code it.

server/odc-service/src/main/java/com/oceanbase/odc/service/queryprofile/OBQueryProfileManager.java
```diff
@@ -0,0 +1,108 @@
...
+                String key = PROFILE_KEY_PREFIX + traceId;
+                ConnectionSessionUtil.setBinaryContentMetadata(session, key, metaData);
+            } catch (IOException e) {
+                log.warn("Failed to cache profile.", e);
+            }
+        });
+    }
+
+    public SqlExplain getProfile(@NonNull String traceId, ConnectionSession session) {
+        if (session.isExpired()) {
```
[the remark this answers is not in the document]
LuckyPickleZZ: removed.

server/odc-service/src/main/java/com/oceanbase/odc/service/queryprofile/OBQueryProfileManager.java
```diff
@@ -0,0 +1,108 @@
...
+            throw new BadRequestException(ErrorCodes.ConnectionReset, null,
+                    ErrorCodes.ConnectionReset.getLocalizedMessage(null));
+        }
+        if (VersionUtils.isLessThan(ConnectionSessionUtil.getVersion(session), "4.2.0")) {
```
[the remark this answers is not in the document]
LuckyPickleZZ: done

server/odc-service/src/main/java/com/oceanbase/odc/service/queryprofile/OBQueryProfileManager.java
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * @author: liuyizhuo.lyz
+ * @date: 2024/4/19
+ */
+@Slf4j
+@Component
+public class OBQueryProfileManager {
+    private static final String PROFILE_KEY_PREFIX = "query-profile-";
+
+    private final ExecutorService executor =
```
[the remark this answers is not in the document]
LuckyPickleZZ: done, I added a thread pool in ScheduleConfiguration. 
It's core size and max size are `Math.max(SystemUtils.availableProcessors() * 2, 8)`

server/odc-service/src/main/java/com/oceanbase/odc/service/session/ConnectConsoleService.java
LuckyPickleZZ: This method was removed.

server/odc-service/src/main/java/com/oceanbase/odc/service/session/ConnectConsoleService.java
LuckyPickleZZ: done. We would use a more reasonable logic here. 
If the results are not empty, ODC will return them instantly. Otherwise, it will busy-waiting for 1s to get any result.

server/plugins/connect-plugin-api/src/main/java/com/oceanbase/odc/plugin/connect/model/diagnose/PlanGraph.java
```diff
@@ -0,0 +1,95 @@
...
+import com.oceanbase.odc.core.shared.model.QueryStatus;
+
+import lombok.Data;
+
+/**
+ * @author liuyizhuo.lyz
+ * @date 2024/4/11
+ */
+@Data
+public class PlanGraph {
```
yizhouxw, line 38: some fields are not easy to read

1. duration miss unit
2. overview contains Object as value of map
3. overview/statistics/topNodes，what is the key and value

often, map structure is a bad choice, as the map cannot explain itself.

server/odc-service/src/main/java/com/oceanbase/odc/service/queryprofile/OBQueryProfileManager.java
```diff
@@ -0,0 +1,109 @@
...
+        executor.execute(() -> {
+            if (session.isExpired()
+                    || ConnectionSessionUtil.getBinaryContentMetadata(session, PROFILE_KEY_PREFIX + traceId) != null) {
+                return;
+            }
+            try {
+                SqlExplain profile = session.getSyncJdbcExecutor(BACKEND_DS_KEY).execute(
+                        (ConnectionCallback<SqlExplain>) conn -> ConnectionPluginUtil
+                                .getDiagnoseExtension(session.getConnectType().getDialectType())
+                                .getQueryProfileByTraceId(conn, traceId));
```
yizhouxw, line 72: getQueryProfileByTraceId may throw Verify exception, however it is not handled here,
what is the expected behaivor after then, will it affect sql execution feature, or result in too much warn level log output, which contains exception stack trace print ?

MarkPotato777 (dismissed):
LGTM

server/odc-service/src/main/java/com/oceanbase/odc/service/queryprofile/OBQueryProfileManager.java
LuckyPickleZZ: Thanks, I fixed this by catching all the exceptions instead of IOException only.

yizhouxw (approved):
lgtm

PR #2435, opened 2024-05-20
build: fix build_jar.sh remove old version jar not works

yizhouxw:
fix build_jar.sh remove old version jar not works


yhilmare (approved):
lgtm

PR #2436, opened 2024-05-20
feat(object-search): individual space support global object search

### LINKED ISSUE #2432: [Feature]: Personal space support global object search

Personal space does not support global object search.

Personal space support global object search.

### Additional context

_No response_

### LINKED ISSUE #2443: [Bug]: Database object retrieval results not sorted by match level
### ODC version

4.3.0

### OB version

Not matter.

### What happened?

image
The result list is not sorted by matching level.

The result list is sorted by matching level.

.

_No response_

smallsheeeep:
#### What type of PR is this?
type-feature
module-global object search

This PR add support for global object search:
1. Add system config `odc.database.schema.global-search.enabled` to control the database object sync, the default value  is `true`.
2. Add a user config `odc.database.default.enableGlobalObjectSearch` to control whether open the global object search UI, the default value is `true`.
3. For web mode, if the system config value is `false`, then all sync operation will be ignored; if it is `true`, then we will only sync  databases in team space and these in individual space that the user config is `true`.
4. For desktop mode, if user config is `false`, then sync will be ignored.

Fixes #2432 
Fixes #2443 

Self-test passed.

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/db/schema/DBSchemaIndexService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/db/schema/DBSchemaIndexService.java
@@ -183,21 +183,38 @@ public class DBSchemaIndexService {
         Set<Database> databases = new HashSet<>();
         if (req.getResourceType() == ResourceType.ODC_CONNECTION) {
             Set<Database> dbs = new HashSet<>(databaseService.listExistDatabasesByConnectionId(req.getResourceId()));
-            Set<Long> projectIds = projectService.getMemberProjectIds(authenticationFacade.currentUserId());
-            databases.addAll(dbs.stream()
-                    .filter(e -> e.getProject() != null && projectIds.contains(e.getProject().getId()))
-                    .collect(Collectors.toSet()));
+            if (authenticationFacade.currentUser().getOrganizationType() == OrganizationType.INDIVIDUAL) {
+                ConnectionConfig config = connectionService.getBasicWithoutPermissionCheck(req.getResourceId());
+                if (!Objects.equals(authenticationFacade.currentUserId(), config.getCreatorId())) {
+                    throw new NotFoundException(ResourceType.ODC_CONNECTION, "id", req.getResourceId());
+                }
+                databases.addAll(dbs);
+            } else {
+                Set<Long> projectIds = projectService.getMemberProjectIds(authenticationFacade.currentUserId());
+                databases.addAll(dbs.stream()
+                        .filter(e -> e.getProject() != null && projectIds.contains(e.getProject().getId()))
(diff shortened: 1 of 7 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/db/schema/DBSchemaIndexService.java
```diff
@@ -153,7 +157,7 @@ public QueryDBObjectResp listDatabaseObjects(@NonNull @Valid QueryDBObjectParams
             resp.setDatabases(databases.stream().filter(e -> e.getName().toLowerCase().contains(key))
                     .collect(Collectors.toList()));
         }
-        Pageable pageable = PageRequest.of(0, 1000, Sort.by(Direction.DESC, DBColumnEntity_.DATABASE_ID));
+        Pageable pageable = PageRequest.of(0, MAX_SEARCH_SIZE, Sort.by(Direction.DESC, DBColumnEntity_.DATABASE_ID));
```
MarkPotato777, line 160: 1000 --> 10000? Is this change expected?

server/odc-service/src/main/resources/config-meta/user-config-meta.yml
```diff
@@ -170,4 +170,14 @@
   allowed_values: [ 'true','false' ]
   default_value: 'true'
   nullable: false
-  description: 'Whether the query is being rewritten in ob-oracle mode and increases the RowId.'
\ No newline at end of file
+  description: 'Whether the query is being rewritten in ob-oracle mode and increases the RowId.'
+- id: 18
+  category: 'user'
+  key: 'odc.database.default.enableGlobalObjectSearch'
```
MarkPotato777, line 176: The key is not the same as the system config key, is it expected?

server/odc-service/src/main/java/com/oceanbase/odc/service/db/schema/DBSchemaIndexService.java
smallsheeeep: I modified it from 10000 to 5000, which means that we will query at most 5000 records. It is excepted and has been tested. The time cost is about 200ms.

server/odc-service/src/main/resources/config-meta/user-config-meta.yml
smallsheeeep: This is user config, it is different from system config. It is excepted.

MarkPotato777 (approved):
LGTM

yhilmare (dismissed):
lgtm

MarkPotato777 (approved):

PeachThinking (approved):
lgtm

yhilmare (approved):
lgtm

PR #2445, opened 2024-05-21
fix(notification): DLM events were missed when task framework not enabled

### LINKED ISSUE #2335: [Bug]: Close the task framework and the message notification cannot be received
### ODC version

ODC424bp1

### OB version

independent

### What happened?

Close the task framework and the message notification cannot be received

。

close the framework
create task

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

When task framework is not enabled, the DLM jobs' event would not be created.

Fixes #2335

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/notification/helper/EventBuilder.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/notification/helper/EventBuilder.java
@@ -23,6 +23,7 @@ import static com.oceanbase.odc.service.notification.constant.EventLabelKeys.CRE
 import static com.oceanbase.odc.service.notification.constant.EventLabelKeys.CREATOR_NAME;
 import static com.oceanbase.odc.service.notification.constant.EventLabelKeys.DATABASE_ID;
 import static com.oceanbase.odc.service.notification.constant.EventLabelKeys.DATABASE_NAME;
+import static com.oceanbase.odc.service.notification.constant.EventLabelKeys.DESCRIPTION;
 import static com.oceanbase.odc.service.notification.constant.EventLabelKeys.ENVIRONMENT;
 import static com.oceanbase.odc.service.notification.constant.EventLabelKeys.PROJECT_ID;
 import static com.oceanbase.odc.service.notification.constant.EventLabelKeys.PROJECT_NAME;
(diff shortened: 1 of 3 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/notification/helper/EventBuilder.java
```diff
@@ -188,6 +189,7 @@ private Event ofTask(TaskEntity task, TaskEvent status) {
         labels.putIfNonNull(CREATOR_ID, task.getCreatorId());
         labels.putIfNonNull(TRIGGER_TIME, LocalDateTime.now().format(DATE_FORMATTER));
         labels.putIfNonNull(REGION, OB_ARN_PARTITION);
+        labels.putIfNonNull(DESCRIPTION, task.getDescription());
```
MarkPotato777, line 192: Is this change related to the issue?

server/odc-service/src/main/java/com/oceanbase/odc/service/quartz/OdcJobListener.java
```diff
@@ -165,14 +165,18 @@ public void jobWasExecuted(JobExecutionContext context, JobExecutionException jo
...
-            if (jobException != null && notificationProperties.isEnabled()) {
-                try {
+        if (notificationProperties.isEnabled() && scheduleEntityOptional.isPresent()) {
+            try {
+                ScheduleEntity scheduleEntity = scheduleEntityOptional.get();
+                if (jobException != null) {
                     broker.enqueueEvent(eventBuilder.ofFailedTask(scheduleEntity));
-                } catch (Exception e) {
-                    log.warn("Failed to enqueue event.", e);
+                } else if (!taskFrameworkEnabledProperties.isEnabled()
```
MarkPotato777, line 173: Why should the notification module need to be aware of the task framework?

LuckyPickleZZ: This is not in notification module. It's in schedule module, to determine whether to call the notification module.
The notification module is still not aware of task framework.

server/odc-service/src/main/java/com/oceanbase/odc/service/notification/helper/EventBuilder.java
LuckyPickleZZ: No, this is additional code, just added to this PR by the way~

MarkPotato777: Why do we need this label?

MarkPotato777 (approved):
LGTM

yhilmare (approved):
lgtm

PR #2467, opened 2024-05-22
fix(metadb): change systemConfigDao to systemConfigRepository.

ungreat:
#### What type of PR is this?

The field ConfigEntity.key is a reserved word in MySQL, so the workaround employed is using @Column("\`key\`") to avoid conflicts. However, due to differences in how H2 and MySQL handle reserved keywords, the original method is still implemented using jdbcTemplate to maintain compatibility with integration tests.

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/metadb/config/ConfigEntity.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/metadb/config/ConfigEntity.java
@@ -17,6 +17,11 @@ package com.oceanbase.odc.metadb.config;
 
 import java.util.Date;
 
+import javax.persistence.Column;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.MappedSuperclass;
 import javax.validation.constraints.NotBlank;
 
 import lombok.Data;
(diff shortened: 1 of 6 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/metadb/config/ConfigEntity.java
```diff
@@ -27,26 +32,35 @@
...
 public class ConfigEntity {
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private Long id;
     /**
      * configuration key
      */
     @NotBlank
+    @Column(name = "`key`")
```
yizhouxw, line 45: bad practive for identify quote char here,
can we set spring.jpa.properties.hibernate.dialect in application.yml to workaround ?

ungreat: This is used to bypass mysql reserved words.
It's true that hibernate have these configuration, such as `globally_quoted_identifier `, but the impact of these changes is far too broad.Custom implementations may be required to adapt to MySQL and H2 compatibility, and this could also lead to issues such as case sensitivity.

yizhouxw (approved):
lgtm

yhilmare (approved):
lgtm

PR #2468, opened 2024-05-22
enhancement(flow): add heartbeat for flow task

### LINKED ISSUE #2476: [Bug]: ticket's status is illegal when odc node is crashed
### ODC version

4.2.4

### OB version

not related

### What happened?

1. startup a database change task which contains a long time costing sql statement:
截屏2024-05-22 21 49 05

2. kill odc app when the task is running:
截屏2024-05-22 21 51 32

截屏2024-05-22 21 52 05

3. restart odc and the target ticket will ket 'RUNNING' without any changes forever:

the task is ended when the odc app is crashed and we need to set this ticket 'CANCELED' to tell the user that this ticket is not running anymore.

follow 'what happened'

_No response_

yhilmare:
#### What type of PR is this?

type-enhancement

ticket's status will not be changed when the odc node is crashed and there are flow task instance running on it and the user will find that the ticket will keep 'RUNNING' status without any changes.

I use heartbeat to solve this problem, the ticket will be cancelled if the related heartbeat is timeout

Fixes #2476

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/metadb/task/TaskRepository.java
```diff
@@ -43,6 +44,13 @@ public interface TaskRepository extends JpaRepository<TaskEntity, Long>, JpaSpec
     @Modifying
     int update(@Param("param") TaskEntity entity);
 
+    @Transactional
```
yizhouxw, line 47: we can only update heatbeat time while task is running

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/model/FlowTaskProperties.java
```diff
@@ -47,4 +47,8 @@ public class FlowTaskProperties {
 
     @Value("${odc.task.async.index-change-max-timeout-millis:432000000}")
     private long indexChangeMaxTimeoutMillisecond;
+
+    @Value("${odc.flow.task.heartbeat-timeout-seconds:15}")
+    private long heartbeatTimeoutSeconds;
```
yizhouxw, line 52: too short timeout interval, may 180s+

yizhouxw (commented):
I didn't see the initial value of lastHeartbeatTime while task started, what is the behavior if lastHeatbeatTime is null ?

yhilmare:
> I didn't see the initial value of lastHeartbeatTime while task started, what is the behavior if lastHeatbeatTime is null ?

added, we will update heartbeat before task instance's status is updated to 'EXECUTING'
```java
try {
                flowTaskInstance = getFlowTaskInstance(flowInstanceId, activityId);
                BaseRuntimeFlowableDelegate<?> delegate = getDelegateInstance(flowTaskInstance);
                // we init heartbeat time here
                delegate.updateHeartbeatTime();
                List<Class<? extends ExecutionListener>> list = delegate.getExecutionListenerClasses();
                if (CollectionUtils.isNotEmpty(list)) {
                    // we update flow task instance's status here, 'CREATED' -> 'EXECUTING'
                    list.forEach(c -> doCallListener(FlowConstants.EXECUTION_START_EVENT_NAME, executionFacade, c));
                }
                delegate.execute(executionFacade);
                ...
            } catch (Exception e) {
                ...
            } finally {
                ...
            }
```

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/model/FlowTaskProperties.java
yhilmare: changed

yhilmare:
I tested the performance with this data size:
截屏2024-05-23 15 00 00

|SQL|duration|Return rows|
|:----:|:----:|:----:|
|`select * from flow_instance_node_task where status='EXECUTING'`|746.16ms|10w|
|`select * from task_task where id in (?) and last_heartbeat_time < ?`|703ms|5233|
|`select * from flow_instance_node_task where status='EXECUTING' and id in (?) for update`|90.89ms|2000|
|`update flow_instance_node_task set status='CANCELLED' where id in (?)`|51.85ms|2000|
|`update flow_instance set status='CANCELLED' where id in (?)`|9.67ms|130|

yizhouxw (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowSchedules.java
```diff
@@ -0,0 +1,134 @@
...
+    @Autowired
+    private ServiceTaskInstanceRepository serviceTaskInstanceRepository;
+
+    @Scheduled(fixedDelayString = "${odc.flow.task.heartbeat-timeout-check-interval-millis:15000}")
+    public void cancelHeartbeatTimeoutFlow() {
+        long timeoutSeconds = this.flowTaskProperties.getHeartbeatTimeoutSeconds();
+        if (timeoutSeconds <= 0) {
+            return;
+        }
+        long minTimeoutSeconds = RuntimeTaskConstants.DEFAULT_TASK_CHECK_INTERVAL_SECONDS * 3;
```
MarkPotato777, line 72: What's the interval of updating the heartbeat? Should we keep the updating interval less than the heartbeat timeout?

yhilmare: yes, you are right.
the updating interval is set to `RuntimeTaskConstants.DEFAULT_TASK_CHECK_INTERVAL_SECONDS` and we set the min heartbeat timeout is more than 3 times of it

smallsheeeep (approved):
LGTM

MarkPotato777 (approved):
LGTM

krihy (approved):
LGTM

PR #2485, opened 2024-05-23
feat(config):add creator_id to config entity

ungreat:

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/metadb/config/SystemConfigEntity.java
```diff
@@ -32,6 +33,12 @@
 @Table(name = "config_system_configuration")
 public class SystemConfigEntity extends ConfigEntity {
 
+    @Column(name = "creator_id")
+    private String creatorId;
```
yhilmare, line 37: the column data type of `creator_id` and `last_modifier_id` is bigint

yizhouxw (approved):
lgtm

PR #2489, opened 2024-05-23
fix(dlm): sync table structure failed

### LINKED ISSUE #2482: [Bug]: Oracle creation archive task synchronization table structure failed
### ODC version

ODC430

### OB version

independent

### What happened?

Oracle creation archive task synchronization table structure failed
image

image

。

1. The oracle table is created on the source side but not on the target side.
2. Create an archiving task and uncheck the synchronization table structure
```
CREATE TABLE "NON_PARTRANGE_RANGE_PRI" (
  "COL1" NUMBER(*,0) NOT NULL ENABLE,
  "COL2" NUMBER,
  "COL3" CHAR(10),
  "COL4" VARCHAR2(10),
  "COL7" DATE,
  "COL8" TIMESTAMP(6),
  CONSTRAINT "NON_PARTRANGE_RANGE_PRI_PK" PRIMARY KEY ("COL1")
) COMPRESS FOR ARCHIVE REPLICA_NUM = 1 BLOCK_SIZE = 16384 USE_BLOOM_FILTER = FALSE TABLET_SIZE = 134217728 PCTFREE = 0
 partition by range("COL1")
(partition "P1" values less than (0),
partition "P2" values less than (10000),
partition "P3" values less than (100000),
partition "P4" values less than (MAXVALUE));
```

_No response_

guowl3:
#### What type of PR is this?
type-bug

Fixed the issue with structure synchronization failure.

Fixes #2482

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/dlm/DLMTableStructureSynchronizer.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/dlm/DLMTableStructureSynchronizer.java
@@ -17,9 +17,11 @@ package com.oceanbase.odc.service.dlm;
 
 import java.sql.Connection;
 import java.sql.SQLException;
+import java.sql.Statement;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/dlm/DLMTableStructureSynchronizer.java
```diff
@@ -67,25 +69,35 @@ public static void sync(ConnectionConfig sourceConnectionConfig, ConnectionConfi
...
-            for (DBObjectComparisonResult result : results) {
-                String changeSqlScript =
-                        result.toEntity(1L, targetConnectionConfig.getDialectType()).getChangeSqlScript();
-                if (StringUtils.isNotEmpty(changeSqlScript)) {
-                    log.info("Start to sync target table structure,sqls={}", changeSqlScript);
-                    conn.prepareStatement(changeSqlScript).execute();
-                    log.info("Sync table structure success.");
-                } else {
-                    log.info("Table structure comparison has finished,no action is necessary.");
+        DBObjectComparisonResult result = comparator.compare(sourceConfig, targetConfig).get(0);
```
smallsheeeep, line 72: Will there occurs that the compare result has both `ONLY_IN_SOURCE` and `INCONSISTENT`?

guowl3: No, the INCONSISTENT enum value indicates that it exists, but is not consistent.

smallsheeeep (approved):
LGTM

Issue #2460, opened 2024-05-22
[Bug]: DLM task, source and target table names are inconsistent, table structure synchronization fails


Huangxiao-mas:
### ODC version

ODC430

### OB version

Oceanbase430

### What happened?

DLM task, source and target table names are inconsistent, table structure synchronization fails

### What did you expect to happen?

.

### How can we reproduce it (as minimally and precisely as possible)?

1. Create a table a at the source end. The target creates a b table. The source table has several more fields than the target table.
2. Create an archiving task and select table structure synchronization

Huangxiao-mas:
test pass

PR #2497, opened 2024-05-25
fix(dlm): table structure synchronization failure when table names are inconsistent

guowl3:
#### What type of PR is this?
type-bug

The structure comparison module has added a target table name field, which is used to generate change scripts for tables with different names. After this PR is merged, the DLM module will support table structure synchronization for tables with different names.

Fixes #[2460](https://github.com/oceanbase/odc/issues/2460)

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/structurecompare/comparedbobject/AbstractDBObjectStructureComparator.java
```diff
@@ -32,10 +32,15 @@ public abstract class AbstractDBObjectStructureComparator<T extends DBObject>
         implements DBObjectStructureComparator<T> {
     protected final String srcSchemaName;
     protected final String tgtSchemaName;
+    protected final String srcTableName;
+    protected final String tgtTableName;
```
PeachThinking, line 36: It is recommended to change these 2 fields to：
`projceted final List<String> srcObjectNames;`
`projceted final List<String> tgtObjectNames;`

PeachThinking:
👍👍👍～
could you please add some compare tables with different name in ut test like `DefaultDBStructureComparatorTest`

guowl3:
> 👍👍👍～ could you please add some compare tables with different name in ut test like `DefaultDBStructureComparatorTest`
added unit test 👉 DBTableStructureComparatorTest.java

PeachThinking (approved):
lgtm

smallsheeeep (approved):

yizhouxw (approved):

yhilmare (approved):
lgtm

PR #2511, opened 2024-05-27
fix(multiple databases):database changing order in template cannot be edited 

### LINKED ISSUE #2508: [Bug]: The database in the multi-database change template cannot be edited.
### ODC version

ODC 4.3.0

### OB version

No related.

### What happened?

The database in the multi-database change template cannot be edited.If I want to add a database to the template, I need to create a new template, which is inconvenient to use.

Support for editing templates.

Follow what happened.

_No response_

zijiacj:
#### What type of PR is this?
type-bug

Currently, only the name of the template can be modified. Therefore, the function of modifying the database changing order is added.
Also fixed the bug that the execution record was not displayed after the first batch of execution of the multiple databases ticket

Fixes #2508 #2516

```docs

```

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,5 +1,6 @@
 <component name="ProjectCodeStyleConfiguration">
   <code_scheme name="Project" version="173">
+    <option name="LINE_SEPARATOR" value="&#10;" />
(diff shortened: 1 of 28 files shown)


yhilmare (approved):
lgtm

PR #2520, opened 2024-05-28
fix(web): cannot return a page with more than 2000 records.

### LINKED ISSUE #2446: [Bug]: There are more than 2000 databases that cannot be switched in the sql window
### ODC version

ODC 424-bp2

### OB version

NO related

### What happened?

There are more than 2000 databases that cannot be switched in the sql window.

More than 2000 databases can also switch databases in the sql window.

When there are 2000 databases, switch databases in the sql window.

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug

Spring sets a default max page size of 2000 which means that if the front-end requests for a more than 2000 page size, the backend will return only 2000 records in one page.
This PR adjusts the default value 2000 to 99999.

Fixes #2446 

```docs

```


yhilmare (approved):
lgtm

yizhouxw (approved):
lgtm

PR #2523, opened 2024-05-28
fix(database): database sync involved no-privilege databases in OBMySQL

MarkPotato777:
#### What type of PR is this?

type-bug

`information_schema.schemeta` may return databases that the current database user cannot access, which is an issue of OceanBase MySQL mode.

This PR uses `show databases` to filter the querying result of `information_schema.schemeta`.

Fixes #

```docs

```


yizhouxw (approved):
lgtm

yhilmare (approved):
lgtm

Issue #2524, opened 2024-05-28
[Bug]: exception will be thrown when connection is reset


yhilmare:
### ODC version

4.2.4

### OB version

not related

### What happened?

1. connect to a mysql database by sql-console and obclient
2. query connection id by sql-console and kill this connection by this connection id in obclient:
<img width="483" alt="截屏2024-05-28 14 59 12" src="https://github.com/oceanbase/odc/assets/22992296/491970c6-c7c7-4a3f-9b92-79a2fb34244b">
<img width="409" alt="截屏2024-05-28 14 59 35" src="https://github.com/oceanbase/odc/assets/22992296/1cb3bb05-ca84-40ad-a07f-54d36bc29388">

3. query any sql in sql-console and an exception will be thrown:
<img width="1121" alt="截屏2024-05-28 15 00 19" src="https://github.com/oceanbase/odc/assets/22992296/a347ee15-293d-4eb7-8c9e-cd108c6ded72">

### What did you expect to happen?

I kill the connection in obclient which means that the connection is invalid, I expect that I can auto reconnect it but not throw an exception

### How can we reproduce it (as minimally and precisely as possible)?

follow 'what happened'

yhilmare:
this exception is involved by this pr #2307

the lock will be update when the connection is reset but we expect this lock as the old lock

sl01388797:
![Image](https://github.com/oceanbase/odc/assets/143492234/f69ebf06-9aa0-4118-9907-39659a15259d)

pass

PR #2528, opened 2024-05-28
fix(connection): concurrent exception will be thrown when a connection is reset

### LINKED ISSUE #2525: [Bug]: error message appear in odc.log
### ODC version

4.2.4

### OB version

not related

### What happened?

startup odc server and wait for a minutes(wait for database sync)

grep odc.log

截屏2024-05-28 15 18 18

there is not error log printing

follow 'what happened'

_No response_

yhilmare:
#### What type of PR is this?

type-bug

the lock attached to the connection will be updated when the connection is reset. the thread which hold the old lock should not attempt to unlock the new lock.

Fixes #2525 #2524

```docs

```


yizhouxw (dismissed):
lgtm

MarkPotato777 (dismissed):
LGTM

krihy (commented):
lgtm

krihy (approved):
lgtm

PR #2533, opened 2024-05-28
fix(dlm):if data cleaning target table name is null, set default value is source table name

### LINKED ISSUE #2507: [Bug]: Data cleaning, data cleaning is linked to the historical database data verification report null pointer
### ODC version

ODC430

### OB version

independent

### What happened?

Data cleaning, data cleaning is linked to the historical database data verification report null pointer
image
image

。

Data cleaning: If the target data source table does not exist, the null pointer will be created when it is created.

_No response_

kiko-art:
#### What type of PR is this?
type-bug

In data cleaning , the target table name has been modified to be an optional field. If the target table name is  blank, the source table name will be used as the default.

Fixes #2507 

```docs

```


guowl3 (approved):
LGTM

PR #2536, opened 2024-05-28
fix(multiple database): no execution record is generated before or during a multi-database change task 

### LINKED ISSUE #2516: [Bug]: No execution record is generated before or during a multi-database change task
### ODC version

ODC 4.3.0

### OB version

No related

### What happened?

No execution record is generated before or during a multi-database change task.
image
image

The execution record of a multi-library change task must be generated before and during the task execution.

Deliver a multi-database change task and view the task record before and during the execution.

_No response_

zijiacj:
…played in real time

#### What type of PR is this?
type-bug

1. Previously, the multi-database change ticket would have  execution records only after finishing the first batch. Now it's changed to that there are records of execution displaying once the multi-database change ticket is created.
2. Added the execution progress display.

Fixes #2516

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/task/TaskService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/task/TaskService.java
@@ -288,11 +288,30 @@ public class TaskService {
     }
 
     @Transactional(rollbackFor = Exception.class)
+    public void succeed(Long id) {
+        TaskEntity taskEntity = nullSafeFindById(id);
+        taskEntity.setStatus(TaskStatus.DONE);
+        taskEntity.setProgressPercentage(100);
+        taskRepository.save(taskEntity);
+        log.info("Task ended: taskId={}", id);
+    }
+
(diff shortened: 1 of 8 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/task/TaskService.java
```diff
@@ -287,6 +287,15 @@ public void succeed(Long id, Object taskResult) {
         log.info("Task ended: taskId={}", id);
     }
 
+    @Transactional(rollbackFor = Exception.class)
```
yizhouxw, line 290: we already have ` public void succeed(Long id, Object taskResult)` , seems not necessary add a method with less responsibility. may handle `null` taskResult inside current `succeed` method.

so as method ` public void succeed(Long id, Object taskResult)`.

zijiacj: ok

PR #2538, opened 2024-05-28
refactor(flow): add skip auth to flow mapper

ungreat:

Fixes #

```docs

```


yizhouxw (commented):
seem too complex to add so much skipAuth parameter,
can we just change authorize logic instead ?

ungreat:
> seem too complex to add so much skipAuth parameter, can we just change authorize logic instead ?

Currently, the method to bypass authentication is by passing parameters, which is indeed too complicated. In the future, it could be simplified by adopting a technical change to uniformly switch to using ThreadLocal.



> seem too complex to add so much skipAuth parameter, can we just change authorize logic instead ?

It‘s  true that most method to skip auth is by passing parameters at present, may need a technical refactor issue to deal with it uniformly.

yizhouxw (approved):

yizhouxw (approved):

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/model/FlowInstanceDetailResp.java
```diff
@@ -96,6 +97,7 @@ public class FlowInstanceDetailResp {
     private Date completeTime;
     private List<FlowNodeInstanceDetailResp> nodeList;
 
+    @Getter
```
yhilmare, line 100: this getter seems useless?

yhilmare (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/model/FlowInstanceDetailResp.java
yhilmare: you may remove it in another pr~

PR #2584, opened 2024-05-30
fix(multiple database): The current database does not match the corresponding sql check result

### LINKED ISSUE #2470: [Bug]: The sql check result of multi-database change is missing
### ODC version

ODC430

### OB version

No related

### What happened?

The sql check ‘the initial value of the auto-increment column’ is not matched when the ‘create  table gsh.ttt(a int)’ statement is executed
image
Executing in the sql window matches the sql check
image

Multi-database changes accurately match sqlcheck

Follow what happened.

_No response_

zijiacj:
#### What type of PR is this?
type-bug

make each database matches its own sql check result

Fixes #2470

```docs

```


yhilmare (approved):
lgtm

PR #2585, opened 2024-05-31
fix(multiple database): add project permission verification to the exist method

zijiacj:
#### What type of PR is this?
type-bug

Absent project permission check in exist method, users who do not belong to the current project can also get the response result. Add project verification to avoid this situation

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #2587, opened 2024-05-31
fix(data-transfer): exception occurs when object exists  and configured continue when error

### LINKED ISSUE #2555: [Bug]: After the column storage table is exported, the column storage definition statement is missing in the schema file
### ODC version

ODC430

### OB version

Oceanbase430

### What happened?

After the column storage table is exported, the column storage definition statement is missing in the schema file
image
image

。

```
CREATE TABLE tt_column_store (c1 int PRIMARY KEY, c2 int , c3 int) WITH COLUMN GROUP (each column);
```
export it

_No response_

### LINKED ISSUE #2583: [Bug]: data loading would not be executed when schema was imported failed
### ODC version

4.2.4

### OB version

-

### What happened?

When schema was imported failed, the data task would not be executed even user configured continue on error.

The data task would be continued if schema task failed.

-

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

When a imported object exists , obloader would report error even if user configured continue when error. So we should eat this exception in this case. To fix #2583 .

And obdumper could only dump table and index with column group configuration by `show create table`. It's configured by the `compactSchema` option. To fix #2555 

Fixes #

```docs

```


yhilmare (approved):
lgtm

Issue #2548, opened 2024-05-29
[Bug]: The user is deactivated. The user is a library administrator of a project. At this time, the user can still approve work orders.


qymsummer:
### ODC version

ODC 4.3.0

### OB version

No related

### What happened?

The user is deactivated. The user is a library administrator of a project. At this time, the user can still approve work orders.
![image](https://github.com/oceanbase/odc/assets/40919496/65f119e1-f629-4574-b126-bbccddbcfd1b)
![截屏2024-05-29 13 46 23](https://github.com/oceanbase/odc/assets/40919496/119101a1-9fb2-486e-81df-2ac40c54f72f)

### What did you expect to happen?

The user is deactivated and cannot approve work orders

### How can we reproduce it (as minimally and precisely as possible)?

First, create a work order that requires approval from the library administrator. At this time, change the library administrator account to deactivated. Then, in the library administrator approval work order menu, click the approved work order and click OK.

smallsheeeep:
The root cause is that ODC cannot force the disabled user to logout. It is hard to fix it completely, I just make the disabled user cannot approve the flow.

qymsummer:
![Image](https://github.com/oceanbase/odc/assets/40919496/27e89b64-9166-400e-ac68-88a25d3e6864)

pass

PR #2589, opened 2024-05-31
fix(flow): the disabled user can still approving or rejecting a flow

### LINKED ISSUE #2550: [Bug]: A deactivated user should not be able to select
### ODC version

ODC 4.3.0

### OB version

not related

### What happened?

A deactivated user should not be able to select
image

A deactivated user should not be able to select

First deactivate a user, then set the library administrator, and the deactivated user appears in the list

_No response_

smallsheeeep:
#### What type of PR is this?
type-bug
module-flow

For a disabled user, there are some issues:
1. The user can still be selected as the database owner;
2. In ticket list, the user still displaied as the flow approval candidates;
3. The user can still approve the flow.
This PR fixed it.

Fixes #2550 #2548 

Self-test passed.

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/collaboration/project/model/Project.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/collaboration/project/model/Project.java
@@ -123,6 +123,9 @@ public class Project implements SecurityResource, OrganizationIsolated, Serializ
         private String accountName;
 
         @JsonProperty(access = Access.READ_ONLY)
+        private Boolean userEnabled;
+
+        @JsonProperty(access = Access.READ_ONLY)
         private String name;
 
         private ResourceRoleName role;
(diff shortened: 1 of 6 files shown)


yhilmare (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/collaboration/project/model/Project.java
```diff
@@ -122,6 +122,9 @@ public static class ProjectMember {
         @JsonProperty(access = Access.READ_ONLY)
         private String accountName;
 
+        @JsonProperty(access = Access.READ_ONLY)
+        private Boolean userEnabled;
```
MarkPotato777, line 126: It's a bit odd that the project has this `userEnabled` field. How should we understand this?

MarkPotato777: Oh sorry, I didn't see it was added to `ProjectMember` rather than `Project`. Then it's cool to me~

MarkPotato777 (approved):
LGTM

krihy (commented):
Lgtm

PR #2593, opened 2024-05-31
fix(multiple database): optimize error message when creating and updating template

### LINKED ISSUE #2541: [Bug]: Optimize error messages when the database is empty during template creation
### ODC version

ODC4.3.0

### OB version

No related.

### What happened?

Optimize error messages when the database is empty during template creation.
image

Error messages should be optimized，such as：The number of databases must be greater than 1 and not more than 100.

Do not select the database when creating the template and save it.

_No response_

zijiacj:
#### What type of PR is this?
type-bug

Optimized error messages when creating and modifying templates. It is best that the front end restricts the case where the database is empty and cannot initiate creation and modification

Fixes #2541

```docs

```


yhilmare (approved):
lgtm

PR #2636, opened 2024-06-04
fix(schema): function and procedure list is not ordered by their name in ob-mysql

yhilmare:
#### What type of PR is this?

type-bug

this pr fix an issue that the procedure and function list os not ordered by their name

Fixes #

```docs

```


PeachThinking (approved):
lgtm

PR #2637, opened 2024-06-04
fix(dlm): table not found in task framework mode

guowl3:
#### What type of PR is this?
type-bug

In task framework mode, the task throws a 'table not found' exception because the 'dlm_job' was renamed to 'dlm_table_unit'.

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #2638, opened 2024-06-04
fix(structure-comparison): syntax error when executing structure comparison

### LINKED ISSUE #2491: [Bug]: Syntax error after structure comparison, ‘PARTITIONS 5’ was added in the structure change
### ODC version

ODC430

### OB version

Oceanbase421 oracle

### What happened?

Syntax error after structure comparison, ‘PARTITIONS 5’ was added in the structure change

。

before comparing
```
CREATE TABLE "USER_RO"."NON_PARTRANGE_RANGE_PRI2" (
  "COL1" NUMBER(*,0),
  "COL2" NUMBER,
  "COL3" CHAR(10),
  "COL4" VARCHAR2(10),
  "COL8" TIMESTAMP(6),
  PRIMARY KEY ("COL1")
) COMPRESS FOR ARCHIVE REPLICA_NUM = 1 BLOCK_SIZE = 16384 USE_BLOOM_FILTER = FALSE TABLET_SIZE = 134217728 PCTFREE = 0
 partition by range(COL1)
(partition "P1" values less than (500),
partition "P2" values less than (1000),
partition "P3" values less than (1500),
partition "P4" values less than (2000),
partition "P5" values less than (MAXVALUE));
```
after 
```
CREATE TABLE "YUC"."NON_PARTRANGE_RANGE_PRI2" (
"COL1" NUMBER(*, 0) NOT NULL DEFAULT NULL,
"COL2" NUMBER NULL  DEFAULT NULL,
"COL3" CHAR(10) NULL  DEFAULT NULL,
"COL4" VARCHAR2(10) NULL  DEFAULT NULL,
"COL8" TIMESTAMP(6)  NULL  DEFAULT NULL,
CONSTRAINT "NON_PARTRANGE_RANGE_PRI2_OBPK_1716531531720584" PRIMARY KEY ("COL1")
)  PARTITION BY RANGE(COL1) 
PARTITIONS 5(
PARTITION "P1" VALUES LESS THAN (500),
PARTITION "P2" VALUES LESS THAN (1000),
PARTITION "P3" VALUES LESS THAN (1500),
PARTITION "P4" VALUES LESS THAN (2000),
PARTITION "P5" VALUES LESS THAN (MAXVALUE)
);
```

_No response_

yhilmare:
#### What type of PR is this?

type-bug

the table editor append `partition 5` into the partition defs which is wrong, this pr fix this problem~

Fixes #2491

```docs

```

### DIFF
--- a/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/editor/DBTablePartitionEditor.java
+++ b/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/editor/DBTablePartitionEditor.java
@@ -81,6 +81,7 @@ public abstract class DBTablePartitionEditor implements DBObjectEditor<DBTablePa
             if (Objects.nonNull(partition.getPartitionOption().getPartitionsNum())) {
                 sqlBuilder.append("\nPARTITIONS ")
                         .append(String.valueOf(partition.getPartitionOption().getPartitionsNum()));
+                return sqlBuilder.toString();
             }
         }
         appendDefinitions(partition, sqlBuilder);
--- a/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/editor/DBTablePartitionEditor.java
+++ b/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/editor/DBTablePartitionEditor.java
@@ -76,13 +76,9 @@ public abstract class DBTablePartitionEditor implements DBObjectEditor<DBTablePa
         if (Objects.nonNull(partition.getSubpartitionTemplated()) && partition.getSubpartitionTemplated()) {
             generateCreateTemplateSubPartition(partition, sqlBuilder);
         }
-        if (Objects.nonNull(partition.getPartitionOption().getPartitionsNum())) {
-            sqlBuilder.append("\nPARTITIONS ")
-                    .append(String.valueOf(partition.getPartitionOption().getPartitionsNum()));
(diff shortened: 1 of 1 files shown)


libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/editor/DBTablePartitionEditor.java
```diff
@@ -76,13 +76,9 @@ public String generateCreateDefinitionDDL(@NotNull DBTablePartition partition) {
...
-        if (Objects.nonNull(partition.getPartitionOption().getPartitionsNum())) {
-            sqlBuilder.append("\nPARTITIONS ")
-                    .append(String.valueOf(partition.getPartitionOption().getPartitionsNum()));
-            if (partition.getPartitionOption().getType() == DBTablePartitionType.HASH
-                    || partition.getPartitionOption().getType() == DBTablePartitionType.KEY) {
-                return sqlBuilder.toString();
-            }
+        if (partition.getPartitionOption().getType() == DBTablePartitionType.HASH
+                || partition.getPartitionOption().getType() == DBTablePartitionType.KEY) {
+            return sqlBuilder.toString();
```
PeachThinking, line 81: If we return directly here, then there will be no partition definition information for hash and key types?

PeachThinking (dismissed):
lgtm

PeachThinking (approved):

yizhouxw (approved):

PR #2639, opened 2024-06-04
fix(multiple database): error occurs when viewing the list without templates

### LINKED ISSUE #2615: [Bug]: When selecting a template for a multi-database change task without a template, an error will be reported
### ODC version

ODC 4.3.0

### OB version

no related

### What happened?

Create a new multi-database change, select a project, select a database (a database without a template) in the execution node, click Select Template, and an error will be reported.
image
image

When the data source has no template, clicking Select Template should not result in an error

Create a new multi-database change, select a project, select a database (a database without a template) in the execution node, click Select Template, and an error will be reported.

_No response_

zijiacj:
#### What type of PR is this?
type-bug

Returns an empty page object when the template in the project is empty. Avoid executing the following logic

Fixes #2615

```docs

```


yhilmare (approved):
lgtm

PR #2661, opened 2024-06-05
fix(db-object): exception occurred when open oracle table in GBK encoding

### LINKED ISSUE #2262: [Bug]: The native Oracle character set is ZHS16GBK, and ODC cannot open the table structure.
### ODC version

ODC424bp1

### OB version

11g

### What happened?

The native Oracle character set is ZHS16GBK, and ODC cannot open the table structure.
image

。

。

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

When Oracle character set is ZHS16GBK, ODC cannot open the table structure.
According to https://stackoverflow.com/questions/36775398/java-sql-sqlexception-non-supported-character-set-oracle-character-set-178 , it's because we lack i18n dependency.
Add this dependency should fix this.

Fixes #2262

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1128,6 +1128,11 @@
                 <version>${oracle.jdbc.version}</version>
             </dependency>
             <dependency>
+                <groupId>com.oracle.database.nls</groupId>
+                <artifactId>orai18n</artifactId>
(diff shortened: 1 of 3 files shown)


yhilmare (approved):
lgtm

PR #2677, opened 2024-06-06
fix(multiple database): the method isIntercepted in multiple database pre check node is error

zijiacj:
#### What type of PR is this?
type-bug

The previous isIntercepted method logic was incorrect. In the pre-check node, the judgment on the violation level of the multiple database result is skipped

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #2679, opened 2024-06-06
fix(data-transfer): truncate will cause the import task to fail

### LINKED ISSUE #2588: [Bug]: Choosing to clear data before importing CSV format files in native Oracle is invalid
### ODC version

odc430

### OB version

*

### What happened?

image
image
image

As long as the imported data is included in the table data

Select the CSV format file for importing native Oracle, and choose to clear data before importing

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

When importing data for native oracle, we use `preSql` option to set `truncate table` statement .
But DataX would not execute `session` options before executing preSqls. And we rely on `session` options to switch the schema. So the truncate would failed.

Fixes #2588


yhilmare (approved):
lgtm

PR #2694, opened 2024-06-06
fix(multiple database):  the return of method intercepted in multiple database pre check node is incorrect

zijiacj:
#### What type of PR is this?
type-bug

The current return logic will directly skip the subsequent permission verification, resulting in users who do not have the permission to change the database can also initiate 

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #2698, opened 2024-06-06
fix(desktop): failed to bootstrap when upgrading from 3.2.0

yhilmare:
#### What type of PR is this?

type-bug

this pr fix the issue that odc can not bootstrap from 3.3.0

Fixes #

```docs

```


yizhouxw (approved):
lgtm

PR #2702, opened 2024-06-06
fix(multiple database): the return of method intercepted in multiple database pre check node is incorrect 

### LINKED ISSUE #2696: [Bug]: The pre-check does not fail when a multi-database change task is delivered for a database that has no change permission
### ODC version

ODC 430

### OB version

No related.

### What happened?

Execute the create statement on a database that does not have permission to change, and the pre-check does not fail.
image

For database without permissions, the precheck should fail.
image

The user does not have the permission to change the chz database. In the multi-database change, execute “create table chz.table(a int);”

_No response_

zijiacj:
#### What type of PR is this?
type-bug

The current return logic will directly skip the subsequent permission verification, resulting in users who do not have the permission to change the database can also initiate 

Fixes #2696

```docs

```


yhilmare (approved):
lgtm

PR #2712, opened 2024-06-07
fix(dlm): the task status does not update properly when structural synchronization fails

### LINKED ISSUE #2703: [Bug]: After archiving tables with different structures, the task is canceled and the log reports a null pointer
### ODC version

ODC430

### OB version

Oceanbase420 ORACLE

### What happened?

After archiving tables with different structures, the task is canceled and the log reports a null pointer
image

.

taskId 330
chat priavtely

_No response_

guowl3:
#### What type of PR is this?
type-bug

Fixes #2703 

```docs

```


yhilmare (approved):
lgtm

smallsheeeep (approved):
LGTM

PR #2716, opened 2024-06-07
fix(ticket): failed to view all tickets

### LINKED ISSUE #2480: [Bug]: The amount of data for the paged query of the tickets is not fully displayed
### ODC version

424

### OB version

uncorrelated

### What happened?

In odc 424, even if the user is admin, when you click to view all lists、export、export result sets、import、mock data、async、shadow_async、 structure comparison、partition plan、data cleaning、and apply for database permission, the result of the back-end response is only one page of data. But in fact ,the number of records in the database is  greater than that. As shown in the following picture
image

Expect to return the correct totalElements and totalPages in the paged query. For example, clicking on 'I initiated' is currently getting the correct response. As shown in the following picture
image

click to view all lists、export、export result sets、import、mock data、async、shadow_async、 structure comparison、partition plan、data cleaning、or apply for database permission.using one of the above operations you can reproduce it.
image

_No response_

yhilmare:
#### What type of PR is this?

type-bug

the root cause of this problem is that we add `distinct` at the query and we will get the query like this when we attempt to `count(1)`:

```sql
select distinct count(a) from xxx group by flow_instance_id, task_type
```

almost every group's size is equal to 1 so the distinct operation will lead to wrong result~

Fixes #2480

```docs

```


yizhouxw (approved):
lgtm

PR #2719, opened 2024-06-11
fix(multiple database): change the method when initiating child tickets

zijiacj:
#### What type of PR is this?
type-bug

Permissions have been checked when creating multi-database change. Initiating a child work order belongs to the execution process of the multi-database and should not include permission verification. Change it to the same as the sql plan initiating child tickets.

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #2740, opened 2024-06-13
enhancement(mulitple database): add cancel for multiple databases change 

zijiacj:

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
@@ -582,6 +582,13 @@ public class FlowInstanceService {
         return cancel(flowInstance, skipAuth);
     }
 
+    // avoid throwing UnexpectedRollbackException in multi-database cancellation
+    @Transactional(rollbackFor = Exception.class, propagation = Propagation.REQUIRES_NEW)
+    public void cancelSubFlowInstance(@NotNull Long id) {
+        this.cancel(id, false);
+
+    }
+
     public FlowInstanceDetailResp cancelNotCheckPermission(@NotNull Long id) {
         FlowInstance flowInstance = mapFlowInstance(id, flowInst -> flowInst, false);
         return cancel(flowInstance, false);
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/MultipleDatabaseChangeRuntimeFlowableTask.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/MultipleDatabaseChangeRuntimeFlowableTask.java
@@ -20,6 +20,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
 
+import javax.persistence.EntityManager;
+
 import org.flowable.engine.delegate.DelegateExecution;
 import org.flowable.engine.delegate.ExecutionListener;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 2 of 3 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
```diff
@@ -582,6 +582,13 @@ public FlowInstanceDetailResp cancel(@NotNull Long id, Boolean skipAuth) {
         return cancel(flowInstance, skipAuth);
     }
 
+    // avoid throwing UnexpectedRollbackException in multi-database cancellation
+    @Transactional(rollbackFor = Exception.class, propagation = Propagation.REQUIRES_NEW)
```
yizhouxw, line 586: what if call method 'cancelSubFlowInstance' for parent flow instance ?

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/MultipleDatabaseChangeRuntimeFlowableTask.java
```diff
@@ -325,4 +380,29 @@ public List<Class<? extends ExecutionListener>> getExecutionListenerClasses() {
         return null;
     }
 
+    public List<Long> listNeedCancelIds(List<Long> instanceIds) {
+        // Clear the level 1 cache of JPA
+        entityManager.clear();
+        List<FlowInstanceEntity> flowInstanceEntityList = flowInstanceService.listByIds(instanceIds);
```
yizhouxw, line 386: may simplify use 'select id from flow_instance where id in () and status in ()' .

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/MultipleDatabaseChangeRuntimeFlowableTask.java
```diff
@@ -75,16 +81,48 @@ public class MultipleDatabaseChangeRuntimeFlowableTask extends BaseODCFlowTaskDe
...
+                break;
+            }
+            if (isContinuedForCancel) {
+                if (this.flowInstanceIds == null || this.flowInstanceIds.isEmpty()) {
+                    break;
+                }
+                this.flowInstanceIds = listNeedCancelIds(this.flowInstanceIds);
+                for (Long flowInstanceId : this.flowInstanceIds) {
+                    try {
+                        this.flowInstanceService.cancelSubFlowInstance(flowInstanceId);
```
yizhouxw, line 102: it may take long time if there exists many sub flow intances, may involve thread pool.

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/MultipleDatabaseChangeRuntimeFlowableTask.java
zijiacj: ok, it's better to do that when querying the database

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/MultipleDatabaseChangeRuntimeFlowableTask.java
zijiacj: ok

zijiacj:
Currently, the cancellation function of a database change ticket can be used to indirectly achieve the termination of batch database change ticket.  If some customers feel that they need to cancellation of batch database change ticket then we will add it.

PR #2745, opened 2024-06-14
fix(taskframework): cannot rollback stop when destroy executor failed

krihy:
#### What type of PR is this?

module-taskframework

because stop remote task and destroy executor is two phases, stop remote action cannot be rollback if destroy failed, we should do check conditional about executor can be destroy

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/K8sJobCaller.java
```diff
@@ -85,6 +85,32 @@ protected void doDestroy(JobIdentity ji, ExecutorIdentifier ei) throws JobExcept
...
+        }
+        if (k8sJobResponse.isPresent()) {
+            if (PodStatus.PENDING == PodStatus.of(k8sJobResponse.get().getResourceStatus())) {
+                JobConfiguration jobConfiguration = JobConfigurationHolder.getJobConfiguration();
+                JobEntity jobEntity = jobConfiguration.getTaskFrameworkService().find(ji.getId());
+                if ((System.currentTimeMillis() - jobEntity.getStartedTime().getTime()) / 1000 <= podConfig
+                        .getPodPendingTimeoutSeconds()) {
+                    // Pod cannot be deleted when pod pending is not timeout,
+                    // so throw exception representative cannot delete
+                    log.warn("Cannot destroy pod, pending is not timeout, jodId={0}, identifier={1}, podStatus={2}",
```
yizhouxw, line 105: `{0}` may contains no position number ?

krihy: okey, done

yizhouxw (approved):
lgtm

yizhouxw (changes requested):
may change target branch to dev/4.3.0

krihy:
> may change target branch to dev/4.3.0

okey , create a new  [PR](https://github.com/oceanbase/odc/pull/2755), and close this pr

krihy:
see new pr: https://github.com/oceanbase/odc/pull/2745

PR #2755, opened 2024-06-17
fix(taskframework): cannot rollback stop when destroy executor failed

krihy:
#### What type of PR is this?

module-taskframework

because stop remote task and destroy executor is two phases, stop remote action cannot be rollback if destroy failed, we should do check conditional about executor can be destroy

Fixes #

```docs

```


yizhouxw (approved):
lgtm

yhilmare (approved):
lgtm

PR #2761, opened 2024-06-17
fix(client-mode): ODC fail to start in client mode

### LINKED ISSUE #2760: [Bug]: ODC 4.3.0 would fail to start in client mode
### ODC version

4.3.0

### OB version

-

### What happened?

ODC 4.3.0 would fail to start in client mode.
image

Start normally.

Clear ODC completely, including h2 db files. Then install ODC 4.3.0.

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

ODC added `IgnoreResourceIdHandle` for client  mode in 4.3.0 version. This results in a default organization ID of 10000 in the iam_organization table. But due to the organization ID that other tables depend on being hardcode to 1, the correct organization cannot be found.

Fixes #2760


yhilmare (approved):
lgtm

PR #2763, opened 2024-06-18
fix(taskframework): running task be canceled incorrect due to heartbeat timeout 

krihy:
#### What type of PR is this?

type-bug
running task will be canceled due to heartbeat timeout,  executor will report `CANCELED` status to odc server, but task will be retry and running again, so odc server receive a incorrect signal and terminal task, so we ignore `CANCELED` result but receive `FAILED` status

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/BaseTask.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/BaseTask.java
@@ -71,12 +71,21 @@ public abstract class BaseTask<RESULT> implements Task<RESULT> {
 
     @Override
     public boolean stop() {
+        return doStop(JobStatus.CANCELED);
+    }
+
+    @Override
+    public boolean stopAndFailed() {
+        return doStop(JobStatus.FAILED);
+    }
+
+    private boolean doStop(JobStatus status) {
         try {
             if (getStatus().isTerminated()) {
-                log.warn("Task is already finished and cannot be canceled, id={}, status={}.", getJobId(), getStatus());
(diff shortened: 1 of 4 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/Task.java
```diff
@@ -40,6 +41,13 @@ public interface Task<RESULT> {
      */
     boolean stop();
 
+
+    /**
+     * Stop current task and set status to failed. This method will be called by {@link TaskMonitor} for
+     * stop a timeout task
+     */
+    boolean stopAndFailed();
```
yizhouxw, line 49: how about `abort` or `timeout`.

krihy: okey, rename to abort

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/BaseTask.java
```diff
@@ -71,12 +71,21 @@ public void start(JobContext context) {
...
     public boolean stop() {
+        return stop(JobStatus.CANCELED);
+    }
+
+    @Override
+    public boolean stopAndFailed() {
+        return stop(JobStatus.FAILED);
+    }
+
+    private boolean stop(JobStatus status) {
```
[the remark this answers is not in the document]
krihy: add `synchronized` on this method

yizhouxw (approved):
lgtm

PR #2765, opened 2024-06-18
fix(multiple database): frequently printing logs

zijiacj:
#### What type of PR is this?
type-bug

Before to get FlowTaskExecutionStrategy object, using  'flowInstanceService.detail(getFlowInstanceId())' in class MultipleDatabaseChangeRuntimeFlowableTask, which will print a log. The solution is to persist FlowTaskExecutionStrategy objec into MultipleDatabaseChangeParameters

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #2776, opened 2024-06-19
fix(table-object): there would be an NPE if fail to parse index ddl

### LINKED ISSUE #2747: [Bug]: NPE would be reported when viewing table structure
### ODC version

4.2.4

### OB version

ob-oracle 
gbk character set
4.3.1

### What happened?

When viewing table structure of OB 4.3.1, ODC maybe failed to parse its ddl. This would cause NPE. 
And the exception stack was eaten.

Parsing failure should not cause NPE.
Exception stack should be print in log files to find problems.

View a table structure of OB 4.3.1 in gbk character set.

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

If fail to parse index ddl, there is an coding problem causing NPE. It should be avoided.
And the exception of viewing table now is eaten. It's difficult to find the problem's root cause.

Fixes #2747


PeachThinking (approved):
lgtm

yhilmare:
what's the root cause of this?

yhilmare (approved):
lgtm

PR #2779, opened 2024-06-19
fix(data-transfer): avoid task failure by processing exception messages

### LINKED ISSUE #2774: [Bug]: transfer would fail when the data files are empty
### ODC version

4.3.0

### OB version

-

### What happened?

If import a zip file , and all subfiles are empty, the task will fail.

the task should success

Export some empty tables, and re-import them.

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

Sometimes ob-loader-dumper's error messages are confusing. 
1. If we catch an exception 'No subfiles are generated from path', it's because all the data files are empty. In this case , the task state should not be failed.
2. If we catch an exception 'table or view does not exist', This may be because the user lacks query permissions for the DBA view. We could give users a prompt in the log.

Fixes #2774


yizhouxw (approved):
lgtm

yhilmare (approved):
lgtm

PR #2797, opened 2024-06-21
fix(clientMode): fail to migrate metadb in client mode

### LINKED ISSUE #2760: [Bug]: ODC 4.3.0 would fail to start in client mode

LuckyPickleZZ:
#### What type of PR is this?
type-bug

ODC desktop failed to start Because an incorrect iam_user_role relationship was generated.

Fixes #2760


yhilmare (approved):
lgtm

PR #2801, opened 2024-06-22
refactor(task-executor): involve pull mode for task monitor

yizhouxw:
involve pull mode for task monitor.

in previous version, odc use push mode for task monitor.
however it is not suitable in 'one-way communication' network architecture.

so a new task monitor mode required now.
- task-executor add 2 API , 'heartbea' for ODCServer check status, and 'result' for ODCServer get the latest task result.
- push mode or pull mode is optional for compatibility, it is controlled by TaskMonitorMode configuration .

another scenario is multiple k8s cluster for task execution, sometimes we need execute tasks in diffrent envionment, even different cloud provider, I involve a K8sJobClient interface for supports multiple k8s clusters.
- DefaultK8sJobClientSelector still use single k8s cluster, return the single one.
- developer may define other implemention, then it will be able to supports multiple k8s clusters, odc does not offer a built-in multiple k8s cluster implemenation due the mechanism of task distribution may unpredictable.


server/odc-service/src/main/java/com/oceanbase/odc/service/task/util/TaskExecutorClient.java
```diff
@@ -0,0 +1,114 @@
...
+        try {
+            SuccessResponse<DefaultTaskResult> response =
+                    HttpClientUtils.request("POST", url, new TypeReference<SuccessResponse<DefaultTaskResult>>() {});
+            if (response != null && response.getSuccessful()) {
+                return response.getData();
+            }
+        } catch (IOException e) {
+            log.warn("Query result from executor occur error, ji={}, url={}", ji.getId(), url, e);
+        }
+        return null;
```
yizhouxw, line 111: should not return null

guowl3 (approved):
LGTM

server/odc-service/src/main/java/com/oceanbase/odc/service/task/schedule/StdJobScheduler.java
```diff
@@ -160,6 +165,7 @@ private void tryCanceling(Long jobId) throws JobException {
 
     private void initDaemonJob() {
         initCheckRunningJob();
+        initPullTaskResultJob();
```
yhilmare, line 168: should we disable the TaskMonitor when the TaskMonitorMode is 'PULL'?

server/odc-service/src/main/java/com/oceanbase/odc/service/task/constants/JobServerUrls.java
```diff
@@ -16,22 +16,18 @@
...
  * @since 4.2.4
  */
-public class JobUrlConstants {
+public class JobServerUrls {
 
-    public static final String TASK_RESULT_UPLOAD = "/api/v2/task/result";
+    public static final String TASK_UPLOAD_RESULT = "/api/v2/task/result";
 
-    public static final String TASK_HEART = "/api/v2/task/heart";
+    public static final String TASK_HEARTBEAT = "/api/v2/task/heart";
```
yhilmare, line 29: should we change this url to '/api/v2/task/heartbeat'

server/odc-service/src/main/java/com/oceanbase/odc/service/task/service/StdTaskFrameworkService.java
```diff
@@ -316,30 +336,82 @@ public void handleResult(TaskResult taskResult) {
...
+        taskResultPullerExecutor.execute(() -> {
+            try {
+                doRefreshResult(id);
+            } catch (Exception e) {
+                log.error("Refresh job result failed, jobId={}", id, e);
+            }
+        });
+    }
+
+    private void doRefreshResult(Long id) throws JobException {
```
yhilmare, line 369: the operation logic is same no matter push mode and pull mode, why shouldn't we call handleResult when pull mode, just like what we do in push mode~

server/odc-service/src/main/java/com/oceanbase/odc/service/task/schedule/StdJobScheduler.java
yizhouxw: Yes, actually it is disabled while monitor mode is push

server/odc-service/src/main/java/com/oceanbase/odc/service/task/service/StdTaskFrameworkService.java
yizhouxw: Play to refactor later, current sprint, we keep minimum change for PUSH monitor mode

yhilmare (approved):
lgtm

PR #2842, opened 2024-06-26
feat(alarm): add  cloud task alarm client

guowl3:
#### What type of PR is this?
type-feature

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/alarm/CloudTaskAlarmClient.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/alarm/CloudTaskAlarmClient.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright (c) 2023 OceanBase.
+ *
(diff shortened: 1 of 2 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/alarm/CloudTaskAlarmClient.java
```diff
@@ -0,0 +1,27 @@
...
+package com.oceanbase.odc.service.alarm;
+
+import com.oceanbase.odc.service.alarm.model.CloudTaskAlarmEvent;
+
+/**
+ * @Author：tinker
+ * @Date: 2024/6/26 21:05
+ * @Descripition:
+ */
+public interface CloudTaskAlarmClient {
```
yizhouxw, line 25: we may add generic hook point in odc, then implements in odc cloud.

how about extends current AlarmService ?

PR #2851, opened 2024-06-30
feat(task-framework): handle executor endpoint for PULL monitor mode

yizhouxw:
## main changes

invovle ExecutorEndpointManager for PULL task monitor mode
- will refresh executor point from K8sJobClient for K8S runMode
- ExecutorHostAdapter is designed for self defined network solution extension
- K8bJobClient add podIdAddress in response for support PULL task monitor mode
- while build K8sJobCaller, set ODC_EXECUTOR_PORT/REPORT_ENABLED env variables
- offer a hard code executor listen port, 8080 by default

invovle JobCredentialProvider for separated metadb and cloud object storage in task executor
- return same credential which is used by odc server by default
- developer may offer other implementation for self define executor metadb and cloud object storage

related enhancement
- change jobProperties from SingleJobProperties to Map<String,String> for easy to extends features.
- move CloudProvider to package service.cloud.model, for reuse outside of cloud object storage part.
 

## other changes
1. update CODEOWNERS
2. add JdbcUrlParser utils, may referenced in other plugins
3. some code refine


server/odc-common/src/main/java/com/oceanbase/odc/common/jdbc/JdbcUrlParser.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import com.oceanbase.odc.common.util.StringUtils;
+
+import lombok.Data;
+import lombok.NonNull;
+
+public class JdbcUrlParser {
```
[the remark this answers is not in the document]
yizhouxw: Rename to SimpleJdbcUrlParser

server/plugins/connect-plugin-api/src/main/java/com/oceanbase/odc/plugin/connect/api/JdbcUrlParser.java
```diff
@@ -13,11 +13,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.oceanbase.odc.core.shared.jdbc;
+package com.oceanbase.odc.plugin.connect.api;
 
 import java.util.List;
 import java.util.Map;
 
+import com.oceanbase.odc.core.shared.jdbc.HostAddress;
```
yhilmare, line 21: HostAddress may be also moved here

server/odc-common/src/main/java/com/oceanbase/odc/common/jdbc/JdbcUrlParser.java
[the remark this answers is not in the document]
yizhouxw: change to  SimpleJdbcUrlParser  now

server/plugins/connect-plugin-api/src/main/java/com/oceanbase/odc/plugin/connect/api/JdbcUrlParser.java
yizhouxw: okay, moved now

yhilmare (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/task/util/JobUtils.java
```diff
@@ -132,14 +133,15 @@ public static boolean isProcessRunModeOfEnv() {
     public static ConnectionConfig getMetaDBConnectionConfig() {
         ConnectionConfig config = new ConnectionConfig();
         config.setHost(System.getProperty(JobEnvKeyConstants.ODC_EXECUTOR_DATABASE_HOST));
```
guowl3, line 135: environment variable could expire while the main application is still running?

yizhouxw: may occur in process mode scenario, let's trace in separated issue

guowl3 (approved):
lgtm

PR #2863, opened 2024-07-02
fix(audit): client ip length more langer then audit column client_ip_address

### LINKED ISSUE #1179: [Bug]: client ip length more langer then  audit column `client_ip_address`
### ODC version

4.2.2

### OB version

any

### What happened?

[2023-12-07 17:03:55.809][http-nio-8989-exec-6][8076ce2d24094063,,MI8ZUAW10IL7DAZNIO77WI][INFO][com.oceanbase.odc.server.web.controller.RestExceptionHandler][375]: ODC_FAILED_REQUEST, httpMethod=POST, fullURL=http://ob-odc.pek01.in.zhihu.com/api/v2/datasource/datasources?currentOrganizationId=1&wantCatchError=false&holdErrorTip=true, remoteHost=10.152.4.120, clientAddress=10.19.225.76, 10.255.252.1, 10.19.225.76, 10.152.8.80,  handler=com.oceanbase.odc.server.web.controller.v2.DataSourceController#createDataSource[1 args], exceptionType=DataIntegrityViolationException, message=could not execute statement; SQL [n/a]; nested exception is org.hibernate.exception.DataException: could not execute statement, rootReason=     at com.oceanbase.jdbc.internal.protocol.AbstractQueryProtocol.executeQuery(AbstractQueryProtocol.java:399)      at com.oceanbase.jdbc.internal.protocol.AbstractQueryProtocol.getResult(AbstractQueryProtocol.java:2115)        at com.oceanbase.jdbc.internal.protocol.AbstractQueryProtocol.readPacket(AbstractQueryProtocol.java:2227)       at com.oceanbase.jdbc.internal.protocol.AbstractQueryProtocol.readErrorPacket(AbstractQueryProtocol.java:2364) java.sql.SQLException: Data too long for column 'client_ip_address' at row 1, perfLevel=P1, response=ErrorResponse(error=Error(code=DataAccessError, message=访问元数据库错误，请联系管理员, details=[]), code=DataAccessError, message=访问元数据库错误，请联系管理员)

`10.19.225.76, 10.255.252.1, 10.19.225.76, 10.152.8.80`  length then 46 of client_ip_address in audit table

save client ip normal

use nginx revert proxy odc and append middle ip in heard_client

_No response_

### Cloud

_No response_

CHLK:
#### What type of PR is this?

type-bug
client ip length more langer then audit column client_ip_address

Fixes #1179 

```docs

```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,4 +52,3 @@ lib/odc-*.jar
 maven/
 venv/
 **/local-unit-test.properties
-/.github/hooks/pre-push
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,4 @@ lib/odc-*.jar
 maven/
 venv/
 **/local-unit-test.properties
+/.github/hooks/pre-push
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/common/util/WebRequestUtils.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/common/util/WebRequestUtils.java
@@ -118,37 +118,6 @@ public class WebRequestUtils {
         return request.getRemoteAddr();
     }
 
-    public static String getTrulyClientIp(HttpServletRequest request) {
-        return getTrulyClientIp(getClientAddress(request));
-    }
-
-    /**
-     * <pre>
-     * Find truly ip of client.
-     * The X-Forwarded-For header may contain multiple IP addresses, separated
-     * by commas, and typically, the first non-unknown IP is considered to be the client's IP address.
-     * </pre>
-     * 
-     * @author keyang.lk
-     * @date 2024-07-02
-     * @param remoteAddress
-     * @return The truly ip of remoteAddress
-     */
-    public static String getTrulyClientIp(String remoteAddress) {
-        if (remoteAddress == null || remoteAddress.isEmpty() || "unknown".equalsIgnoreCase(remoteAddress)) {
-            return "N/A";
-        }
-        // 处理X-Forwarded-For可能包含多个IP地址的情况（由逗号分隔），通常第一个非unknown的IP是客户端的IP
-        String[] ips = remoteAddress.split(",");
-        for (String ip : ips) {
-            if (ip != null && !ip.isEmpty() &&
-                    !"unknown".equalsIgnoreCase(ip.trim())) {
-                return ip.trim();
-            }
-        }
-        return remoteAddress;
-    }
-
     public static String getUserAgent(HttpServletRequest request) {
         if (Objects.isNull(request)) {
             return "N/A";
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/common/util/WebRequestUtils.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/common/util/WebRequestUtils.java
@@ -118,6 +118,37 @@ public class WebRequestUtils {
         return request.getRemoteAddr();
     }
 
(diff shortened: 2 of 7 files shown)


server/odc-service/src/test/java/com/oceanbase/odc/service/common/WebRequestUtilsTest.java
```diff
@@ -49,4 +52,28 @@ public void isRedirectUrlValid_HostNotMatch_InValid() {
 
         Assert.assertFalse(redirectUrlValid);
     }
+
+    @Test
+    public void testGetTrulyClientIp() {
```
yhilmare, line 57: test case's name suggestion: ${test method name}_${test scenarios}_${expected result}

this test case may be named: getTrulyClientIp_normalInput_getSucceed

server/odc-service/src/main/java/com/oceanbase/odc/service/common/util/WebRequestUtils.java
```diff
@@ -118,6 +118,37 @@ public static String getClientAddress(HttpServletRequest request) {
...
+     * The X-Forwarded-For header may contain multiple IP addresses, separated
+     * by commas, and typically, the first non-unknown IP is considered to be the client's IP address.
+     * </pre>
+     * 
+     * @author keyang.lk
+     * @date 2024-07-02
+     * @param remoteAddress
+     * @return The truly ip of remoteAddress
+     */
+    public static String getTrulyClientIp(String remoteAddress) {
```
yhilmare, line 137: your method may be renamed to `getTrulyClientIpFromXForwardedForHttpHeader` if this method can only process the remote address from this http header~

CHLK: Considering that the client's real IP address can be obtained not only from the X-Forwarded-For header but also potentially from other headers such as Remote_ADDR, would it be appropriate to rename this method to getTrulyClientIpFromHttpHeader?

server/odc-service/src/test/java/com/oceanbase/odc/service/common/WebRequestUtilsTest.java
CHLK: Would it be appropriate to name this test method, which encompasses both correct and erroneous test scenarios, as getTrulyClientIp_allTestCases?

server/odc-service/src/main/java/com/oceanbase/odc/service/common/util/WebRequestUtils.java
yizhouxw: how about `getFirstClientIpAddress`, and may move this method to audit module.

considering:
1. we cannot confirm this method will return the 'truly' client ip
2. it is not a problem in web request area

.gitignore
```diff
@@ -52,3 +52,4 @@ lib/odc-*.jar
 maven/
 venv/
 **/local-unit-test.properties
+/.github/hooks/pre-push
```
yhilmare, line 55: why should we add this?

CHLK: In Windows, I need to add a pre-push  to my github hooks, but others do not require this. Therefore, I have added it to .gitignore.

yizhouxw (approved):
Congratulations for first Pull Request!

yhilmare (approved):
lgtm

PR #2866, opened 2024-07-02
feat(task-framework): task log download support multiple region scenario

yizhouxw:
task log download support multiple region scenario.
add TENCENT_CLOUD and HUAWEI_CLOUD supports.


yhilmare (approved):
lgtm

PR #2872, opened 2024-07-02
fix(sql): the sql of modifying session parameter in oracle is error

### LINKED ISSUE #2736: [Bug]: oracle datasource nls_date_format do not use "set "command
### ODC version

ODC 4.3.0 

### OB version

it is not ob ,it is oracle

### What happened?

set command is used to mysql or ob database ,not use to oracle. 
截屏2024-06-13 10 08 13

"set parameter_xxx=aaa"  change to "alter session set parameter_xxx=aaa"

"set parameter_xxx=aaa"  change to "alter session set parameter_xxx=aaa"

_No response_

zijiacj:

Fixes #2736

```docs

```

### DIFF
--- a/server/plugins/connect-plugin-api/src/main/java/com/oceanbase/odc/plugin/connect/api/SessionExtensionPoint.java
+++ b/server/plugins/connect-plugin-api/src/main/java/com/oceanbase/odc/plugin/connect/api/SessionExtensionPoint.java
@@ -35,5 +35,5 @@ public interface SessionExtensionPoint extends ExtensionPoint, SessionOperations
 
     String getVariable(Connection connection, String variableName);
 
-    String getAlterSessionVariableStatement(String variableScope, String variableName, String variableValue);
+    String getAlterVariableStatement(String variableScope, String variableName, String variableValue);
 }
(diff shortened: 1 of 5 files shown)


yhilmare (dismissed):
lgtm

server/plugins/connect-plugin-api/src/main/java/com/oceanbase/odc/plugin/connect/api/SessionExtensionPoint.java
```diff
@@ -34,4 +34,6 @@ public interface SessionExtensionPoint extends ExtensionPoint, SessionOperations
     String getCurrentSchema(Connection connection);
 
     String getVariable(Connection connection, String variableName);
+
+    String getAlterSessionVariableStatement(String variableScope, String variableName, String variableValue);
```
yhilmare, line 38: the method name is not so good, we should not emphasize conference in method names

server/plugins/connect-plugin-oracle/src/main/java/com/oceanbase/odc/plugin/connect/oracle/OracleSessionExtension.java
```diff
@@ -89,4 +89,9 @@ public String getVariable(@NonNull Connection connection, @NonNull String variab
         }
         return value;
     }
+
+    @Override
+    public String getAlterVariableStatement(String variableScope, String variableName, String variableValue) {
```
yhilmare, line 94: how to process global/system parameter altering?

yhilmare (approved):
lgtm

PR #2874, opened 2024-07-03
fix(i18n): fix i18n docs encoding error

yhilmare:
#### What type of PR is this?

type-bug

fix i18n docs encoding error

Fixes #

```docs

```


yizhouxw (approved):
lgtm

PR #2883, opened 2024-07-03
fix(migrate): fix login process resource load faild

### LINKED ISSUE #2880: [Bug]: Login failed
### ODC version

odc431

### OB version

*

### What happened?

image

Login successful

Login to shilin/aaAA11__

_No response_

yiminpeng:
#### What type of PR is this?
fix

fix login process resource load faild, caused by delete useless yaml

Fixes #2880

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/collaboration/IndividualOrganizationMigrator.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/collaboration/IndividualOrganizationMigrator.java
@@ -99,8 +99,6 @@ public class IndividualOrganizationMigrator {
         List<String> resourceLocations = new LinkedList<>();
         resourceLocations.add("migrate/common/V_3_2_0_6__iam_permission.yaml");
         resourceLocations.add("migrate/common/V_3_3_0_4__iam_permission.yaml");
-        resourceLocations.add("migrate/common/V_3_4_0_13__data_masking_rule.yaml");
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/collaboration/IndividualOrganizationMigrator.java
```diff
@@ -99,7 +99,6 @@ private ResourceConfig getResourceConfig(DataSource dataSource, Long userId, Lon
         List<String> resourceLocations = new LinkedList<>();
         resourceLocations.add("migrate/common/V_3_2_0_6__iam_permission.yaml");
         resourceLocations.add("migrate/common/V_3_3_0_4__iam_permission.yaml");
-        resourceLocations.add("migrate/common/V_3_4_0_2__iam_permission.yaml");
```
MarkPotato777, line 102: You may need to remove all Yaml files which were removed in your last PR, rather than only this `V_3_4_0_2__iam_permission.yaml`

yiminpeng: The other two file are missing, re-pushing now. A total of three related documents.

yiminpeng: The other two file are missing, re-pushing now. A total of three related documents.

MarkPotato777 (approved):
LGTM

PR #2908, opened 2024-07-05
fix(osc): the online schema change blocked when rate limiter modified before swap table action

LioRoger:
#### What type of PR is this?
BugFix

This PR include two bug fix

1. When online schema change service start, but the async flow task has not scheduled. Request 
 to /api/v2/osc/updateRateLimitConfig may throw NPE.
2. When  /api/v2/osc/updateRateLimitConfig  requested before swap table action. server will loop restart OMS task cause limit config not saved correctly.

Rate limit diff check depends on message stored in schedule_schedule and schedule_task table. The message will not generated until flow task triggered. issue 1 happened before flow task triggered. 

If rate limit resolved different between schedule_schedule and schedule_task, update OMS config operation will triggered. Then schedule_task will assign same rate limit of schedule_schedule. issue 2 save previous rate limit again.

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/DefaultOnlineSchemaChangeTaskHandler.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/DefaultOnlineSchemaChangeTaskHandler.java
@@ -486,12 +486,12 @@ public class DefaultOnlineSchemaChangeTaskHandler implements OnlineSchemaChangeT
     }
 
     private OnlineSchemaChangeScheduleTaskParameters parseOnlineSchemaChangeScheduleTaskParameters(String jsonStr) {
-        OnlineSchemaChangeScheduleTaskParameters OnlineSchemaChangeScheduleTaskParameters = JsonUtils.fromJson(
+        OnlineSchemaChangeScheduleTaskParameters onlineSchemaChangeScheduleTaskParameters = JsonUtils.fromJson(
                 jsonStr, OnlineSchemaChangeScheduleTaskParameters.class);
         // correct null value to default RateLimiterConfig object
-        if (null == OnlineSchemaChangeScheduleTaskParameters.getRateLimitConfig()) {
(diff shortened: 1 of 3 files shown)


yizhouxw:
For pull request tittle , let’s follow convenionalcommits guide: https://www.conventionalcommits.org/en/v1.0.0/#summary

E.g.
fix(online-schema-change): modify rate limit cause the task cannot continue

Better simply in title, only contains the problem behavior, the title will be used for auto generation CHANGELOG document

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/DefaultOnlineSchemaChangeTaskHandler.java
```diff
@@ -468,4 +469,29 @@ private void dropNewTableIfExits(OnlineSchemaChangeScheduleTaskParameters taskPa
...
+                jsonStr, OnlineSchemaChangeParameters.class);
+        // correct null value to default RateLimiterConfig object
+        if (null == onlineSchemaChangeParameters.getRateLimitConfig()) {
+            onlineSchemaChangeParameters.setRateLimitConfig(new RateLimiterConfig());
+        }
+        return onlineSchemaChangeParameters;
+    }
+
+    private OnlineSchemaChangeScheduleTaskParameters parseOnlineSchemaChangeScheduleTaskParameters(String jsonStr) {
+        OnlineSchemaChangeScheduleTaskParameters OnlineSchemaChangeScheduleTaskParameters = JsonUtils.fromJson(
```
LuckyPickleZZ, line 489: The naming of variables should use lower camera case

LioRoger: fixed

LuckyPickleZZ (approved):
LGTM

yizhouxw (approved):
lgtm

PR #2924, opened 2024-07-08
fix(connection): automatically reconnect during the cloud upgrade process

PeachThinking:
#### What type of PR is this?
type-bug
module-connection
Fixed the issue that the connection could not be automatically reconnected during the cloud upgrade process.
The `stateRouteMethods` method has already determined whether the forwarded node is a healthy node before forwarding, so there is no need to change the code logic of `StateRouteAspect`：
image

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/session/ConnectSessionService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/session/ConnectSessionService.java
@@ -330,7 +330,8 @@ public class ConnectSessionService {
         ConnectionSession session = connectionSessionManager.getSession(sessionId);
         if (session == null) {
             CreateSessionReq req = new DefaultConnectSessionIdGenerator().getKeyFromId(sessionId);
-            if (!autoCreate || !StringUtils.equals(req.getFrom(), stateHostGenerator.getHost())) {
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/session/ConnectSessionService.java
```diff
@@ -330,7 +330,8 @@ public ConnectionSession nullSafeGet(@NotNull String sessionId, boolean autoCrea
         ConnectionSession session = connectionSessionManager.getSession(sessionId);
         if (session == null) {
             CreateSessionReq req = new DefaultConnectSessionIdGenerator().getKeyFromId(sessionId);
-            if (!autoCreate || !StringUtils.equals(req.getFrom(), stateHostGenerator.getHost())) {
+            if (!autoCreate || (!StringUtils.equals(req.getFrom(), stateHostGenerator.getHost())
```
yizhouxw, line 333: may add 
```
if (cloudMetadataClient.supportsCloudMetadata()) {
  autoRecreate = true;
}
```

for easy to read code.

yhilmare (approved):
lgtm

yizhouxw (approved):

PR #2928, opened 2024-07-08
fix(global-search): unable to stop data object synchronization

### LINKED ISSUE #2767: [Bug]: Metabase synchronization has been initiated continuously
### ODC version

odc430bp

### OB version

*

### What happened?

image

Metabase synchronization will not be initiated continuously

Enter link

_No response_

yhilmare:
#### What type of PR is this?

type-bug

the front-end will call sync api continuously, the reason of it is that the object sync status is 'PENDING' without changing by the time. 

I add a new property `max-pending-hours` to limit the maximum time an object's state can be kept on 'PENDING' to fix this issue.

Fixes #2767

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
@@ -801,9 +801,10 @@ public class DatabaseService {
     @SkipAuthorize("odc internal usage")
     @Transactional(rollbackFor = Exception.class)
     public void refreshExpiredPendingDBObjectStatus() {
-        this.databaseRepository.setObjectSyncStatusByObjectSyncStatusAndObjectLastSyncTimeBefore(
-                DBObjectSyncStatus.INITIALIZED, DBObjectSyncStatus.PENDING,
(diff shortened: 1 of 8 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
```diff
@@ -793,6 +798,14 @@ public void updateObjectLastSyncTimeAndStatus(@NotNull Long databaseId,
         databaseRepository.setObjectLastSyncTimeAndStatusById(databaseId, new Date(), status);
     }
 
+    @SkipAuthorize("odc internal usage")
+    @Transactional(rollbackFor = Exception.class)
+    public void refreshExpiredPendingDBObjectStatus() {
+        this.databaseRepository.setObjectSyncStatusByObjectSyncStatusAndObjectLastSyncTimeBefore(
```
yizhouxw, line 804: we may add log here ,how much rows affected.

yhilmare: done

yizhouxw (approved):
lgtm

MarkPotato777 (approved):
LGTM

PeachThinking (approved):
lgtm

PR #2930, opened 2024-07-08
fix(pl-debug): avoid npe during the pl debugging

### LINKED ISSUE #2693: [Bug]: optimize error message when pl-deug encounter an exception
### ODC version

4.2.4

### OB version

not related

### What happened?

the front end will give a message when encounter an exception:
63D2FA85-2492-4B40-9251-54C253316A66

we should optimize the error message

1. start pl debug
2. wait for 10min
3. continue debug

_No response_

yhilmare:
#### What type of PR is this?

type-bug

the null pointer exception will be thrown in some situations.

Fixes #2693

```docs

```


yizhouxw (approved):
lgtm

PR #2943, opened 2024-07-09
fix(pre-check): failed to get sql check result when the check result file is not on this machine

### LINKED ISSUE #2724: [Bug]: Multiple nodes access ODC , and the pre-check results cannot be viewed in task details
### ODC version

ODC430

### OB version

independent

### What happened?

Multiple nodes access ODC through domain names, and the pre-check results cannot be viewed in task details
image
image

can view it

chat privately

_No response_

yhilmare:
#### What type of PR is this?

type-bug

the type reference of the pre_check task is wrong and this pr fix this~

Fixes #2724

```docs

```


yizhouxw (approved):
lgtm

Issue #2869, opened 2024-07-02
[Bug]: Creating function with SQL statement TRACE ID unable to view execution profile


sl01388797:
### ODC version

odc431

### OB version

ob424

### What happened?

![image](https://github.com/oceanbase/odc/assets/143492234/02b041f2-e8ce-4c42-908d-e0853d006164)
![image](https://github.com/oceanbase/odc/assets/143492234/9f9bacd1-f4c2-4515-87f0-9f1d4d7b3b75)

### What did you expect to happen?

No error reported

### How can we reproduce it (as minimally and precisely as possible)?

1. Create stored procedure
create procedure `pro6` (INOUT `a` varchar(45)) begin
declare
  v1 int;
set
  a = 2;
set
  v1 = a + 1;
select
  v1;
end;
2. Click on TRACE ID

sl01388797:
![image](https://github.com/user-attachments/assets/87d5f541-1398-41d8-823f-138e904f29ac)
pass

Issue #2852, opened 2024-06-30
[Bug]: there maybe error occurs when explaining sql


LuckyPickleZZ:
### ODC version

4.3.1

### OB version

4.2.4

### What happened?

![image](https://github.com/oceanbase/odc/assets/36879124/b1ef0371-6c86-4163-b543-8992cbee86e0)

### What did you expect to happen?

-

### How can we reproduce it (as minimally and precisely as possible)?

explain on tpch database:
```sql
SELECT NATION, O_YEAR, SUM(AMOUNT) AS SUM_PROFIT
FROM (SELECT N_NAME AS NATION,
             TO_CHAR(O_ORDERDATE, 'YYYY') AS O_YEAR,
             L_EXTENDEDPRICE*(1-L_DISCOUNT)-PS_SUPPLYCOST*L_QUANTITY AS AMOUNT
      FROM PART, SUPPLIER, LINEITEM, PARTSUPP, ORDERS, NATION
      WHERE S_SUPPKEY = L_SUPPKEY AND
            PS_SUPPKEY= L_SUPPKEY AND
            PS_PARTKEY = L_PARTKEY AND
            P_PARTKEY= L_PARTKEY AND
            O_ORDERKEY = L_ORDERKEY AND
            S_NATIONKEY = N_NATIONKEY AND
            P_NAME LIKE '%%green%%') PROFIT
GROUP BY NATION, O_YEAR ORDER BY NATION, O_YEAR DESC;
```

sl01388797:
![Image](https://github.com/user-attachments/assets/ffb933b1-e055-4df8-a020-7b664a26ad14)

pass

PR #2945, opened 2024-07-09
fix(diagnose): failed to view query profile for distributed OB

### LINKED ISSUE #2828: [Bug]: Error reported after clicking TRACE ID after execution failure
### ODC version

odc431

### OB version

*

### What happened?

image
image
image

TRACE ID not displayed

Click on TRACE ID after SQL execution failure

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

After testing query profile on a 3-node cluster, I found several issues in this case. The main works in this PR:
1. use session id, server ip, server port to accurately locate SQL plan records.
2. kill query if there is an exception when getting more results.
3. adapted to the execution plan on OB 4.2
4. added a variable to identify whether the current SQL supports viewing query profiles. The condition for this variable to be true is that the OB version is higher than 4.2.4 and the SQL type is `select`.

Fixes #2828 #2852 #2869

```docs

```

### DIFF
--- a/server/odc-core/src/main/java/com/oceanbase/odc/core/shared/model/ExecutorInfo.java
+++ b/server/odc-core/src/main/java/com/oceanbase/odc/core/shared/model/ExecutorInfo.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright (c) 2023 OceanBase.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.oceanbase.odc.core.shared.model;
+
+import lombok.Data;
+
+/**
+ * @author: liuyizhuo.lyz
+ * @date: 2024/7/4
+ */
+@Data
+public class ExecutorInfo {
+
+    private String ip;
+    private String port;
+    private String tenantId;
+    private String planId;
+
+}
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/session/ConnectConsoleService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/session/ConnectConsoleService.java
@@ -357,7 +357,11 @@ public class ConnectConsoleService {
             shouldRemoveContext = true;
             // Front-end would stop getting more results if there is an exception. In this case the left queries
             // should be killed.
-            killCurrentQuery(sessionId);
+            try {
+                killCurrentQuery(sessionId);
+            } catch (Exception ex) {
+                log.warn("Failed to kill query. Session id={}. Request id={}", sessionId, requestId);
+            }
             throw e;
         } finally {
             if (shouldRemoveContext) {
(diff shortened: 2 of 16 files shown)


server/odc-core/src/main/java/com/oceanbase/odc/core/shared/model/ExecutorInfo.java
```diff
@@ -0,0 +1,32 @@
...
+package com.oceanbase.odc.core.shared.model;
+
+import lombok.Data;
+
+/**
+ * @author: liuyizhuo.lyz
+ * @date: 2024/7/4
+ */
+@Data
+public class ExecutorInfo {
```
yhilmare, line 25: we may change another name, this name is occupied

server/odc-service/src/main/java/com/oceanbase/odc/service/session/ConnectConsoleService.java
```diff
@@ -589,6 +592,13 @@ private SqlExecuteResult generateResult(@NonNull ConnectionSession connectionSes
         } catch (Exception e) {
             log.warn("Failed to init warning message", e);
         }
+        try {
+            String version = ConnectionSessionUtil.getVersion(connectionSession);
```
yhilmare, line 596: we may set this flag by checking if there is anything in connection session's cache~

server/odc-service/src/main/java/com/oceanbase/odc/service/session/OBExecutionListener.java
```diff
@@ -31,16 +31,21 @@
...
 import com.oceanbase.tools.dbbrowser.parser.constant.SqlType;
+import com.oceanbase.tools.dbbrowser.parser.result.BasicResult;
+
+import lombok.extern.slf4j.Slf4j;
 
 /**
  * @author: liuyizhuo.lyz
  * @date: 2024/4/23
  */
+@Slf4j
```
yhilmare, line 48: we may change another name to tell the coder that this listener is only used for query profile~

server/odc-service/src/main/java/com/oceanbase/odc/service/session/OBExecutionListener.java
```diff
@@ -92,14 +98,20 @@ private List<String> getSessionIds() {
...
+            return session.getSyncJdbcExecutor(CONSOLE_DS_KEY).execute((StatementCallback<List<String>>) stmt -> OBUtils
+                    .querySessionIdsByProxySessId(stmt, proxySessId, session.getConnectType()));
+        } catch (Exception e) {
+            log.warn("Failed to init session ids. Reason:{}", e.getMessage());
+            return Collections.emptyList();
+        }
     }
 
-    private boolean isSelect(SqlTuple sqlTuple) {
+    public static boolean isSqlTypeSupportProfile(SqlTuple sqlTuple) {
```
yhilmare, line 110: prefer not static~ and not public

server/odc-core/src/main/java/com/oceanbase/odc/core/shared/model/ExecutorInfo.java
LuckyPickleZZ: sure, I renamed it into `OBExecutionServerInfo`

server/odc-service/src/main/java/com/oceanbase/odc/service/session/ConnectConsoleService.java
LuckyPickleZZ: Sorry but I thought about it again, this cannot be changed. 
Perhaps this SQL actually supports viewing query profiles, but there is currently no record in the cache. Because the process of caching is asynchronous, we cannot guarantee it.

server/odc-service/src/main/java/com/oceanbase/odc/service/session/OBExecutionListener.java
LuckyPickleZZ: renamed to `OBQueryProfileExecutionListener`

yhilmare (dismissed):
lgtm

yhilmare (approved):
lgtm

yizhouxw (approved):
lgtm

PR #2946, opened 2024-07-09
fix(permission): fail to submit ticket if lack of database permission

### LINKED ISSUE #2915: [Bug]: Table change permission cannot take effect
### ODC version

odc431

### OB version

*

### What happened?

image
image
image

Tables with table change permissions can perform database changes

Low privilege account application form change permission, execute database change

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

If  a user is granted some tables' permission without database permission, he could not sumit database change ticket.

Fixes #2915

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/permission/DBResourcePermissionHelper.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/permission/DBResourcePermissionHelper.java
@@ -423,10 +423,9 @@ public class DBResourcePermissionHelper {
                         }));
         Map<Long, Set<DatabasePermissionType>> typesFromTable = userTablePermissionRepository
                 .findNotExpiredByUserIdAndDatabaseIdIn(authenticationFacade.currentUserId(), databaseIds)
-                .stream().collect(Collectors.toMap(
-                        UserTablePermissionEntity::getDatabaseId,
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/permission/DBResourcePermissionHelper.java
```diff
@@ -425,7 +425,7 @@ private Map<Long, Set<DatabasePermissionType>> getInnerDBPermissionTypes(Collect
                 .findNotExpiredByUserIdAndDatabaseIdIn(authenticationFacade.currentUserId(), databaseIds)
                 .stream().collect(Collectors.toMap(
                         UserTablePermissionEntity::getDatabaseId,
-                        e -> Collections.singleton(DatabasePermissionType.ACCESS),
+                        e -> Collections.singleton(DatabasePermissionType.from(e.getAction())),
```
yhilmare, line 428: group by will be better, cause that there will be multi tables that have the same database id

LuckyPickleZZ: done~

yhilmare (approved):
LGTM

MarkPotato777 (approved):
LGTM

PR #2948, opened 2024-07-09
fix(db-browser):show database object create time incorrectly with db session time zone

### LINKED ISSUE #2947: [Bug]: When querying the meta information related to the time of the data source object, the returned result of the database is automatically converted based on the current session time_zone
### ODC version

odc431

### OB version

*

### What happened?

94a24153066051040495fb9864fdf225
79f2d9a55e0e2ef9fc339d9dbbb986c4

Time needs to be returned by the backend

1. Modify Time Zone
2. Create table, query time
3. View table creation time

_No response_

PeachThinking:
#### What type of PR is this?
type-bug

fix show database object create and update time incorrectly with session time zone
`convertToDefaultTimeZone` 

Fixes #2947

```docs

```

### DIFF
--- a/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/doris/DorisSchemaAccessor.java
+++ b/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/doris/DorisSchemaAccessor.java
@@ -1279,4 +1279,9 @@ public class DorisSchemaAccessor implements DBSchemaAccessor {
         throw new UnsupportedOperationException("Not supported yet");
     }
 
+    @Override
+    public String getSessionTimeZone() {
+        throw new UnsupportedOperationException("Not supported yet");
+    }
+
 }
--- a/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/util/TimestampUtils.java
+++ b/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/util/TimestampUtils.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) 2023 OceanBase.
+ *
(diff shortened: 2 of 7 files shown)


libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/util/TimestampUtils.java
```diff
@@ -0,0 +1,52 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * @author jingtian
+ * @date 2024/7/8
+ */
+@Slf4j
+public class TimestampUtils {
+    public static Timestamp convertToDefaultTimeZone(@NonNull Timestamp timestamp, @NotBlank String timeZone) {
+        DateFormat sdf = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
```
yizhouxw, line 39: unnessesary format and parse may involve performance cost,
as timestamp is really an long type, we may use value based convert straightly.

e.g. convert timestamp to instant long value. 

and, I try to understand  TimeZone.getDefault(), is it hardcode as UTC+8, sed in Dockerfile ?

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/doris/DorisSchemaAccessor.java
```diff
@@ -1279,4 +1279,9 @@ public Map<String, DBTable> getTables(String schemaName, List<String> tableNames
         throw new UnsupportedOperationException("Not supported yet");
     }
 
+    @Override
+    public String getSessionTimeZone() {
+        throw new UnsupportedOperationException("Not supported yet");
```
yizhouxw, line 1284: how do we handle this exception?

PR #2958, opened 2024-07-11
enhancement(result-set): only query table constraints when it's necessary

### LINKED ISSUE #2952: [Bug]: A statement that modifies a result set to generate a rowid should not check constraints
### ODC version

ODC430

### OB version

independent

### What happened?

A statement that modifies a result set to generate a rowid should not check constraints
This will cause the interface speed of generating SQL statements to be extremely slow

。

。

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-enhancement

When editing result set, ODC would always query constraints. Sometimes this is not necessary.  Like only edit rows with `rowid`.

Fixes #2952

```docs

```

### DIFF
--- a/server/odc-common/src/main/java/com/oceanbase/odc/common/util/LazyInitObject.java
+++ b/server/odc-common/src/main/java/com/oceanbase/odc/common/util/LazyInitObject.java
@@ -1,0 +1,46 @@
+/*
+ * Copyright (c) 2023 OceanBase.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.oceanbase.odc.common.util;
+
+import java.util.function.Supplier;
+
+/**
+ * @author: liuyizhuo.lyz
+ * @date: 2024/7/11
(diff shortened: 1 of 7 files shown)


yhilmare (approved):
lgtm

PeachThinking (approved):

server/odc-common/src/main/java/com/oceanbase/odc/common/util/LazyInitObject.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.function.Supplier;
+
+import lombok.NonNull;
+
+/**
+ * @author: liuyizhuo.lyz
+ * @date: 2024/7/11
+ */
+public class LazyInitObject<T> {
```
yizhouxw, line 26: may simplify name `Lazy`.

e.g. org.springframework.data.util.Lazy

LuckyPickleZZ: sure , renamed~

yhilmare (approved):
lgtm

PeachThinking (approved):

server/odc-common/src/main/java/com/oceanbase/odc/common/util/Lazy.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.function.Supplier;
+
+import lombok.NonNull;
+
+/**
+ * @author: liuyizhuo.lyz
+ * @date: 2024/7/11
+ */
+public class Lazy<T> {
+    private static final Object NO_INIT = new Object();
```
LioRoger, line 27: In my opinion, Lazy may not give accurate meaning of the class intend to do. This class want to provide a delayed function call if needed. It's more like a local cache, if missing then provide a value use provider.
If supplier call returns NO_INIT, get method will be always invoked. Get method seems not a high frequent access function. So move synchronized keyword to  method declare, use init flag to control supplier get seems more reasonable

PR #2959, opened 2024-07-11
fix(session): support auto recreate session when upgrade

PeachThinking:
#### What type of PR is this?
type-bug

When automatically reconnecting in the cloud upgrade scenario, the host is obtained according to `req.getFrom()`, so that the created sessionId is consistent with the previously not found sessionId

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/session/ConnectSessionService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/session/ConnectSessionService.java
@@ -276,8 +276,7 @@ public class ConnectSessionService {
         DefaultConnectSessionIdGenerator idGenerator = new DefaultConnectSessionIdGenerator();
         idGenerator.setDatabaseId(req.getDbId());
         idGenerator.setFixRealId(StringUtils.isBlank(req.getRealId()) ? null : req.getRealId());
-        if (Objects.nonNull(req.getFrom()) && cloudMetadataClient.supportsCloudMetadata()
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/session/ConnectSessionService.java
```diff
@@ -276,7 +276,8 @@ public ConnectionSession create(@NotNull CreateSessionReq req) {
         DefaultConnectSessionIdGenerator idGenerator = new DefaultConnectSessionIdGenerator();
         idGenerator.setDatabaseId(req.getDbId());
         idGenerator.setFixRealId(StringUtils.isBlank(req.getRealId()) ? null : req.getRealId());
-        idGenerator.setHost(stateHostGenerator.getHost());
+        String host = cloudMetadataClient.supportsCloudMetadata() ? req.getFrom() : stateHostGenerator.getHost();
```
yhilmare, line 279: `req.getFrom` may be null and we should consider of this:
```java
if (!cloudMetadataClient.supportsCloudMetadata() || req.getFrom() == null) {
    idGenerator.setHost(stateHostGenerator.getHost());
} else {
     idGenerator.setHost(req.getFrom());
}
```

PeachThinking: ok

yhilmare (commented):
lgtm

PR #2961, opened 2024-07-11
fix(task-framework): ignore location validation and update task status after refresh

guowl3:

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/objectstorage/cloud/CloudObjectStorageService.java
```diff
@@ -105,7 +105,8 @@ public CloudObjectStorageService(CloudObjectStorage publicEndpointCloudObjectSto
         this.internalEndpointCloudObjectStorage = internalEndpointCloudObjectStorage;
         this.objectStorageConfiguration = objectStorageConfiguration;
         if (this.publicEndpointCloudObjectStorage.supported()) {
-            validateBucket();
+            // TODO 多云获取 location 异常，暂不校验
+            // validateBucket();
```
yizhouxw, line 109: skip validate for cos only.

this validate help detect wrong configuration for alibaba cloud.

and, please use english for comments.

yizhouxw (approved):
lgtm

PR #2974, opened 2024-07-17
fix(diagnose): get incorrect record from sql audit

LuckyPickleZZ:
#### What type of PR is this?
type-bug

There may be some ob inner sqls sharing same trace id with target sql in sql audit view. We should filter these.


PeachThinking (approved):
lgtm

PR #2975, opened 2024-07-17
fix(stateful): remove wrong condition

ungreat:
#### What type of PR is this?
`odc.web.stateful-route.enabled` should only control route, `StatefulUuidStateIdGenerator` is used to generate stateId, should not conditional load.

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #2976, opened 2024-07-17
fix(stateful): remove wrong generator conditional load

ungreat:
#### What type of PR is this?

so as https://github.com/oceanbase/odc/pull/2975

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #3006, opened 2024-07-24
fix(data-transfer): clean work directory before import

### LINKED ISSUE #3005: [Bug]: ODC client may import dirty data files
### ODC version

4.3.x client-mode

### OB version

-

### What happened?

When importing sql files in ODC, the historical dirty files would be imported.

Clean files before import.

1. Import sql files in ODC client.
2. Clean the h2 metadb  files.
3. Restart odc, import some sql files.
4. The files in step 2 would be imported too.

_No response_

LuckyPickleZZ:
If ODC was upgraded from versions earlier than 4.2.3, and the metadb files are deleted, we may import some dirty data files. So we should clean the work directory first.

fix #3005


yhilmare (approved):
lgtm

PR #3012, opened 2024-07-25
fix(schedule): validate trigger interval when create scheduler

### LINKED ISSUE #3008: [Enhancement]: we need to limit the frequency of the partition plan

we need to limit the frequency of the partition plan

### Why is this needed?

there may be a lot of task executed when users execute partition planning at a very high frequency

guowl3:
#### What type of PR is this?
type-bug

validate cron trigger and throw exception if trigger interval less than 600s.

Fixes #3008

```docs

```


yhilmare (approved):
lgtm

yizhouxw (approved):

Issue #2845, opened 2024-06-27
[Bug]: Tables outside the required range appear in the archive cleanup log


Huangxiao-mas:
### ODC version

ODC430bp

### OB version

Tables outside the required range appear in the archive cleanup log

### What happened?

。

### What did you expect to happen?

。

### How can we reproduce it (as minimally and precisely as possible)?

。

PR #3014, opened 2024-07-25
fix(dlm): don't compare the table structure if syncTableStructure is off

### LINKED ISSUE #2844: [Bug]: Data archiving fails, but the cleanup task is still executed
### ODC version

ODC430bp

### OB version

independent

### What happened?

Data archiving fails, but the cleanup task is still executed

。

。

_No response_

guowl3:
#### What type of PR is this?
type-bug

1. Don't compare the table structure if syncTableStructure is off.
2. Sync table structure only supports MySQL and OMySQL. It throws an unsupported exception if other databases are chosen.

Fixes #2844 #2845 

```docs

```


yizhouxw (approved):
lgtm

yhilmare (approved):
lgtm

PR #3029, opened 2024-07-29
feat(migrate): create built-in project for migrate and add hook to execute pre logic

yiminpeng:

1. create built-in project for migrator.
2. delete some yaml when ODC version less than 4.3.0.

Fixes #


server/odc-core/src/main/java/com/oceanbase/odc/core/migrate/Migrates.java
```diff
@@ -85,7 +88,15 @@ public void migrate() {
                 version2Migratables.keySet().stream().map(Version::new).sorted().collect(Collectors.toList());
         log.info("versionCount={}", sortedVersions.size());
 
-        degradeCheck(sortedVersions.get(sortedVersions.size() - 1).getVersion());
+        String currentVersion = sortedVersions.get(sortedVersions.size() - 1).getVersion();
+
+        if (VersionUtils.isLessThan(configuration.getInitVersion(), "4.3.0.0")) {
```
MarkPotato777, line 93: This logic should be moved to the apsara implementation.

server/odc-core/src/main/java/com/oceanbase/odc/core/migrate/BeforeCheckDeleteHook.java
```diff
@@ -0,0 +1,29 @@
...
+package com.oceanbase.odc.core.migrate;
+
+import javax.sql.DataSource;
+
+/**
+ * @author yiminpeng
+ * @date 2024/07/26
+ * @version : BeforeCheckDeleteHook.java, v 1.0
+ */
+public interface BeforeCheckDeleteHook {
```
MarkPotato777, line 25: It is unnecessary to limit it to `delete`. `MigratePreHook` is better?

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/auth/CustomAuthenticationSuccessHandler.java
```diff
@@ -124,7 +125,12 @@ public void onAuthenticationSuccess(HttpServletRequest httpServletRequest, HttpS
             SecurityContextUtils.switchCurrentUserOrganization(user, team, httpServletRequest, true);
             // If bastion is enabled, every user must hold a built-in project for create temporary SQL console
             if (bastionEnabled) {
-                projectService.createProjectIfNotExists(user);
+                ProjectEntity project = projectService
+                        .createProjectIfNotExists(user,
+                                "USER_PROJECT_" + user.getAccountName(),
```
MarkPotato777, line 130: Is the original project name `'USER_PROJECT_xxx' + user.getAccountName()`?

server/odc-core/src/main/java/com/oceanbase/odc/core/migrate/BeforeCheckDeleteHook.java
yiminpeng: Yes, considering expandability, deletion only needs to be reflected in the name of the method.

server/odc-core/src/main/java/com/oceanbase/odc/core/migrate/Migrates.java
yiminpeng: got it, done.

server/odc-core/src/main/java/com/oceanbase/odc/core/migrate/BeforeCheckDeleteHook.java
yiminpeng: Yes, considering expandability, deletion only needs to be reflected in the name of the method.

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/auth/CustomAuthenticationSuccessHandler.java
yiminpeng: Yes, `USER_PROJECT_` was as a constant.

MarkPotato777 (approved):
LGTM

Issue #3032, opened 2024-07-29
[Bug]: When enabling or disabling SQL plans, the interface reports "the parameter triggerConfig value cannot be empty"


Huangxiao-mas:
### ODC version

ODC431

### OB version

independent

### What happened?

When enabling or disabling SQL plans, the interface reports "the parameter triggerConfig value cannot be empty"
![image](https://github.com/user-attachments/assets/9a20343e-132d-4cac-be1f-1d07acc4d56c)
![image](https://github.com/user-attachments/assets/d5e7a894-f1a6-4e76-84d0-637dbe8a026b)

### What did you expect to happen?

。

### How can we reproduce it (as minimally and precisely as possible)?

Enable or disable SQL plan

Issue #3027, opened 2024-07-26
[Bug]: After editing the data cleaning task, the target data is displayed as empty when viewing the details


Huangxiao-mas:
### ODC version

ODC431

### OB version

independent

### What happened?

After editing the data cleaning task, the target data is displayed as empty when viewing the details
![image](https://github.com/user-attachments/assets/610676ed-416f-4e4c-9970-0524c6ef5bc2)

### What did you expect to happen?

。

### How can we reproduce it (as minimally and precisely as possible)?

1. Create scheduled tasks
2. Edit tasks
3. Save
4. View tasks

Issue #3024, opened 2024-07-26
[Bug]: After editing the task, a new task was initiated


Huangxiao-mas:
### ODC version

ODC431

### OB version

independent

### What happened?

After editing the task, a new task was initiated
![image](https://github.com/user-attachments/assets/dd88c606-ec18-4d21-86c8-7a07f5e15014)

### What did you expect to happen?

No new tasks will be initiated

### How can we reproduce it (as minimally and precisely as possible)?

Create a periodic task, disable a task
Edit a task
Found a new task initiated

PR #3033, opened 2024-07-29
fix(dlm):  several bug related to editing data cleaning

### LINKED ISSUE #3022: [Bug]: The task can only be edited if it is disabled.
### ODC version

ODC431

### OB version

independent

### What happened?

The task can only be edited if it is disabled.
image

The task can only be edited if it is disabled.

1. Create a periodic task
2. Edit the task
3. Found that it can be edited

_No response_

guowl3:
#### What type of PR is this?
type-bug

Fixes #3022 #3027 #3024 #3032 

```docs

```


yhilmare (approved):
lgtm

PR #3035, opened 2024-07-29
fix(table-permission): not select the specific database/table by default when create the permission application ticket

### LINKED ISSUE #2912: [Bug]: There is no default selection of items for application form permissions in the SQL window
### ODC version

odc431

### OB version

*

### What happened?

image
image

image

Low permission accounts execute query statements without permission tables in SQL windows

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug

Add the projected in the response so that the front-end could select the specific database/table when apply the permissions.

Fixes #2912 

```docs

```


yhilmare (approved):
lgtm

Issue #2957, opened 2024-07-11
[Bug]: The data source creates a database and reports that Column 'type' cannot be null


Huangxiao-mas:
### ODC version

ODC431

### OB version

independent

### What happened?

The data source creates a database and reports that Column 'type' cannot be null
![image](https://github.com/oceanbase/odc/assets/59468957/915f83a9-dca3-40c3-8ede-5b3e5f910f01)

### What did you expect to happen?

。

### How can we reproduce it (as minimally and precisely as possible)?

。

PR #3037, opened 2024-07-29
fix(database): creating databases under the data source failed

### LINKED ISSUE #2913: [Bug]: Creating a database in the data source, reporting that column does not exist
### ODC version

ODC431

### OB version

independent

### What happened?

Creating a database in the data source, reporting that column does not exist
image

。

。

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug

Create the database under the data source failed. The root cause is that the `connect_database` ddl changed in this version, but the code did not set the `type` and the `dialect_type` when creating the database.

Fixes #2913 #2957 

```docs

```


yhilmare (approved):
lgtm

PR #3039, opened 2024-07-29
fix(partition-plan): can not recognize the partition key's data type on mysql mode

### LINKED ISSUE #3016: [Bug]: can not recognize the partition key's data type on mysql mode
### ODC version

4.3.1

### OB version

not related 

### What happened?

1. create a partition table:
```sql
create table test_tbl1(
    id bigint,
    CREATE_TIME datetime
) partition by range columns(create_time) (
    partition p202407 values less than('2024-07-25 00:00:00')
);
```

as you can see the partition key reference a column but the case is different which is valid for mysql mode

2. attempt to start up a partition plan
截屏2024-07-25 15 47 54

as you can see, the datatype is wrong for the table.

recognize the datatype correctly~

follow what happened

_No response_

guowl3:
#### What type of PR is this?
type-bug

Fixes #3016 

```docs

```

### DIFF
--- a/server/plugins/task-plugin-ob-mysql/src/main/java/com/oceanbase/odc/plugin/task/obmysql/partitionplan/datatype/BasePartitionKeyDataTypeFactory.java
+++ b/server/plugins/task-plugin-ob-mysql/src/main/java/com/oceanbase/odc/plugin/task/obmysql/partitionplan/datatype/BasePartitionKeyDataTypeFactory.java
@@ -70,8 +70,8 @@ public abstract class BasePartitionKeyDataTypeFactory implements DataTypeFactory
          * assume that the partition key is a column
          */
         Map<String, DBTableColumn> colName2Col = this.dbTable.getColumns().stream()
-                .collect(Collectors.toMap(c -> unquoteIdentifier(c.getName()).toLowerCase(), c -> c));
-        DBTableColumn column = colName2Col.get(unquotedPartitionKey.toLowerCase());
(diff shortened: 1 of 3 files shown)


server/plugins/task-plugin-ob-mysql/src/main/java/com/oceanbase/odc/plugin/task/obmysql/partitionplan/datatype/BasePartitionKeyDataTypeFactory.java
```diff
@@ -70,8 +70,8 @@ public DataType generate() {
          * assume that the partition key is a column
          */
         Map<String, DBTableColumn> colName2Col = this.dbTable.getColumns().stream()
-                .collect(Collectors.toMap(c -> unquoteIdentifier(c.getName()), c -> c));
-        DBTableColumn column = colName2Col.get(unquotedPartitionKey);
+                .collect(Collectors.toMap(c -> unquoteIdentifier(c.getName()).toLowerCase(), c -> c));
```
yhilmare, line 73: we may modify `OBMySQLAutoPartitionExtensionPoint#unquoteIdentifier` directly to fix this~

yizhouxw: Collectors.toMap may change to use TreeMap with CASE_INSENSITIVE Comparator

yhilmare (approved):
lgtm

Issue #3019, opened 2024-07-26
[Bug]: When only having the permission to change a single table, a new table can be created by importing


sl01388797:
### ODC version

odc431

### OB version

*

### What happened?

![image](https://github.com/user-attachments/assets/3e4347e8-af98-49ff-809f-3021fccfe5e8)

### What did you expect to happen?

Only with database change permission can files be imported into the database

### How can we reproduce it (as minimally and precisely as possible)?

Enable change permission for a table in the database and then import a table into this database

sl01388797:
![Image](https://github.com/user-attachments/assets/19736f73-929c-47cc-8a91-cc3a6b363334)

pass

Issue #3015, opened 2024-07-25
[Bug]: Data archiving and data cleaning can select tables without permission


sl01388797:
### ODC version

odc431

### OB version

*

### What happened?

![image](https://github.com/user-attachments/assets/86ad936a-07f2-42e7-be33-4d23f46ebceb)
![image](https://github.com/user-attachments/assets/68076a98-fce4-4a87-b63f-572e9365844f)

### What did you expect to happen?

Do not display tables without permission

### How can we reproduce it (as minimally and precisely as possible)?

1. Application Form for Change of Permissions for Low Permission Accounts
2. Create data archive, data cleaning

yezaoshu:
This may require adjustments to the authorizedPermissionTypes field, or waiting for further processing solutions

sl01388797:
![Image](https://github.com/user-attachments/assets/b511c35c-acc2-4c64-a31e-cf928c96fd9e)

pass

Issue #3013, opened 2024-07-25
[Bug]: Partition plans can still be executed when tables do not have permission to change


sl01388797:
### ODC version

odc431

### OB version

*

### What happened?

![image](https://github.com/user-attachments/assets/7e0df4cf-0311-43a1-9c98-d34dd5edc657)
![image](https://github.com/user-attachments/assets/afd5ffe8-8d0e-49e4-a253-c0586b630eae)

### What did you expect to happen?

Do not display tables without permission

### How can we reproduce it (as minimally and precisely as possible)?

1. Create a range partition table
2. Apply for permission to change the external appearance of the range table database using a low privilege account
3. Execute partition plan

yezaoshu:
This may require adjustments to the authorizedPermissionTypes field, or waiting for further processing solutions

sl01388797:
![Image](https://github.com/user-attachments/assets/01aa5443-0d58-4d47-921a-2588046ee239)

pass

Issue #3011, opened 2024-07-25
[Bug]: Structural comparison can select tables without permission for comparison


sl01388797:
### ODC version

odc431

### OB version

*

### What happened?

![image](https://github.com/user-attachments/assets/3435a5c3-9831-4e69-ae9f-78c053ddacc4)

### What did you expect to happen?

Cannot select tables without change permission

### How can we reproduce it (as minimally and precisely as possible)?

1. Give a table change permission
2. Select a table without change permission when performing structure comparison

yezaoshu:
This may require adjustments to the authorizedPermissionTypes field, or waiting for further processing solutions

sl01388797:
![Image](https://github.com/user-attachments/assets/cc8a39bb-d0c0-49fa-a5d4-9f5d32e385d9)

pass

PR #3046, opened 2024-07-29
fix(table-permission): could create tickets when users have no permission to the database

### LINKED ISSUE #3010: [Bug]: Tables that do not have permission to simulate data can also be executed
### ODC version

odc431

### OB version

*

### What happened?

image
image
image

Unauthorized tables cannot perform tasks

1. Give a table change permission
2. Select tables without change permission to execute database simulation

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug

rollback code of the PR 2946

Fixes #3010 #3019 #3011 #3013 #3015 

```docs

```


yhilmare (approved):
lgtm

PR #3069, opened 2024-07-31
fix(parser): failed to recognize the schema or package name from an anonymous block

### LINKED ISSUE #3064: [Bug]:  access control not valid in anonymous sql block
### ODC version

4.2.4

### OB version

4.x oracle mode

### What happened?

current env:
    database: aps
    user access authority: no

in anonymous blocks:
     execute sql:
```sql
     declare
        i number;
    begin
      update aps.base_logs t set t.last_update_time = sysdata where t.id = 1;
       commit;
    end;
```
sql expected to be blocked cause not access authority granted to aps database

anonymous sql block should also check access authority

current env:
    database: aps
    user access authority: no

in sql console
execute sql:
```
    update aps.base_logs t set t.last_update_time = sysdata where t.id = 1
```
result:
     sql was blocked cause write authority not granted

in anonymous blocks:
     execute sql:
```sql
     declare
        i number;
    begin
      update aps.base_logs t set t.last_update_time = sysdata where t.id = 1;
       commit;
    end;
```
sql expected to be blocked as sql console does but was execute successful

_No response_

yhilmare:
#### What type of PR is this?

type-bug

failed to recognize the schema or package name from an anonymous block

Fixes #3064

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/DBSchemaExtractor.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/DBSchemaExtractor.java
@@ -61,6 +61,7 @@ import com.oceanbase.tools.sqlparser.oboracle.OBParser.Routine_nameContext;
 import com.oceanbase.tools.sqlparser.oboracle.OBParser.Var_nameContext;
 import com.oceanbase.tools.sqlparser.oboracle.PLParser.IdentifierContext;
 import com.oceanbase.tools.sqlparser.oboracle.PLParser.Pl_schema_nameContext;
+import com.oceanbase.tools.sqlparser.oboracle.PLParser.Sql_stmtContext;
 import com.oceanbase.tools.sqlparser.statement.Expression;
 import com.oceanbase.tools.sqlparser.statement.common.RelationFactor;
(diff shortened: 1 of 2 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/DBSchemaExtractor.java
```diff
@@ -420,6 +421,18 @@ public RelationFactor visitPl_schema_name(Pl_schema_nameContext ctx) {
             return null;
         }
 
+        @Override
+        public RelationFactor visitSql_stmt(Sql_stmtContext ctx) {
+            try {
+                OBOracleRelationFactorVisitor visitor = new OBOracleRelationFactorVisitor();
+                visitor.visit(ctx.getChild(0).getChild(0));
```
LioRoger, line 428: is there any possible that ctx.getChild(0) is null or ctx.getChild() is zero length?

yhilmare: yes, you are right, I have changed it~

yizhouxw (approved):
lgtm

PR #3072, opened 2024-07-31
fix(flow): wrong approval flow for database/table permission apply ticket

### LINKED ISSUE #3068: [Bug]: If the risk level is set to [Apply for library permission] and [Environment], the approval process is invalid
### ODC version

ODC 4.3.0-bp

### OB version

No related.

### What happened?

If the risk level is set to [Apply for library permission] and [Environment], the approval process is invalid.
image
The approval process is automatic.
image

The approval process should be effective.

Follow what happened.

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug
As database/table permission ticket may relate to more than one database or environment, so the approval flow is not accurately selected. This PR will always select the highest risk level approval flow for these two kinds of tickets.

Fixes #3068 

```docs

```


yhilmare (approved):
lgtm

PR #3086, opened 2024-08-02
fix(osc): enhance osc feature

LioRoger:
…witch table, fix thread leak issue when relase oms projcet

#### What type of PR is this?
type-bug

1.support lock table for ob-mysql 4.2.5
2.fix data loss issue when switch table,
3.fix thread leak issue when release OMS project
4.support special schema/table

Fixes #
1. support lock table for ob-mysql with version greater than 4.2.5
2. cause  the missing check of data apply progress，data may loss during table switch,. Now we will wait until increment checkpoint in timestamp greater than the timestamp recorded after lock table. Notice, this waiting may take about 10 - 20 seconds now.  We will optimize this lock table waiting time in future release.
3. fix thread leak when release OMS project. change to sync call instead of create a new Thread pool.
4. add escape character to MySQL/Oracle schema/table name

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/oms/OmsRequestUtil.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/oms/OmsRequestUtil.java
@@ -1,0 +1,117 @@
+/*
+ * Copyright (c) 2023 OceanBase.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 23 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/oms/OmsRequestUtil.java
```diff
@@ -0,0 +1,117 @@
...
+     * @return
+     */
+    public static OmsProjectControlRequest getProjectRequest(String uid, String projectID) {
+        OmsProjectControlRequest projectRequest = new OmsProjectControlRequest();
+        projectRequest.setUid(uid);
+        projectRequest.setId(projectID);
+        return projectRequest;
+    }
+
+    public static boolean OMSTaskReady(ProjectStepResult projectStepResult) {
```
LuckyPickleZZ, line 55: maybe `isOMSTaskReady`?

Issue #2203, opened 2024-04-13
[Bug]: show history log when odc upgrade use same mount path but with a new meta db


krihy:
### ODC version

4.2.4

### OB version

any

### What happened?

show history log when odc upgrade use same mount path but with a new meta db
![image](https://github.com/oceanbase/odc/assets/51393259/81b7d261-fab8-4d08-8bd9-e8ecd499c930)

### What did you expect to happen?

show latest log but do not show early log

### How can we reproduce it (as minimally and precisely as possible)?

 upgrade use same mount path but with a new meta db

PR #3088, opened 2024-08-05
fix(task):  show history log when odc upgrade use same mount path but with a new meta db

CHLK:
What type of PR is this?
type-bug

triggering task log rollover when task start up, and the history log will not show in new task log.

Which issue(s) this PR fixes:
Fixes https://github.com/oceanbase/odc/issues/2203


yizhouxw (approved):
lgtm

Issue #3094, opened 2024-08-05
[Bug]: When only one table has export permission, the entire database export is successful


sl01388797:
### ODC version

odc431

### OB version

*

### What happened?

![image](https://github.com/user-attachments/assets/0a442b72-7fd0-4ca3-866c-324a9d4f4ddf)
![image](https://github.com/user-attachments/assets/b2a7145d-d1ea-4fbd-90cb-f18a302b381a)

### What did you expect to happen?

Export failed

### How can we reproduce it (as minimally and precisely as possible)?

1. Create a table with export permissions
2. Select the database for whole database export

sl01388797:
![Image](https://github.com/user-attachments/assets/7d68c08a-a8ea-4b39-8a1a-dc1858a77edb)

pass

PR #3101, opened 2024-08-06
fix(permission): 1) users could export the whole database even if they don't have database export permissions 2) SQL contains function call will be wrongly intercepted

### LINKED ISSUE #3091: [Bug]: Tables with change permissions cannot add large objects
### ODC version

odc431

### OB version

*

### What happened?

image

Can add large objects

1. Create a table with change permission for data types and large objects
2. Add data on white screen

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug
This PR fixed 2 issues:
- Exporting the whole database from the database resource tree on the left side of the sql console will not be intercepted if users only have table export permissions rather than database export permissions. This is because, in this scene, the front-end will pass the `exportAllObjects` and `exportDbObjects` at the same time. However, the back-end will first judge the `exportDbObjects`.
- When SQL contains a function call, it will be intercepted by notifying that 'need database permissions' even if users have that table permission. This is because the backend think the function call is operating the database itself, even if it may just a built-in function like `count()`, `now()`. The solution is that we passby all the function calls.

Fixes #3091 #3094 

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
@@ -779,7 +779,12 @@ public class FlowInstanceService {
         if (req.getTaskType() == TaskType.EXPORT) {
             DataTransferConfig parameters = (DataTransferConfig) req.getParameters();
             Map<DBResource, Set<DatabasePermissionType>> resource2Types = new HashMap<>();
-            if (CollectionUtils.isNotEmpty(parameters.getExportDbObjects())) {
+            if (parameters.isExportAllObjects()) {
+                ConnectionConfig config = connectionService.getBasicWithoutPermissionCheck(req.getConnectionId());
+                resource2Types.put(
+                        DBResource.from(config, req.getDatabaseName(), null, ResourceType.ODC_DATABASE),
+                        DatabasePermissionType.from(TaskType.EXPORT));
+            } else if (CollectionUtils.isNotEmpty(parameters.getExportDbObjects())) {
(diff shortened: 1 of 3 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
```diff
@@ -779,7 +779,12 @@ private void checkCreateFlowInstancePermission(CreateFlowInstanceReq req) {
         if (req.getTaskType() == TaskType.EXPORT) {
             DataTransferConfig parameters = (DataTransferConfig) req.getParameters();
             Map<DBResource, Set<DatabasePermissionType>> resource2Types = new HashMap<>();
-            if (CollectionUtils.isNotEmpty(parameters.getExportDbObjects())) {
+            if (parameters.isExportAllObjects()) {
```
LuckyPickleZZ, line 782: Will this mislead users into thinking that exporting the entire database has been successful? I tend to throw exceptions when do not have export dtabase permission. 
What are the opinions of others?

MarkPotato777: If users don't have database export permissions but try to export the whole database, the exception will be thrown. This PR just change the order of judging the `isExportAllObjects` as you could see in the PR description.

LuckyPickleZZ (approved):
lgtm

yizhouxw (commented):
PR title should be issue scenarios

MarkPotato777:
> PR title should be issue scenarios

done

yhilmare (approved):
lgtm

PR #3106, opened 2024-08-06
feat(osc): refactor online schame change module, introduce fms package

LioRoger:
#### What type of PR is this?
type-feature

1. refactor osc module, introduce finite state machine to drive the work flow
2. split complex logic into simple step, make code testable
3. prepare for alternative implement, eg ghost

Fixes #
1. introduce fms module
2. split logic in several state. (defined in OscStates)
3. each state correspond to a action implment
4. action is designed reentrant 

```docs

```


LuckyPickleZZ (dismissed):

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/fsm/Action.java
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Action that do real job in finite state machine
+ * 
+ * @param <Context> input context for execute
+ * @param <ActionResult> function result of execute
+ * @author longpeng.zlp
+ * @date 2024/7/5 17:24
+ * @since 4.3.1
+ */
+public interface Action<Context extends ActionContext, ActionResult> {
```
yizhouxw, line 27: the fsm implementation looks like for common scenario, how about move to odc-common module ?

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/oms/client/BaseOmsClient.java
```diff
@@ -100,6 +100,7 @@ private <T> OmsApiReturnResult<T> resolveResponseEntity(ClientRequestParams requ
                     : ErrorCodes.BadRequest;
             throw new OmsException(errorCode, responseEntity.toString(), null, responseEntity.getStatusCode());
         }
+        log.info("process oms request [{}] with response [{}]", requestParams, responseEntity.getBody());
```
yizhouxw, line 103: odc use k=v style for log output,
such as log.info("process oms request, request={}, response={}", xxx, xxx)

yizhouxw (commented):
perfect refactoring! and the integration test!

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/fsm/Action.java
LioRoger: currently, it's only used by online schema change scene。 it's need a step driver now implement with QZ. 
It may will be moved to odc-common module if there is similar situation。 So I prefer take it as a private module.

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/oms/client/BaseOmsClient.java
LioRoger: corrected

yizhouxw (approved):

LuckyPickleZZ (approved):

PR #3114, opened 2024-08-08
fix(notification): support response type `text/plain` and others

### LINKED ISSUE #3113: [Bug]: when content type of reponse is text/plain, the notification would fail
### ODC version

4.2+

### OB version

-

### What happened?

when content type of reponse is text/plain, the notification would fail.

ODC should process response correctlly.

Send a message to a webhook with a response format of text/plain

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

In the previous version, when ODC were using `RestTemplate#Exchange`, we specified the class as MAP, which means we can only handle responses in the `applicaton/json` format.

Fixes #3113

```docs

```


yhilmare (approved):
lgtm

MarkPotato777 (approved):
LGTM

PR #3116, opened 2024-08-09
feat(git-integration): add service layer for git repository integration

LuckyPickleZZ:

Fixes #

```docs

```

### DIFF
--- a/server/odc-core/src/main/resources/i18n/BusinessMessages_zh_CN.properties
+++ b/server/odc-core/src/main/resources/i18n/BusinessMessages_zh_CN.properties
@@ -82,6 +82,8 @@ com.oceanbase.odc.ResourceType.ODC_TEAM_ORGANIZATION=团队空间
 com.oceanbase.odc.ResourceType.ODC_NOTIFICATION_CHANNEL=通知通道
 com.oceanbase.odc.ResourceType.ODC_NOTIFICATION_POLICY=通知规则
 com.oceanbase.odc.ResourceType.ODC_NOTIFICATION_MESSAGE=通知消息
+com.oceanbase.odc.ResourceTppe.ODC_GIT_REPOSITORY=Git 存储库
+
 com.oceanbase.odc.ResourceType.ODC_STRUCTURE_COMPARISON_TASK=结构对比任务
 com.oceanbase.odc.ResourceType.ODC_DATABASE_CHANGE_ORDER_TEMPLATE=数据库变更顺序模板
 
--- a/server/odc-core/src/main/resources/i18n/BusinessMessages_zh_CN.properties
+++ b/server/odc-core/src/main/resources/i18n/BusinessMessages_zh_CN.properties
@@ -82,7 +82,7 @@ com.oceanbase.odc.ResourceType.ODC_TEAM_ORGANIZATION=团队空间
 com.oceanbase.odc.ResourceType.ODC_NOTIFICATION_CHANNEL=通知通道
 com.oceanbase.odc.ResourceType.ODC_NOTIFICATION_POLICY=通知规则
 com.oceanbase.odc.ResourceType.ODC_NOTIFICATION_MESSAGE=通知消息
-com.oceanbase.odc.ResourceTppe.ODC_GIT_REPOSITORY=Git 仓库
+com.oceanbase.odc.ResourceType.ODC_GIT_REPOSITORY=Git 仓库
 
 com.oceanbase.odc.ResourceType.ODC_STRUCTURE_COMPARISON_TASK=结构对比任务
 com.oceanbase.odc.ResourceType.ODC_DATABASE_CHANGE_ORDER_TEMPLATE=数据库变更顺序模板
--- a/server/odc-core/src/main/resources/i18n/BusinessMessages_zh_CN.properties
+++ b/server/odc-core/src/main/resources/i18n/BusinessMessages_zh_CN.properties
@@ -82,7 +82,7 @@ com.oceanbase.odc.ResourceType.ODC_TEAM_ORGANIZATION=团队空间
 com.oceanbase.odc.ResourceType.ODC_NOTIFICATION_CHANNEL=通知通道
 com.oceanbase.odc.ResourceType.ODC_NOTIFICATION_POLICY=通知规则
 com.oceanbase.odc.ResourceType.ODC_NOTIFICATION_MESSAGE=通知消息
-com.oceanbase.odc.ResourceTppe.ODC_GIT_REPOSITORY=Git 存储库
+com.oceanbase.odc.ResourceTppe.ODC_GIT_REPOSITORY=Git 仓库
 
 com.oceanbase.odc.ResourceType.ODC_STRUCTURE_COMPARISON_TASK=结构对比任务
 com.oceanbase.odc.ResourceType.ODC_DATABASE_CHANGE_ORDER_TEMPLATE=数据库变更顺序模板
--- a/server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/GitIntegrationController.java
+++ b/server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/GitIntegrationController.java
@@ -62,9 +62,10 @@ public class GitIntegrationController {
         return Responses.paginated(integrationService.list(projectId, params, pageable));
     }
 
-    @ApiOperation(value = "listCandidates", notes = "List all candidate repos from vsc provider")
-    @RequestMapping(value = "/candidates", method = RequestMethod.POST)
-    public List<GitRepository> listCandidates(@PathVariable Long projectId, @RequestBody GitProvider provider) {
+    @ApiOperation(value = "listCandidateRepositories", notes = "List all candidate repos from vsc provider")
+    @RequestMapping(value = "/listCandidateRepositories", method = RequestMethod.POST)
+    public List<GitRepository> listCandidateRepositories(@PathVariable Long projectId,
+            @RequestBody GitProvider provider) {
         return integrationService.listCandidates(provider);
     }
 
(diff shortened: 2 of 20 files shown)


server/odc-core/src/main/resources/i18n/BusinessMessages_zh_CN.properties
```diff
@@ -82,6 +82,8 @@ com.oceanbase.odc.ResourceType.ODC_TEAM_ORGANIZATION=团队空间
 com.oceanbase.odc.ResourceType.ODC_NOTIFICATION_CHANNEL=通知通道
 com.oceanbase.odc.ResourceType.ODC_NOTIFICATION_POLICY=通知规则
 com.oceanbase.odc.ResourceType.ODC_NOTIFICATION_MESSAGE=通知消息
+com.oceanbase.odc.ResourceTppe.ODC_GIT_REPOSITORY=Git 存储库
```
yhilmare, line 85: `Git 仓库` will be better?

server/odc-server/src/main/resources/data.sql
```diff
@@ -840,3 +840,8 @@ VALUES ('odc.permission-check.whitelist.database.ob-oracle',
             'dbms_trusted_certificate_manager, dbms_workload_repository, dbms_xplan',
         'Permission check whitelist for database in OB Oracle dialect type')
 ON DUPLICATE KEY UPDATE `id`=`id`;
+
+INSERT INTO config_system_configuration(`key`, `value`, `description`) VALUES('git.repository.preserve.minutes',
```
yhilmare, line 844: `odc.integration.git.repository.preserve.minutes`
we may add a prefix~, same as below

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/GitIntegrationController.java
```diff
@@ -0,0 +1,91 @@
...
+        QueryGitRepositoryParams params = QueryGitRepositoryParams.builder()
+                .fuzzyName(fuzzyName)
+                .providers(providers)
+                .projectId(projectId)
+                .build();
+        return Responses.paginated(integrationService.list(projectId, params, pageable));
+    }
+
+    @ApiOperation(value = "listCandidates", notes = "List all candidate repos from vsc provider")
+    @RequestMapping(value = "/candidates", method = RequestMethod.POST)
```
MarkPotato777, line 66: How should we understand `candidates` ?

server/odc-service/src/main/java/com/oceanbase/odc/service/git/GitIntegrationService.java
```diff
@@ -0,0 +1,161 @@
...
+        List<GitRepositoryEntity> entities = repositories.stream()
+                .map(r -> {
+                    GitRepositoryEntity entity = modelToEntity(r);
+                    entity.setProjectId(projectId);
+                    entity.setCreatorId(creatorId);
+                    entity.setOrganizationId(organizationId);
+                    return entity;
+                })
+                .collect(Collectors.toList());
+        List<GitRepositoryEntity> saved = gitRepoRepository.saveAll(entities);
```
yhilmare, line 105: saveAll is not a batch operation
we should figure out another way if there are too many repositories~

server/odc-server/src/main/resources/data.sql
MarkPotato777, line 844: We better add `odc.` prefix in the config key. How about `odc.integration.git.repository-retention-minutes`?

server/odc-service/src/main/java/com/oceanbase/odc/service/git/repo/RepositoryManager.java
```diff
@@ -0,0 +1,212 @@
...
+                        ((GitClientOperator) operator).close();
+                    }
+                })
+                .build(this::createGitOperator);
+    }
+
+    @PostConstruct
+    public void init() {
+        try {
+            FileUtils.cleanDirectory(new File(GIT_REPO_DIR));
```
MarkPotato777, line 105: Why is it necessary to clean this directory every time ODC restarts? Is this directory intended to be a temporary directory?

LuckyPickleZZ: yes. It's a temporary dir. User's repository would be removed regularly. 
And the remove operations were managed in memory. So if we do not clean the directory when ODC restarts, there maybe some

MarkPotato777: But what if the ODC never restarts or restarts very infrequently? Shouldn't we have a scheduled cleanup mechanism instead of relying on the ODC restart to clean the temporary directory?

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/GitIntegrationController.java
```diff
@@ -0,0 +1,91 @@
...
+import com.oceanbase.odc.service.git.model.VcsProvider;
+
+import io.swagger.annotations.ApiOperation;
+
+/**
+ * @author: liuyizhuo.lyz
+ * @date: 2024/8/9
+ */
+@RestController
+@RequestMapping("/api/v2/project/{projectId}/integration/repos")
```
yizhouxw, line 45: url style, {module}/resources/{id}/subresources/

-->

/api/v2/projects/{projectId}/gitRepos/


emm, what is the module name, collaboration？

server/odc-service/src/main/java/com/oceanbase/odc/service/git/repo/RepositoryManager.java
```diff
@@ -0,0 +1,212 @@
...
+
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * @author: liuyizhuo.lyz
+ * @date: 2024/8/7
+ */
+@Slf4j
+@Component
+public class RepositoryManager {
```
yizhouxw, line 63: GitRepositoryManager

server/odc-core/src/main/resources/i18n/BusinessMessages_zh_CN.properties
LuckyPickleZZ: sure, updated

server/odc-server/src/main/resources/data.sql
LuckyPickleZZ: done

server/odc-service/src/main/java/com/oceanbase/odc/service/git/GitIntegrationService.java
LuckyPickleZZ: sure,  used `GitRepoRepository#batchCreate` to replace this .

server/odc-server/src/main/resources/data.sql
LuckyPickleZZ: done

server/odc-service/src/main/java/com/oceanbase/odc/service/git/repo/RepositoryManager.java
LuckyPickleZZ: I dont think clean up periodically is a good choice. A one-time cleaning is simple, stable, and meets the needs of our business. 
Even if ODC restarts frequently, the cost of cleaning temporary folders is not very high. What are your specific concerns about this place?

server/odc-service/src/main/java/com/oceanbase/odc/service/git/repo/RepositoryManager.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    private GitClientOperator createGitOperator(Pair<Long, Long> pair) throws IOException, GitAPIException {
+        File repo = getRepoDir(pair.left, pair.right);
+        FileUtils.forceMkdir(repo);
+        String[] children = repo.list();
+        if (children.length == 0) {
+            return initRepo(repo, pair.left, pair.right);
+        }
+        Verify.equals(1, children.length, "found more than one repositories");
```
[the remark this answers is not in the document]
LuckyPickleZZ: I modified this into : if the `children.lenth > 0` , we would clean up this directory.
Normally, it would not follow this logic. This is for safety reasons

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/GitIntegrationController.java
LuckyPickleZZ: I changed it to `/api/v2/collaboration/projects/{projectId}/repos`.
How about this?

server/odc-service/src/main/java/com/oceanbase/odc/service/git/repo/RepositoryManager.java
LuckyPickleZZ: done

server/odc-service/src/main/java/com/oceanbase/odc/service/git/repo/RepositoryManager.java
MarkPotato777: I mean **infrequently**, not **frequently**. If your purpose is to clean up the directory to ensure that some unnecessary folders are eventually deleted, then your 'one-time cleaning' approach cannot guarantee this. Conversely, if you don't care about the temp directory, why do we need this cleanup operation at all?

LuckyPickleZZ: ... Sorry I misread. Then this question is simple. Each GitOperator in the cache will expire after 24 hours. At the same time as this expires, the workspace directory will be cleaned.
So you can think of it as regular cleaning

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/GitIntegrationController.java
yizhouxw: not repos, but gitRepos

LuckyPickleZZ: okay, done.

yizhouxw (approved):

yhilmare (approved):
lgtm

PR #3118, opened 2024-08-12
fix(notification): ClassCastException would be thrown when sending to feishu or wecom

### LINKED ISSUE #3117: [Bug]: Unable to create WeChat Enterprise message channel
### ODC version

ODC431

### OB version

independent

### What happened?

Unable to create WeChat Enterprise message channel
image

。

chat priavtely

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

In https://github.com/oceanbase/odc/pull/3114 , we changed the reponse extract type from `Map` to `String`. But missed to update it in `FeishuSender` and `WeComSender`. 

Fixes #3117

```docs

```


yhilmare (approved):
lgtm

PR #3123, opened 2024-08-14
feat(osc):enhance osc, let swap table reentrant

LioRoger:
#### What type of PR is this?
type-feature

make osc swap table reentrant for future error recovery
connection will be recreated when retry swap table

Fixes #
test if swap table has finished before required lock. if ghost table not exist, then consider swap table has done
1. if swap table has not done, do swap table action
2. if swap table has done, ignore swap table
3. if drop old table enabled, check if table has been dropped to avoid unexpected exception

integrate test has made up for above situation

```docs

```


yizhouxw (approved):
lgtm

LuckyPickleZZ (approved):
lgtm

PR #3124, opened 2024-08-14
fix(connection): choosing the wrong mode when testing connection with sys tenant result in false success

### LINKED ISSUE #1462: [Bug]: sys tenant connection ob oracle mode test connection successful
### ODC version

4.2.3

### OB version

all

### What happened?

sys tenant connection ob oralce mode test connection successful.
image
After creating a new data source with sys tenant account, the data source is grayed out.
image

sys tenant connection ob oralce mode test connection successful.

Create a new data source and select ob oracle mode. Testing the connection with the sys tenant should return connection failure information. odc 4.1.3 works fine：

_No response_

### Cloud

_No response_

abigsaltyfish:
**What type of PR is this?**
type-bug

**What this PR does / why we need it:**
In version 4.1.3, incorrect schema settings can result in schema information values being set to username and receiving an unknown database. I restored it in the scenario where the sys tenant mistakenly selected Oracle mode.

Which issue(s) this PR fixes:
Fixes #1462


yizhouxw:
the bugfix PR title should be problem scenario view, instead of repeat 'fix xxx'

yizhouxw:
please select project and realted fields, and milestone
image

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionTesting.java
```diff
@@ -214,16 +214,21 @@ private ConnectionConfig reqToConnectionConfig(TestConnectionReq req) {
         ConnectionConfig config = new ConnectionConfig();
         if (StringUtils.equals("sys", req.getTenantName())) {
             config.setType(ConnectType.OB_MYSQL);
+            if (req.getType() != ConnectType.OB_MYSQL) {
+                config.setDefaultSchema("ROOT");
```
yizhouxw, line 218: why use 'ROOT' as default schema ?

yizhouxw (commented):
there exists 2 scenarios for sys tenant connection test

1. user connect to sys tenant straightly
2. user connect to user tenant, but also configure sys tenant account/password,  odc will test connection for both user tenant and sys tenant

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionTestService.java
```diff
@@ -65,6 +67,8 @@ public ConnectionTestResult test(@NotNull @Valid TestConnectionReq req) {
         Long connectionId = req.getId();
         if (ConnectionAccountType.SYS_READ == req.getAccountType()) {
             req.setTenantName("sys");
+        } else if (StringUtils.equals("sys", req.getTenantName()) && req.getType() != ConnectType.OB_MYSQL) {
+            throw new BadRequestException("sys租户是一个 MySQL 兼容租户");
```
yizhouxw, line 71: please handle i18n for error message, 
you may refer ErrorCodes usage

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionTesting.java
```diff
@@ -217,13 +217,13 @@ private ConnectionConfig reqToConnectionConfig(TestConnectionReq req) {
         } else {
             config.setType(req.getType());
         }
+        config.setDefaultSchema(req.getDefaultSchema());
```
yizhouxw, line 220: after ODC v4.2.0, datasource does not has 'defaultSchema' concept for oceanbase connectType

abigsaltyfish: just rolled back the previous modification operation in this way

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionTestService.java
abigsaltyfish: correct to ErrorCode ConnectionDatabaseTypeMismatched

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionTesting.java
yizhouxw: what is the reason to  move from line 226 to line 220 ?

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionTestService.java
```diff
@@ -65,6 +67,9 @@ public ConnectionTestResult test(@NotNull @Valid TestConnectionReq req) {
         Long connectionId = req.getId();
         if (ConnectionAccountType.SYS_READ == req.getAccountType()) {
             req.setTenantName("sys");
+        } else if (StringUtils.equals("sys", req.getTenantName()) && req.getType() != ConnectType.OB_MYSQL) {
+            return ConnectionTestResult.fail(ErrorCodes.ConnectionDatabaseTypeMismatched,
```
yizhouxw, line 71: `ConnectionTestResult.connectTypeMismatch(req.getType())`

abigsaltyfish: q1:It doesn't have any special meaning. I put it in the if else statement before, but now I'm just moving it out
q2:why "req.getType()",i see it means target type instead of choose type
`ConnectType connectType = ConnectTypeUtil.getConnectType(
                    connectionExtensionPoint.generateJdbcUrl(jdbcUrlProperties),
                    testConnectionProperties, queryTimeoutSeconds);
            ConnectionTestResult testResult = new ConnectionTestResult(result, connectType);
            if (type != null && connectType != null && !Objects.equals(connectType, type)) {
                return ConnectionTestResult.connectTypeMismatch(connectType);
            }`

yizhouxw (approved):

JessieWuJiexi (approved):

PR #3142, opened 2024-08-19
fix(dlm): alter file seq for 'connect_connection.sql' from 4 to 2

kiko-art:
#### What type of PR is this?
fix

alter file seq for 'connect_connection.sql' from 4 to 2

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #3147, opened 2024-08-19
feat(sqlcheck): SQL check supports affected  rows

yiminpeng:
#### What type of PR is this?
feat

SQL check supports affected  rows
1. add meta info and rule applying
2. affected rows check logic

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/factory/SQLAffectedRowsFactory.java
```diff
@@ -0,0 +1,51 @@
...
+    @Override
+    public SqlCheckRuleType getSupportsType() {
+        return SqlCheckRuleType.RESTRICT_SQL_AFFECTED_ROWS;
+    }
+
+    @Override
+    public SqlCheckRule generate(@NonNull DialectType dialectType, Map<String, Object> parameters) {
+        String key = getParameterNameKey("allowed-max-sql-affected-count");
+        if (parameters == null || parameters.isEmpty() || parameters.get(key) == null) {
+            return new MySQLAffectedRows(1000, jdbc);
```
MarkPotato777, line 47: Why 1000?

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/model/SqlCheckRuleType.java
```diff
@@ -243,7 +243,11 @@ public enum SqlCheckRuleType implements Translatable {
     /**
      * Truncate table 语句存在
      */
-    TRUNCATE_TBLE_EXISTS("truncate-table-exists");
+    TRUNCATE_TBLE_EXISTS("truncate-table-exists"),
+    /**
+     * 限制 SQL 执行影响行数，默认影响超过 1000 行时，拦截
```
MarkPotato777, line 248: Please use English instead. And why default 1000? why should it have a default value inside the rule itself?

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
```diff
@@ -0,0 +1,166 @@
...
+        return SqlCheckRuleType.RESTRICT_SQL_AFFECTED_ROWS;
+    }
+
+    /**
+     * Execution rule check
+     */
+    @Override
+    public List<CheckViolation> check(@NonNull Statement statement, @NonNull SqlCheckContext context) {
+
+        if (statement instanceof Insert || statement instanceof Update || statement instanceof Delete) {
```
MarkPotato777, line 86: What about `Truncate`?

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
```diff
@@ -0,0 +1,166 @@
...
+            if (maxSQLAffectedRows == 0) {
+                return Collections.emptyList();
+            }
+            String originSql = statement.getText();
+            long affectedRows = executeExplain(originSql, jdbcOperations);
+            if (affectedRows > maxSQLAffectedRows) {
+                return Collections.singletonList(SqlCheckUtil
+                        .buildViolation(statement.getText(), statement, getType(), new Object[] {}));
```
MarkPotato777, line 94: It doesn't match your i18n messages. You may need to pass the acutal and expected affected-rows into the messages.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
```diff
@@ -0,0 +1,166 @@
...
+     * @return affected rows
+     */
+    private long executeExplain(String originSql, JdbcOperations jdbc) {
+
+        if (originSql == null || jdbc == null) {
+            return 0;
+        }
+        String explainSql = "EXPLAIN " + originSql;
+
+        try {
```
MarkPotato777, line 124: It's kind of hard to understand the logic. Could you explain it in more detail? Including but not limited to
- Why we need the `ifFindValue` flag?
- what does `rs.getString(1)` actually represent?
- Why when only `affectedRows != 0 ` then you stop?
- Why `rowNum  > 2`?
- so on...

MarkPotato777:
please add unit test cases for your code

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/factory/SQLAffectedRowsFactory.java
yiminpeng: When there is no input value for this rule, the most conservative tactics is used by default, which is the default value of the production environment.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/model/SqlCheckRuleType.java
yiminpeng: Other rules are in Chinese. The default value of the rule itself is set to avoid limiting the number of rows being entered as 0 or empty. It is a bottom-up strategy. If simplified, it can be removed.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
yiminpeng: It's missing here. I'll make it up later.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
yiminpeng: Got it.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
yiminpeng: These are all related to the result set executed by explain on odc. These operations are all to get the value of "est.rows" in the result.

yiminpeng: Maybe I can achieve it in another way.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
yiminpeng: This issue only supports DML statements and does not involve DDL.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/model/SqlCheckRuleType.java
yiminpeng: The processing here is actually redundant, and the default value can be set without setting.

yhilmare:
Why is the target branch not dev/4.3.x?

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
MarkPotato777: `Truncate` belongs to DML. And if it's all about DML, your rule name should be precise. Please double check with @Jane201510

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/model/SqlCheckRuleType.java
MarkPotato777: Do not mind other rules, the current recommendation is all in English.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
```diff
@@ -0,0 +1,182 @@
...
+            });
+            Long firstNonNullResult = resultSet.stream()
+                    .filter(Objects::nonNull)
+                    .findFirst()
+                    .orElse(null);
+
+            return firstNonNullResult != null ? firstNonNullResult : 0;
+
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to execute sql: " + explainSql + ", error: " + e.getMessage());
```
MarkPotato777, line 164: What if the input SQL itself has a syntax error? What's the result of explaining the SQL?

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
```diff
@@ -0,0 +1,182 @@
...
+         */
+        try {
+            AtomicBoolean ifFindAffectedRow = new AtomicBoolean(false);
+            List<Long> resultSet = jdbc.query(explainSql, (rs, rowNum) -> {
+
+                String resultRow = rs.getString("Query Plan");
+                if (!ifFindAffectedRow.get() && rowNum > HEAD_LINE) {
+                    long affectedRows = getEstRowsValue(resultRow);
+                    // first non-null value is the column 'EST.ROWS'
+                    if (affectedRows != 0) {
```
MarkPotato777, line 149: return 0 seems not a good design. What if the 'affected rows' is indeed 0?

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
```diff
@@ -0,0 +1,182 @@
...
+     * Execution rule check
+     */
+    @Override
+    public List<CheckViolation> check(@NonNull Statement statement, @NonNull SqlCheckContext context) {
+
+        if (statement instanceof Insert || statement instanceof Update || statement instanceof Delete) {
+            if (maxSQLAffectedRows == 0) {
+                return Collections.emptyList();
+            }
+            String originSql = statement.getText();
```
MarkPotato777, line 92: refer to `SqlDiagnoseExtensionPoint`, which already have implemented.

server/odc-core/src/main/resources/i18n/BusinessMessages.properties
```diff
@@ -518,6 +518,12 @@ com.oceanbase.odc.builtin-resource.regulation.rule.sql-check.too-many-in-expr.de
 com.oceanbase.odc.builtin-resource.regulation.rule.sql-check.too-many-in-expr.max-in-expr-count=Maximum number of IN expressions
 com.oceanbase.odc.builtin-resource.regulation.rule.sql-check.too-many-in-expr.max-in-expr-count.description=The maximum number of expressions in the IN operation in the conditional predicate
 
+com.oceanbase.odc.builtin-resource.regulation.rule.sql-check.restrict-sql-affected-rows.message=The number of rows affected by this SQL statement exceeds the limit, and the maximum limit affects {0}, and the estimated is {1}
```
yizhouxw, line 521: @JessieWuJiexi

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/factory/SQLAffectedRowsFactory.java
yizhouxw, line 47: void hardcode default parameter here, if not set, throw exception instead

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
```diff
@@ -0,0 +1,182 @@
...
+ * limitations under the License.
+ */
+/*
+* Copyright (c) 2023 OceanBase.
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
+*     http://www.apache.org/licenses/LICENSE-2.0
```
yizhouxw, line 23: duplicated license header

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
```diff
@@ -0,0 +1,182 @@
...
+ * @version 1.0
+ * @date 2024-08-01 18:18
+ */
+public class MySQLAffectedRows implements SqlCheckRule {
+
+    private final Integer maxSQLAffectedRows;
+
+    private final JdbcOperations jdbcOperations;
+
+    private static final int HEAD_LINE = 2;
```
yizhouxw, line 67: what is HEAD_LINE, why set to 2 ?

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
```diff
@@ -0,0 +1,182 @@
...
+    /**
+     * execute 'explain' statement
+     *
+     * @param originSql target sql
+     * @param jdbc jdbc Object
+     * @return affected rows
+     */
+    private long getAffectedRows(String originSql, JdbcOperations jdbc) {
+
+        if (originSql == null || jdbc == null) {
```
yizhouxw, line 122: better throw exception for invalid parameter

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
yizhouxw: better if encapsulate 'extract explain result' into Utils, e.g OBExplainUtils . may reuse some method in diagnose module, i'm not sure.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/factory/SQLAffectedRowsFactory.java
yiminpeng: got it

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
yiminpeng: finished

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
yiminpeng: The result set in json format, the second line is the header.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
yiminpeng: got it

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/factory/SQLAffectedRowsFactory.java
yiminpeng: OK, throw an exception instead.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/model/SqlCheckRuleType.java
yiminpeng: OK, I will use all the new comments in English.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
yiminpeng: Ok, I only paid attention to insert/update/delete during the research.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
yiminpeng: I know, but the statement parameter is not the same. The latest code has been reprocessed.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
yiminpeng: Yes, I realized this problem, and the latesr code has been modified.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
yiminpeng: Maybe we can extract it together after we improve the support for Oracle database next time.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
yiminpeng: I have no idea about this, unless explain can be executed after many grammar checks.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
yiminpeng: Confirmed with the PM, This issue does not support `Truncate` for the time being, but it will include `select`.

yiminpeng: Confirmed with the PM, This issue does not support `Truncate` for the time being, but it will include `select`.

server/odc-service/src/test/java/com/oceanbase/odc/service/sqlcheck/MySQLCheckerTest.java
```diff
@@ -1362,6 +1362,107 @@ public void check_wrongPlExists_violationGenerated() {
         Assert.assertEquals(expect, actual);
     }
 
+    @Test
+    public void check_restrictSqlAffectedRows4MySQL_violationGenerated() {
```
MarkPotato777, line 1366: unit test should involve your code logic which is parsing the resultset of explain SQL.

server/odc-service/src/test/java/com/oceanbase/odc/service/sqlcheck/MySQLCheckerTest.java
```diff
@@ -1362,6 +1362,107 @@ public void check_wrongPlExists_violationGenerated() {
...
+                .thenReturn(Collections.singletonList(2L));
+        DefaultSqlChecker selectChecker = new DefaultSqlChecker(DialectType.MYSQL, "$$",
+                Collections.singletonList(
+                        new MySQLAffectedRows(2, DialectType.MYSQL, jdbcTemplate)));
+        List<CheckViolation> actualSelect = selectChecker.check(select);
+        Assert.assertEquals(0, actualSelect.size());
+    }
+
+    @Test
+    public void check_restrictSqlAffectedRows4OBMySQL_violationGenerated() {
```
yizhouxw, line 1416: one case one method, may split into cases

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
yizhouxw: > The result set in json format, the second line is the header.

unnessesary filed extraction, cuase hard to understand, and it is only referenced by specific sql query, may move to funtion block.

server/odc-service/src/test/java/com/oceanbase/odc/service/sqlcheck/MySQLCheckerTest.java
yiminpeng: OK, the style is  follow with other rules.

server/odc-service/src/test/java/com/oceanbase/odc/service/sqlcheck/MySQLCheckerTest.java
yiminpeng: Got it. The mock process has been adjusted.

server/odc-service/src/test/java/com/oceanbase/odc/service/sqlcheck/MySQLCheckerTest.java
yiminpeng: OK, the style is  follow with other rules.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
yiminpeng: ok, move to function done.

server/odc-core/src/main/resources/i18n/BusinessMessages.properties
yiminpeng: check

server/odc-service/src/test/java/com/oceanbase/odc/service/sqlcheck/MySQLCheckerTest.java
```diff
@@ -1362,6 +1362,212 @@ public void check_wrongPlExists_violationGenerated() {
...
+
+        Mockito.when(jdbcTemplate.query(Mockito.anyString(), Mockito.any(RowMapper.class)))
+                .thenThrow(new RuntimeException("Failed to execute sql"));
+        try {
+            DefaultSqlChecker errorChecker = new DefaultSqlChecker(DialectType.OB_MYSQL, "$$",
+                    Collections.singletonList(
+                            new MySQLAffectedRows(2L, DialectType.OB_MYSQL, jdbcTemplate)));
+            List<CheckViolation> actualError = errorChecker.check(update);
+        } catch (RuntimeException e) {
+            Assert.assertEquals("Failed to execute sql: " + "EXPLAIN " + update,
```
yizhouxw, line 1476: if no exception thrown, the Assert will not be execute, so please avoid try catch in unit testing logic.
use @Rule ExpectedExcetion mechanism instead.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
```diff
@@ -0,0 +1,204 @@
...
+
+            Long firstNonNullResult = resultSet.stream()
+                    .filter(Objects::nonNull)
+                    .findFirst()
+                    .orElse(null);
+
+            return firstNonNullResult != null ? firstNonNullResult : 0;
+
+        } catch (Exception e) {
+            throw new RuntimeException("Failed to execute sql: " + explainSql);
```
yizhouxw, line 186: how to handle this exception?

yizhouxw (commented):
test scenarios is very important for sql check feature.

how about select into ?

server/odc-service/src/test/java/com/oceanbase/odc/service/sqlcheck/MySQLCheckerTest.java
yiminpeng: got it, it's done.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
yiminpeng: I forgot to restore it when running the unit test. now finished.

yiminpeng:
Select is not included in the scope of check at present. From the explanation of affecting the number of lines, select does not affect the database in essence, but only retrieves. This is also consistent with competitors.

server/odc-service/src/test/java/com/oceanbase/odc/service/sqlcheck/MySQLCheckerTest.java
```diff
@@ -1362,6 +1364,211 @@ public void check_wrongPlExists_violationGenerated() {
...
+                .thenReturn(resultSet);
+        DefaultSqlChecker selectChecker = new DefaultSqlChecker(DialectType.OB_MYSQL, "$$",
+                Collections.singletonList(
+                        new MySQLAffectedRows(2L, DialectType.OB_MYSQL, jdbcTemplate)));
+        List<CheckViolation> actualSelect = selectChecker.check(select);
+        Assert.assertEquals(0, actualSelect.size());
+    }
+
+    @Rule
+    public final ExpectedException exceptionRule = ExpectedException.none();
```
MarkPotato777, line 1464: The field is supposed to be put in first place in the class.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    /**
+     * Execution rule check
+     */
+    @Override
+    public List<CheckViolation> check(@NonNull Statement statement, @NonNull SqlCheckContext context) {
+
+        if (statement instanceof Update || statement instanceof Delete || statement instanceof Insert) {
+            if (maxSQLAffectedRows == 0 || jdbcOperations == null) {
```
MarkPotato777, line 78: why returning `Collections.emptyList()` when `maxSQLAffectedRows == 0` ? It doesn't seem to make sense.

yiminpeng:
> Why is the target branch not dev/4.3.x?

merging process: feat/yizhuo_43x_sql_infect_rows --> feat/sql_infect_rows --> dev/4.3.x

yiminpeng:
> test scenarios is very important for sql check feature.
> 
> how about select into ?

Select is not included in the scope of check at present. From the explanation of affecting the number of lines, select does not affect the database in essence, but only retrieves. This is also consistent with competitors.

yiminpeng:
> please add unit test cases for your code

OK, done.

server/odc-service/src/test/java/com/oceanbase/odc/service/sqlcheck/MySQLCheckerTest.java
yiminpeng: That's true, but what I'm considering is that it may be easy to understand on the actual case. It's not a big problem. I put it at the top.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
yiminpeng: `maxSQLAffectedRows == 0` should not be included. It should have been omitted when reconstructing the code and has been modified now.

yizhouxw (dismissed):

MarkPotato777 (approved):
LGTM

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/factory/SQLAffectedRowsFactory.java
```diff
@@ -0,0 +1,51 @@
...
+
+import com.oceanbase.odc.core.shared.constant.DialectType;
+import com.oceanbase.odc.service.sqlcheck.SqlCheckRule;
+import com.oceanbase.odc.service.sqlcheck.SqlCheckRuleFactory;
+import com.oceanbase.odc.service.sqlcheck.model.SqlCheckRuleType;
+import com.oceanbase.odc.service.sqlcheck.rule.MySQLAffectedRows;
+
+import lombok.NonNull;
+
+public class SQLAffectedRowsFactory implements SqlCheckRuleFactory {
```
yhilmare, line 30: SQLAffectedRowsFactory -> SqlAffectedRowsFactory

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
```diff
@@ -0,0 +1,212 @@
...
+
+/**
+ * {@link MySQLAffectedRows}
+ *
+ * @author yiminpeng
+ * @version 1.0
+ * @date 2024-08-01 18:18
+ */
+@Slf4j
+public class MySQLAffectedRows implements SqlCheckRule {
```
yhilmare, line 50: MySQLAffectedRows -> MySQLAffectedRowsExceedLimit

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
```diff
@@ -0,0 +1,212 @@
...
+
+    private final Long maxSQLAffectedRows;
```
yhilmare, line 52: maxSQLAffectedRows -> maxSqlAffectedRows

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    /**
+     * Execution rule check
+     */
+    @Override
+    public List<CheckViolation> check(@NonNull Statement statement, @NonNull SqlCheckContext context) {
+
+        if (statement instanceof Update || statement instanceof Delete || statement instanceof Insert) {
+            if (jdbcOperations == null) {
```
yhilmare, line 80: don't throw the exception at the runtime, we should check it not null when we construct it~

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
```diff
@@ -0,0 +1,212 @@
...
+                        break;
+                    case OB_MYSQL:
+                        affectedRows = getOBMySqlAffectedRows(explainSql, jdbcOperations);
+                        break;
+                    default:
+                        log.warn("Unsupported dialect type: {}", dialectType);
+                        break;
+                }
+            } catch (Exception e) {
+                log.error("Error in calling getAffectedRows method", e);
```
yhilmare, line 98: log.warn

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/SqlCheckRules.java
```diff
@@ -151,12 +154,15 @@ public static List<SqlCheckRuleFactory> getAllFactories(DialectType dialectType,
...
         rules.add(new OfflineDdlExistsFactory(jdbc));
         rules.add(new TruncateTableExistsFactory());
+        rules.add(new SQLAffectedRowsFactory(jdbc));
         return rules;
     }
 
     public static List<SqlCheckRule> getAllDefaultRules(JdbcOperations jdbc, @NonNull DialectType dialectType) {
         return SqlCheckRules.getAllFactories(dialectType, jdbc).stream()
-                .map(f -> f.generate(dialectType, null)).filter(Objects::nonNull).collect(Collectors.toList());
+                .map(f -> f.generate(dialectType,
```
yhilmare, line 163: why should we change this?

yizhouxw (dismissed):
lgtm, please implement select .. into affected rows check in separated PR

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
yiminpeng: > MySQLAffectedRows -> MySQLAffectedRowsExceedLimit

done

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/factory/SQLAffectedRowsFactory.java
yiminpeng: done

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
yiminpeng: done

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
yiminpeng: Ok, checked it before running.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLAffectedRows.java
yiminpeng: done

yiminpeng:
> lgtm, please implement select .. into affected rows check in separated PR

got it.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/SqlCheckRules.java
yiminpeng: give a default value by function

yhilmare (dismissed):
lgtm

PeachThinking (approved):

PeachThinking (approved):

PeachThinking (approved):

JessieWuJiexi (approved):

Jane201510 (approved):
agree

MarkPotato777 (approved):

yhilmare (approved):
lgtm

PR #3154, opened 2024-08-20
feat(module): increase sub-module loading capabilities 

yhilmare:
#### What type of PR is this?

type-feature

Some functions need to be made pluggable. This pr adds a modules module, which is specifically used to store these pluggable function modules.

Fixes #

```docs

```


yizhouxw (approved):
lgtm

PR #3160, opened 2024-08-21
fix(logicaldatabase): index out of bound when there only exists one table in the database

MarkPotato777:
#### What type of PR is this?

type-bug
When there only exists one table in the database, `LogicalTableRecognitionUtils.recognizeLogicalTablesWithExpression` will return empty list. In this case, we should just return the full-qualified name of the table.

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #3201, opened 2024-08-26
fix(web): fix tomcat threads config

ungreat:
#### What type of PR is this?

The tomcat configuration is incompatible after the springboot version upgrade.

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #3203, opened 2024-08-26
fix(connect): make odc connect to the obcloud instance

yhilmare:
#### What type of PR is this?

type-bug

odc can not connect to the obcloud's instance, the reason is below:

the connection info of the obcloud instance may contain tenant id or cluster id, just like this:
`obclient -h tenantId.xxx.xxx -P 1521 -u ODCTEST@tenantId -pxxxx`

but we expect that the connection info does not contain tenant or cluster info when user select 'OBCloud Oracle/MySQL' connect type.

To solve this problem, we add some specific logic: we check the username and find out if there is '@' or '#' char in it. we will split the username by '@' or '#'.

Fixes #

```docs

```


yizhouxw (approved):
lgtm

PR #3224, opened 2024-08-27
[WIP] refactor(task): task refactor 

LioRoger:
#### What type of PR is this?

type-feature

this PR split the binding relationship of resource and job.
make resource a dependent module.
when job finished,  it won't wait resource destroy. 
resource became a transparent concept for job submitter

Fixes #
1. make resource as a dependent module
2. introduce resource operator as resource manipulate interface
3. introduce abstract k8s resource manager implement.take ocp k8s resource manager as default value.

```docs

```


server/odc-migrate/src/main/resources/migrate/common/V_4_3_2_4__add_task_resource.sql
```diff
@@ -0,0 +1,17 @@
+--
+-- Add task_resource table
+--
+CREATE TABLE IF NOT EXISTS `task_resource` (
```
yhilmare, line 4: 4.3.2.4 is occupied, 4.3.2.6 is available

PR #3235, opened 2024-08-28
feat(data-transfer): data import supports file directory

persistafei:
What type of PR is this?
type-feature

data import now supports choosing file directory

Which issue(s) this PR fixes:
Fixes #

Special notes for your reviewer:
Additional documentation e.g., usage docs, etc.:

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/datatransfer/loader/PlSqlFormat.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/datatransfer/loader/PlSqlFormat.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024 OceanBase.
+ * Copyright (c) 2023 OceanBase.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -25,8 +25,8 @@ import com.oceanbase.odc.common.util.StringUtils;
  * @date 2024/8/27
  */
 public class PlSqlFormat {
-    private static final Set<String>     LEGAL_FILE_SUFFIXES = new HashSet<>();
-    private static final Set<ObjectType> PL_OBJECTS          = new HashSet<>();
+    private static final Set<String> LEGAL_FILE_SUFFIXES = new HashSet<>();
+    private static final Set<ObjectType> PL_OBJECTS = new HashSet<>();
 
     static {
         for (ObjectType type : ObjectType.values()) {
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/datatransfer/loader/PlSqlFormat.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/datatransfer/loader/PlSqlFormat.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2024 OceanBase.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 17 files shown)


server/plugins/task-plugin-api/src/main/java/com/oceanbase/odc/plugin/task/api/datatransfer/dumper/ExportOutput.java
```diff
@@ -102,6 +107,17 @@ public void toFolder(@NonNull File folder) throws IOException {
         if (!folder.exists()) {
             FileUtils.forceMkdir(folder);
         }
+
+        if (inputDir != null) {
+            // if input file is a dir, try to use symbolic link, avoid copy files if possible
+            try {
+                Files.createSymbolicLink(this.inputDir, folder.toPath());
```
yizhouxw, line 114: when the symbolic link will be unlinked ?

is it possibile to read from from origin directory?

server/plugins/task-plugin-api/src/main/java/com/oceanbase/odc/plugin/task/api/datatransfer/dumper/ExportOutput.java
```diff
@@ -66,12 +66,16 @@ public class ExportOutput {
     private final BinaryFile<Manifest> manifest;
     private final BinaryFile<List<?>> checkpoints;
     private final List<DumpDBObject> dumpDbObjects;
+    private boolean isZip;
```
yizhouxw, line 69: no `is` prefix for boolean field, use `zip` straightly

server/odc-service/src/main/java/com/oceanbase/odc/service/datatransfer/loader/PlSqlFormat.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+import com.oceanbase.odc.common.util.StringUtils;
+
+/**
+ * @author youshu
+ * @date 2024/8/27
+ */
+public class PlSqlFormat {
```
yizhouxw, line 27: PLSqlDeveloperExportFormat ?

1. it is the file format used by PL/SQL Developer tool exported, not PL/SQL languange identified.
2. Sql should be SQL better indeed, however too much Sql used in odc already..

server/plugins/task-plugin-api/src/main/java/com/oceanbase/odc/plugin/task/api/datatransfer/dumper/ExportOutput.java
persistafei: after data transfer task finished, either success or failure, the /datatransfer/${seq}  temp directory will be cleaned, and files will be unlinked.

if read from origin directory, log files and some hidden files will be written to that place, which might cause some trouble for user.

server/plugins/task-plugin-api/src/main/java/com/oceanbase/odc/plugin/task/api/datatransfer/dumper/ExportOutput.java
persistafei: OK

server/odc-service/src/main/java/com/oceanbase/odc/service/datatransfer/loader/PlSqlFormat.java
persistafei: OK

yizhouxw (dismissed):
lgtm

yizhouxw (approved):

PR #3244, opened 2024-08-30
fix(deployment): ODC_APP_EXTRA_ARGS not works

### LINKED ISSUE #2781: [Bug]: Startup parameter ODC_APP_EXTRA_ARGS does not work as expected
### ODC version

4.3.0

### OB version

4.2.1.6

### What happened?

When I was developing in integrating ODC in [ob-operator](https://github.com/oceanbase/ob-operator), I found that the startup argument `ODC_APP_EXTRA_ARGS` not working. 

(The following IP address is an internal mock one)

I set `ODC_APP_EXTRA_ARGS`="--odc.integration.bastion.account.query.request-url='http://10.42.0.1:18084/api/v1/auth/${account_verify_token}' --odc.integration.bastion.enabled='true' --odc.integration.bastion.account.auto-login-enabled='true' --odc.connect.database.sync.block-internal-database='false' --odc.database.schema.sync.exclude-schemas.ob-mysql=''"

And I found that the `sun.java.command` indeed contains the extra args in `odc.log`.

image

But the parameters do not work at all. I need to update them in the metadata database and restart odc server to make them work as expected.

Parameters set in `ODC_APP_EXTRA_ARGS` work as expected. The bastion link could be opened normally.

Run odc server with environment variable `ODC_APP_EXTRA_ARGS` set to `--odc.integration.bastion.enabled='true' --odc.integration.bastion.account.auto-login-enabled='true'`, visit URL `http://xx.xx.xx.xx:8989/#/gateway/test` and you will found "session expired, please log in again". After login, visit the above URL again and you will get another error message.

image

image

If the parameters were set in metadata database and restart the odc server, the problem disappeared. 

_No response_

yizhouxw:
#### What type of PR is this?

type-bug

ODC startup environment variable ODC_APP_EXTRA_ARGS does not works.

root cauase was RefreshScope mechasnim reload configuration from MetaDB config_system_configuration.

involved persistence process for configuration items set via ODC_APP_EXTRA_ARGS.

Fixes #2781

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/config/SystemConfigBootstrap.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/config/SystemConfigBootstrap.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024 OceanBase.
+ * Copyright (c) 2023 OceanBase.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -13,7 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package com.oceanbase.odc.service.config;
 
 import java.util.Arrays;
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/config/SystemConfigBootstrap.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/config/SystemConfigBootstrap.java
@@ -1,0 +1,93 @@
+/*
+ * Copyright (c) 2024 OceanBase.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.oceanbase.odc.service.config;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import javax.annotation.PostConstruct;
+import javax.sql.DataSource;
+
+import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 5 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/config/SystemConfigBootstrap.java
```diff
@@ -0,0 +1,92 @@
...
+                .filter(kv -> kv.length == 2)
+                .map(kv -> new Configuration(StringUtils.trim(kv[0]), StringUtils.trim(kv[1])))
+                .collect(Collectors.toList());
+    }
+
+    private void upsertSystemConfigurations(JdbcTemplate jdbcTemplate, List<Configuration> configurations) {
+        if (CollectionUtils.isEmpty(configurations)) {
+            return;
+        }
+        String sql = "INSERT INTO config_system_configuration (`key`, `value`) VALUES (?, ?) "
```
yhilmare, line 79: Will this table `config_system_configuration` be created before this sql?

I mean when start up a total new odc~

yizhouxw: yes, see `@DependsOn` annotation

yhilmare (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/config/SystemConfigBootstrap.java
```diff
@@ -0,0 +1,92 @@
...
+@Component
+@DependsOn("metadbMigrate")
+public class SystemConfigBootstrap {
+    private static final int BATCH_SIZE = 100;
+
+    @Autowired
+    protected DataSource dataSource;
+
+    @PostConstruct
+    public void bootstrap() {
```
MarkPotato777, line 52: What's the timing when these configurations are loading in the Spring Application Context? Will they take effect **right after** the odc-server starts up, or should we wait for the next config refresh?

yizhouxw: after metadb migration, before odc server startup

MarkPotato777: But will Spring be aware of this configuration change? Should we call `SystemConfigService.refresh` after these configs upsert to the table?

yizhouxw: the configuration changed before spring context initialized

MarkPotato777: got it.

MarkPotato777 (approved):
LGTM

PR #3254, opened 2024-09-03
fix(resultset-export): the exported file is empty if use lower table name for oracle mode

LuckyPickleZZ:
#### What type of PR is this?
type-bug

when exporting result set in oracle mode  and using a lower case table name , the exported file would be empty.
It's because obdumper would always generate a file with upper case name.


PeachThinking (approved):
lgtm

persistafei (dismissed):

yhilmare (approved):
lgtm

PeachThinking (approved):

PR #3261, opened 2024-09-04
fix(statefulRoute):sensitive column may can't get result

ungreat:

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/datasecurity/model/SensitiveColumnScanningTaskInfo.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/datasecurity/model/SensitiveColumnScanningTaskInfo.java
@@ -42,7 +42,8 @@ public class SensitiveColumnScanningTaskInfo {
     private ErrorCode errorCode;
     private String errorMsg;
 
-    public SensitiveColumnScanningTaskInfo(String taskId, @NonNull Long projectId, @NonNull Integer allTableCount) {
+    public SensitiveColumnScanningTaskInfo(@NonNull String taskId, @NonNull Long projectId,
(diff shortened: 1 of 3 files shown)


yhilmare (dismissed):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/datasecurity/model/SensitiveColumnScanningTaskInfo.java
```diff
@@ -43,8 +42,8 @@ public class SensitiveColumnScanningTaskInfo {
     private ErrorCode errorCode;
     private String errorMsg;
 
-    public SensitiveColumnScanningTaskInfo(@NonNull Long projectId, @NonNull Integer allTableCount) {
-        this.taskId = StringUtils.uuid();
+    public SensitiveColumnScanningTaskInfo(String taskId, @NonNull Long projectId, @NonNull Integer allTableCount) {
```
LuckyPickleZZ, line 45: Do we need a `NotNull` annotation here

PeachThinking (approved):
lgtm

yizhouxw (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/datasecurity/model/SensitiveColumnScanningTaskInfo.java
ungreat: Of course, adding @NonNull is better, but since @Validated is not added, this annotation is only used by the compiler and does not actually perform validation.

yizhouxw (approved):

yhilmare (approved):
lgtm

PR #3269, opened 2024-09-04
fix(manual execution):manual execution button should not be displayed after clicking manual execution

zijiacj:
#### What type of PR is this?
type-bug

At present, after clicking manual execution button on the ticket, the status of ticket in the database will remain in wait_for_execution for a period of time, and the front end displays the execution button according to this status. Therefore, after clicking manual execution, the execution button will appear for a period of time, and an error will be reported when the user clicks execution button again.
The improvement is to change the status of the ticket to not ‘wait_for_execution’ immediately after clicking Manual execution button

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #3271, opened 2024-09-04
fix(connection): add listByProjectId return fields

### LINKED ISSUE #3178: [Bug]: When scanning and adding sensitive columns, selecting a database will not display database details
### ODC version

odc432

### OB version

*

### What happened?

image

image

Select database when scanning and adding sensitive columns

_No response_

yiminpeng:
#### What type of PR is this?

fix

Fixes #3178 

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
@@ -423,9 +423,19 @@ public class ConnectionService {
         if (basic) {
             connections = repository.findByDatabaseProjectId(projectId).stream().map(e -> {
                 ConnectionConfig c = new ConnectionConfig();
+                Environment environment = environmentService.detail(e.getEnvironmentId());
                 c.setId(e.getId());
                 c.setName(e.getName());
                 c.setType(e.getType());
+
+                c.setHost(e.getHost());
+                c.setPort(e.getPort());
+                c.setUsername(e.getUsername());
+                c.setTenantName(e.getTenantName());
+                c.setClusterName(e.getClusterName());
+                c.setEnvironmentName(environment.getName());
+                c.setEnvironmentStyle(environment.getStyle());
+
                 return c;
             }).collect(Collectors.toList());
         } else {


server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
```diff
@@ -423,9 +423,19 @@ public PaginatedData<ConnectionConfig> listByProjectId(@NotNull Long projectId,
         if (basic) {
             connections = repository.findByDatabaseProjectId(projectId).stream().map(e -> {
                 ConnectionConfig c = new ConnectionConfig();
+                Environment environment = environmentService.detail(e.getEnvironmentId());
```
yhilmare, line 426: On complex, we need to avoid this~

yizhouxw: yes, avoid MetaDB access in loop, which may result in bad performance

yizhouxw (commented):
bugfix PR title should be problem scenario ,not solution.

describe solution in PR description instead

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
yiminpeng: I don't quite understand. Do you mean that the method of executing `.detail` in `stream` is complicated?

yiminpeng: I don't quite understand. Do you mean that the method of executing `.detail` in `stream` is complicated?

> yes, avoid MetaDB access in loop, which may result in bad performance

got it. I just checked the implementation and found this problem.

yiminpeng: I don't quite understand. Do you mean that the method of executing `.detail` in `stream` is complicated?

> yes, avoid MetaDB access in loop, which may result in bad performance

got it. I just checked the implementation and found this problem.

> On complex, we need to avoid this~

Yes, I guessed it. Thanks for the reminder.

yiminpeng:
> bugfix PR title should be problem scenario ,not solution.
> 
> describe solution in PR description instead

ok, it will be modified and described later.

server/odc-service/src/main/java/com/oceanbase/odc/metadb/collaboration/EnvironmentRepository.java
```diff
@@ -39,4 +40,9 @@ public interface EnvironmentRepository extends JpaRepository<EnvironmentEntity,
     @Transactional
     int updateEnabledById(@Param("id") Long id, @Param("enabled") Boolean enabled);
 
+    @Query(value = "select name, style from collaboration_environment where id in (:ids)", nativeQuery = true)
```
yhilmare, line 43: no need to add this native query sql

server/odc-service/src/main/java/com/oceanbase/odc/service/collaboration/environment/EnvironmentService.java
```diff
@@ -155,6 +156,18 @@ public Environment detail(@NonNull Long id) {
         return innerDetail(id);
     }
 
+    @PreAuthenticate(actions = "select", resourceType = "ODC_ENVIRONMENT", indexOfIdParam = 0)
```
yhilmare, line 159: what does this mean?

yiminpeng: perform some authentications of select, refer to the create method above

server/odc-service/src/main/java/com/oceanbase/odc/metadb/collaboration/EnvironmentRepository.java
yiminpeng: don' need any of these three annotations? here is reference `update`  operation. why no need we use query annotation here?

server/odc-service/src/main/java/com/oceanbase/odc/service/collaboration/environment/EnvironmentService.java
MarkPotato777: You don't need to add this annotation when you use this method internally. Add `@SkipAuthorize("odc internal usage")` instead.
BTW. the action is not "select", it should be read. Refer to other methods before you write.

yiminpeng: > You don't need to add this annotation when you use this method internally. Add `@SkipAuthorize("odc internal usage")` instead. BTW. the action is not "select", it should be read. Refer to other methods before you write.

OK, I only refer to the `create` method here.

server/odc-service/src/main/java/com/oceanbase/odc/metadb/collaboration/EnvironmentRepository.java
MarkPotato777: No, you don't. Please read the Spring Data JPA docs before writing the code if you do not actually understand the meaning of these three annotations.

yiminpeng: Right, `EnvironmentRepository` extend JPA. I refer to `update`. I don't have an in-depth understanding of the differences here.

yiminpeng: if don't use `@Query`,  Similar to MybatisPlus, got it.

server/odc-service/src/main/java/com/oceanbase/odc/service/collaboration/environment/EnvironmentService.java
```diff
@@ -155,6 +156,18 @@ public Environment detail(@NonNull Long id) {
         return innerDetail(id);
     }
 
+    @Transactional(rollbackFor = Exception.class)
+    @SkipAuthorize("odc internal usage")
+    public List<Environment> getByIdIn(@NotNull Collection<Long> ids) {
+        List<EnvironmentEntity> entity = environmentRepository.findByIdIn(ids);
```
yizhouxw, line 162: if ids is an empty collection, the JPA query will failed.

may add empty check, or just return empty list while ids is empty.

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
```diff
@@ -420,12 +420,27 @@ public List<ConnectionConfig> listByOrganizationIdIn(@NonNull Collection<Long> o
...
         List<ConnectionConfig> connections;
+        List<Long> environmentIds = repository.findByDatabaseProjectId(projectId).stream()
+                .map(ConnectionEntity::getEnvironmentId)
+                .distinct().collect(Collectors.toList());
+        Map<Long, Environment> environmentMap = environmentService.getByIdIn(environmentIds).stream()
+                .collect(Collectors.toMap(Environment::getId, environment -> environment));
         if (basic) {
             connections = repository.findByDatabaseProjectId(projectId).stream().map(e -> {
                 ConnectionConfig c = new ConnectionConfig();
+                Environment environment = environmentMap.get(e.getEnvironmentId());
```
yizhouxw, line 431: what if environment object is null ?

yiminpeng: There should be a null check here.

server/odc-service/src/main/java/com/oceanbase/odc/service/collaboration/environment/EnvironmentService.java
yiminpeng: got it

yiminpeng: This part of the logic has been integrated into the list.

yhilmare (dismissed):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
```diff
@@ -420,12 +420,32 @@ public List<ConnectionConfig> listByOrganizationIdIn(@NonNull Collection<Long> o
             resourceType = "ODC_PROJECT", indexOfIdParam = 0)
     public PaginatedData<ConnectionConfig> listByProjectId(@NotNull Long projectId, @NotNull Boolean basic) {
         List<ConnectionConfig> connections;
+        List<Long> environmentIds = repository.findByDatabaseProjectId(projectId).stream()
```
MarkPotato777, line 423: There is no need to execute the same query `repository.findByDatabaseProjectId` twice on the line 423 and line 431.

yiminpeng: wow, very detailed, resolved.

yhilmare (dismissed):
lgtm

MarkPotato777 (approved):
LGTM

PR #3281, opened 2024-09-05
fix(data-transfer): fail to start transfer task on windows

### LINKED ISSUE #3125: [Bug]: fail to export mysql data on windows
### ODC version

4.2.x

### OB version

mysql

### What happened?

When export mysql data on windows, there would be an exception `InvalidPathException: Illegal char <*> at index xxx`.
This is because of `*` is an illegal character in path on windows. We should not use Paths#get method here.
```java
 command.add(Paths.get(dataxHomePath, SystemUtils.isOnWindows() ? "lib/*" : "lib/*:.").toString());
```

export successfully

-

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

We use `Paths#get` to get class path for datax task. But on windows, the classpath like 'lib/*:.' are illegal file path. The get method would raise an illegal path format exception.
So we concat the path straightly instead of using `Paths`

Fixes #3125

```docs

```


yhilmare (approved):
lgtm

PR #3289, opened 2024-09-05
feat(dfa): add simple dfa for workflow cluster's status transfer

yhilmare:
#### What type of PR is this?

type-feat

add fsm impl for workflow cluster's status transfer

Fixes #

```docs

```

### DIFF
--- a/server/odc-core/src/main/java/com/oceanbase/odc/core/dfa/AbstractDfa.java
+++ b/server/odc-core/src/main/java/com/oceanbase/odc/core/dfa/AbstractDfa.java
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.oceanbase.odc.core.fsm;
+package com.oceanbase.odc.core.dfa;
 
 import java.util.List;
 import java.util.stream.Collectors;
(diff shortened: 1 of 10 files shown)


yizhouxw (commented):
osc module has an asm abstraction, is it possible reuse ?

yhilmare:
> osc module has an asm abstraction, is it possible reuse ?

It can't be, I know what you mentioned is `ActionFsm` I checked this class and I think that I can not reuse it:

I want to use fsm to transfer workflow, workflow cluster and etc.'s status. the main process of the fsm is: the fsm has the state of A, the event B has happened. The fsm transfer the state from A to C because of the event B.

but the `ActionFsm`'s core process is: I input a `context` into the fsm and the fsm resolves the state from the context. fsm get the event by the state and event is executed based on the context we input. we get the next state based on the current state, event execution result and the context we input.

as you can see, the event and the context are one-to-one correspondence that is to say the event can be 100% confirmed if you input a specified context(the status of workflow and etc.), this is not expected~, what I expected is that I input a context(set the status of the fsm), the event is not specified~

yizhouxw (approved):
lgtm

PR #3290, opened 2024-09-06
fix(database-change): missing sql plan task execution result 

kiko-art:
#### What type of PR is this?
type-bug

sql plan task add schedule job, the value of task_result is obtained from the job_job.

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduleTaskService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduleTaskService.java
@@ -54,6 +54,7 @@ import com.oceanbase.odc.metadb.iam.UserRepository;
 import com.oceanbase.odc.metadb.schedule.ScheduleTaskEntity;
 import com.oceanbase.odc.metadb.schedule.ScheduleTaskRepository;
 import com.oceanbase.odc.metadb.schedule.ScheduleTaskSpecs;
+import com.oceanbase.odc.metadb.task.JobEntity;
 import com.oceanbase.odc.metadb.task.JobRepository;
 import com.oceanbase.odc.service.common.model.InnerUser;
 import com.oceanbase.odc.service.common.response.SuccessResponse;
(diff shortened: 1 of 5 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduleTaskService.java
```diff
@@ -291,6 +294,14 @@ public Page<ScheduleTaskListOverview> getConditionalScheduleTaskListResp(Pageabl
         Map<Long, Database> databaseMap = scheduleResponseMapperFactory.getDatabaseInfoByIds(databaseIds).stream()
                 .collect(Collectors.toMap(Database::getId, Function.identity()));
 
+        // get job result json
+        List<Long> jobIds = scheduleTaskPage.getContent().stream().map(ScheduleTask::getJobId).collect(
```
guowl3, line 298: need to check if jobId is null

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/SqlPlanTask.java
```diff
@@ -97,6 +100,17 @@ protected void doInit(JobContext context) {
                 PublishSqlPlanJobReq.class);
         this.connectionSession = generateSession();
         this.executor = connectionSession.getSyncJdbcExecutor(ConnectionSessionConstants.CONSOLE_DS_KEY);
+        long timeoutUs = TimeUnit.MILLISECONDS.toMicros(parameters.getTimeoutMillis());
+        if (timeoutUs < 0) {
+            throw new IllegalArgumentException(
```
yizhouxw, line 105: the error message is not correct, may simplify as 'Invalid timeout settings', and please use PreConditions  instead

server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduleTaskService.java
kiko-art: Have been modified

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/SqlPlanTask.java
kiko-art: Have been modified.

yizhouxw (approved):
lgtm

guowl3 (approved):
lgtm

yhilmare (approved):
lgtm

PR #3291, opened 2024-09-06
feat(resource): introduce resource module

LioRoger:
#### What type of PR is this?

type-feature
introduce resource module
1. make resource as a dependent module
2. introduce resource operator as resource manipulate interface
3. resource operator can be dynamic registered to adopt multi resource management needs

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/metadb/resource/ResourceEntity.java
```diff
@@ -0,0 +1,121 @@
...
+ */
+@Data
+@Entity
+@ToString
+@Table(name = "task_resource")
+public class ResourceEntity {
+    public static final String CREATE_TIME = "createTime";
+    public static final String STATUS = "status";
+    /**
+     * id for task
```
guowl3, line 52: Is this field supposed to represent the auto-increment ID of the resource? It doesn't seem to be related to the task

server/odc-service/src/main/java/com/oceanbase/odc/metadb/resource/ResourceEntity.java
```diff
@@ -0,0 +1,121 @@
...
+     * region of the resource, eg beijing/shanghai
+     */
+    @Column(name = "region", updatable = false, nullable = false)
+    private String region;
+
+    /**
+     * group of the resource, eg k8s cluster
+     */
+    @Column(name = "group_name", updatable = false, nullable = false)
+    private String groupName;
```
guowl3, line 85: The 'groupName' field is somewhat confusing, but I don't have a better suggestion...

server/odc-service/src/main/java/com/oceanbase/odc/service/resource/ResourceState.java
```diff
@@ -0,0 +1,53 @@
...
+ * @date 2024/8/12 11:50
+ */
+public enum ResourceState {
+    // when submit create resource request, resource not available yet
+    // k8s pending/container creating / init
+    CREATING,
+    // resource is created, compute resource available
+    // task will only yield to resource with state running
+    // k8s running
+    RUNNING,
```
guowl3, line 31: 'AVAILABLE' might be better for the resource if there are no historical compatibility costs

server/odc-service/src/main/java/com/oceanbase/odc/metadb/resource/ResourceEntity.java
LioRoger: this filed not related to task. it's only a table unique key related to resource entity storage. even in resource module, this id is not visible to users.

server/odc-service/src/main/java/com/oceanbase/odc/metadb/resource/ResourceEntity.java
LioRoger: region name union group name should related to a endpoint of resource operator. cloud provider is candidate name but may be too strictly. So group name is acceptable

server/odc-service/src/main/java/com/oceanbase/odc/service/resource/ResourceState.java
LioRoger: agreed

server/odc-service/src/main/java/com/oceanbase/odc/service/resource/ResourceEndPoint.java
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * endpoint of resource
+ * 
+ * @author longpeng.zlp
+ * @date 2024/8/12 11:10
+ */
+@AllArgsConstructor
+@Data
+public class ResourceEndPoint {
+    private final String resourceURL;
```
yhilmare, line 30: resourceUrl

LioRoger: done

yhilmare (approved):
lgtm

PR #3292, opened 2024-09-06
feat(log): support full log to download and refactor the logging module code

ysjemmm:
#### What type of PR is this?
type feat

support download full log

Fixes #

fix 43x branch could not package docker image pipeline

```docs

```


server/odc-common/src/main/java/com/oceanbase/odc/common/unit/BinarySizeUnit.java
```diff
@@ -49,6 +51,15 @@ public enum BinarySizeUnit {
         }
     }
 
+    public static long convertTo(long originalSize, @NotNull BinarySizeUnit originalUnit,
```
yhilmare, line 54: `BinarySizeUnit.GB.of(1).convert(BinarySizeUnit.B)`

which means 1GB to 1024 * 1024 * 1024 B

ysjemmm: OK，I will change

script/functions.sh
```diff
@@ -300,40 +300,46 @@ function maven_build_jar() {
     return 0
 }
 
+function build_module() {
+    local module=$1
+    local setting_xml_path="${ODC_DIR}/builds/settings.xml"
```
yhilmare, line 305: there is no file named this `${ODC_DIR}/builds/settings.xml`

ysjemmm: I have changed scripts that names are build_libs.sh and function.sh

server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/model/ScheduleTaskDetailRespHist.java
```diff
@@ -44,4 +44,6 @@ public class ScheduleTaskDetailRespHist {
 
     private String jobGroup;
 
+    private String fullLogDownloadUrl;
```
guowl3, line 47: same as above

server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/model/ScheduleTaskDetailResp.java
```diff
@@ -48,4 +48,6 @@ public class ScheduleTaskDetailResp {
     private Date createTime;
 
     private Date updateTime;
+
+    private String fullLogDownloadUrl;
```
guowl3, line 52: same as above

server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/model/ScheduleTaskOverview.java
```diff
@@ -46,6 +48,8 @@ public class ScheduleTaskOverview {
      */
     private String jobGroup;
 
+    private String fullLogDownloadUrl;
```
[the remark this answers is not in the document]
ysjemmm: ok

ysjemmm:
see https://github.com/oceanbase/odc/pull/3317

PR #3301, opened 2024-09-06
feat(worksheet): add implementation for flat list

CHLK:
#### What type of PR is this?
type-feature

1. add flatList implementation for WorksheetControllerIntegration；
2. add last access time in DefaultWorksheetService#detail;
3. add max project worksheet number check in worksheet create and upload

Fixes #

```docs

```


server/odc-core/src/main/java/com/oceanbase/odc/core/shared/constant/LimitMetric.java
```diff
@@ -45,7 +45,8 @@ public enum LimitMetric implements Translatable {
     EXPORT_OBJECT_COUNT,
     TABLE_NAME_LENGTH,
     WORKSHEET_CHANGE_COUNT,
-    WORKSHEET_SAME_LEVEL_COUNT;
+    WORKSHEET_SAME_LEVEL_COUNT,
+    WORKSHEET_PROJECT_COUNT;
```
yizhouxw, line 49: concept confused, it is not clear here mean worksheet count of project count,
may WORKSHEET_COUNT_IN_PROJECT

server/integration-test/src/test/java/com/oceanbase/odc/server/web/controller/v2/WorksheetControllerIntegrationTest.java
```diff
@@ -323,6 +324,43 @@ public void listWorksheets() throws Exception {
                         Collectors.toList()));
     }
 
+    @Test
+    public void flatListWorksheets() throws Exception {
```
yizhouxw, line 328: complex unit testing case here, often we don't write logic in Controller class, so ODC does not calculate Unit Testing coverage for Controller classes

server/odc-core/src/main/java/com/oceanbase/odc/core/shared/constant/LimitMetric.java
CHLK: Okay，I'll make this change.

server/integration-test/src/test/java/com/oceanbase/odc/server/web/controller/v2/WorksheetControllerIntegrationTest.java
CHLK: Okay，I'll add WorksheetServiceFacadeImplTest to WorksheetControllerIntegrationTest

CHLK:
@yizhouxw  I've implemented your suggestion, please take a look when you have a chance.

yhilmare (approved):
lgtm

PR #3315, opened 2024-09-09
fix(data-transfer): incorrect schema file would be generated when merge schema

### LINKED ISSUE #3306: [Bug]: When exporting tables and PLs together in a sql file, and reimport in ODC, syntax error happened.
### ODC version

4.3.0

### OB version

Not related.

### What happened?

As said in title.

Import successfully.

1. export tables and PLs, and export them in one single SQL file
2. import that export file again.

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

When obdumper generate ddl file for PL objects, the format would be like :
```sql
drop procedure xxx;
delimiter $$
create or replace xxx
...
end;
&&
```

If we merge then straightly, the output would like :
```sql
drop procedure xxx;
delimiter $$
create or replace xxx
...
end;
&&

drop procedure xxx;
delimiter $$
create or replace xxx
...
end;
&&
```

As you can see, the format is incorrect. The delimiter should be restore to `;` after PL objects ends.

Fixes #3306

```docs

```


yhilmare (approved):
lgtm

PR #3320, opened 2024-09-09
fix(osc): osc manual swap table button not displayed 

LioRoger:
…sult to fix json deserialize problem

#### What type of PR is this?

type-bug
Revert ScheduleTask to ScheduleTaskEntity of OnlineSchemaChangeTaskResult to fix json deserialize problem.
ScheduleTask contains interface member with type ScheduleTaskParameters. Serialization  not add information about this filed. So JsonUtil.fromJson(String, Class) will failed cause deserializer can't determinate which implementation class should used.
The failure of deserialization will cause the task_task result field missing and that  leads to OSC module's 'enableManualSwap' flag missing in flowinstace/id/tasks/result response. The final result is manual swap flag not present.
image

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/model/OnlineSchemaChangeTaskResult.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/model/OnlineSchemaChangeTaskResult.java
@@ -18,7 +18,7 @@ package com.oceanbase.odc.service.flow.task.model;
 import java.util.List;
 
 import com.oceanbase.odc.core.flow.model.AbstractFlowTaskResult;
-import com.oceanbase.odc.service.schedule.model.ScheduleTask;
+import com.oceanbase.odc.metadb.schedule.ScheduleTaskEntity;
 
 import lombok.AllArgsConstructor;
 import lombok.Data;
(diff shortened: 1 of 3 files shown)


yizhouxw:
PR title duplicated 'fix', 
follow template: `fix(module): bug scenario`

guowl3 (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/model/OnlineSchemaChangeTaskResult.java
```diff
@@ -30,5 +30,5 @@
 @NoArgsConstructor
 @EqualsAndHashCode(callSuper = true)
 public class OnlineSchemaChangeTaskResult extends AbstractFlowTaskResult {
-    private List<ScheduleTask> tasks;
+    private List<ScheduleTaskEntity> tasks;
```
yizhouxw, line 33: Entity class is for persistent layer, may not directly referenced from DTO class

LioRoger: look at my fix comment. ScheduleTask is not designed to deserialize from String. cause  ScheduleTaskParameters is interface and no implementation provided. 
A better solution is  use template or provided a deserialize function that may have a lot duplicate code with ScheduleTaskMapper.entityToParameters(). Both solution needs a lots code change. 
This is patch is provided meanly for concern of stability

yizhouxw (approved):
lgtm

LuckyPickleZZ (approved):
LGTM

PR #3322, opened 2024-09-09
feat(monitor): add prometheus actuator

ungreat:
#### What type of PR is this?

add custom prometheus actuactor，mainly include flow、job、schdule，connect_session and datasource.

Fixes #

```docs

```


ungreat:
> how is the prometheus api authenticated ?

Currently this pr does not support prometheus api 's authentication. To enable actuator, you need to configure the api url to b`odc.web.security.configurable-whitelists`.

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MeterClear.java
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.oceanbase.odc.service.monitor;
+
+public interface MeterClear {
```
LioRoger, line 18: In my opinion, this interface  is not restrict required, make it as function call of MeterHolder

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MeterHolder.java
```diff
@@ -0,0 +1,363 @@
...
+import lombok.Data;
+import lombok.Getter;
+import lombok.Value;
+import lombok.extern.slf4j.Slf4j;
+
+@Component
+@Slf4j
+@Getter
+@ConditionalOnProperty(value = "odc.system.monitor.actuator.enabled", havingValue = "true")
+public class MeterHolder implements MeterClear {
```
LioRoger, line 48: maybe MeterCenter or MeterRegistry or MeterManager is more suitable

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MeterHolder.java
```diff
@@ -0,0 +1,363 @@
...
+
+    private final GaugeHolder gaugeHolder;
```
LioRoger, line 50: stretch holders to function call will reduce a lots duplicate code

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MeterHolder.java
```diff
@@ -0,0 +1,363 @@
...
+            return builder.register(registry);
+        }
+
+        public Integer getSize() {
+            return GAUGE_MAP.size();
+        }
+
+        @Data
+        @AllArgsConstructor
+        static class GaugeWrapper {
```
LioRoger, line 139: declare private access privilege?

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MeterHolder.java
```diff
@@ -0,0 +1,363 @@
...
+            Integer value;
+            Gauge gauge;
+
+            public void increment(Integer increment) {
```
LioRoger, line 143: What this gauge designed for ? if we want maintain a number, register a class with a lamda like () -> {
    c.number++;
}
and this wrapper should not contains Gauge field

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MeterHolder.java
```diff
@@ -0,0 +1,363 @@
...
+        }
+
+        public Integer getSize() {
+            return GAUGE_MAP.size();
+        }
+
+        @Data
+        @AllArgsConstructor
+        static class GaugeWrapper {
+            Integer value;
```
LioRoger, line 140: gauge should be thread safe, use thread safe structure like atomic, by the way, is Integer big enough for metric?

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MeterHolder.java
```diff
@@ -0,0 +1,363 @@
...
+@Slf4j
+@Getter
+@ConditionalOnProperty(value = "odc.system.monitor.actuator.enabled", havingValue = "true")
+public class MeterHolder implements MeterClear {
+
+    private final GaugeHolder gaugeHolder;
+    private final CounterHolder counterHolder;
+    private final TimerHolder timerHolder;
+
+    public MeterHolder(BusinessMeterRegistry meterRegistry,
```
LioRoger, line 54: introduce registerMeter and unregisterMeter to avoid use if null operation.
any regular operation after registerMeter find null value should give some warn

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MeterHolder.java
```diff
@@ -0,0 +1,363 @@
...
+                this.markToRemove = true;
+            }
+        }
+    }
+
+    @Value
+    public static class MeterKey {
+        MeterName meterName;
+        Iterable<Tag> tags;
+        String uniqueKey;
```
LioRoger, line 303: meter Name with tags can't be global unique?
what unique key declared for
need remove also should be MeterKey fields
Generally tag is not easy recorded, Suggest make unique symbols easy to construct and record.

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MeterName.java
```diff
@@ -0,0 +1,83 @@
...
+package com.oceanbase.odc.service.monitor;
+
+import lombok.Getter;
+
+@Getter
+public enum MeterName {
+
+    // flow module
+    FLOW_CREATED_COUNT("flow.created.count", "flow created count"),
+    FLOW_TASK_START_COUNT("flow.task.start.count", "flw task start count"),
```
LioRoger, line 25: type flw -> flow

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MonitorEvent.java
```diff
@@ -0,0 +1,104 @@
...
+import com.oceanbase.odc.service.monitor.task.job.TaskJobMonitorHandler;
+import com.oceanbase.odc.service.monitor.task.schdule.ScheduleMonitorEvent;
+import com.oceanbase.odc.service.monitor.task.schdule.ScheduleMonitorHandler;
+import com.oceanbase.odc.service.schedule.model.ScheduleTaskType;
+
+import lombok.Getter;
+import lombok.NonNull;
+
+@Getter
+public class MonitorEvent<T> extends ApplicationEvent {
```
LioRoger, line 40: the process construct MonitorEvent with various Context has seen the implementation of Metrics. 
If we want add a new module, we will construct new context, add new handler, modify MonitorHandler to recognize the new context type.
let deliver see MeterHolder will reduce this step.

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MonitorEvent.java
```diff
@@ -0,0 +1,104 @@
...
+        return new MonitorEvent<>(SCHEDULE_MODULE, scheduleMonitorEvent);
+    }
+
+    public static MonitorEvent<TaskJobMonitorEvent> createJobMonitorEvent(TaskJobMonitorEvent.Action action) {
+        TaskJobMonitorEvent scheduleMonitorEvent = new TaskJobMonitorEvent(action);
+        return new MonitorEvent<>(JOB, scheduleMonitorEvent);
+    }
+
+
+    public enum MonitorModule {
```
LioRoger, line 88: if sub class not private, make it as a dependent class.

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/monitor/MonitorMetrics.java
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.oceanbase.odc.service.monitor.monitor;
+
+import static com.oceanbase.odc.service.monitor.MeterName.METER_COUNTER_HOLDER_COUNT;
```
LioRoger, line 18: static import is not suggested, cause we can't see what this static field belongs to.
Idea can config to avoid static import

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MeterClear.java
ungreat: Make sence~

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MeterHolder.java
ungreat: Refactor to `MetricManager`

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MeterHolder.java
ungreat: Not using the Observer pattern now

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MeterHolder.java
ungreat: Removed

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MeterHolder.java
ungreat: removed

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MeterHolder.java
ungreat: removed

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MeterHolder.java
ungreat: removed

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MeterHolder.java
ungreat: `Meter Name` and` tag` are unique key of meter registery, however, not unique in manager, cause same meter may have multi sample, we have to distinguish a specific sample from other.

`Tag` is declare by `micrometer`.  add a Builer to easier to construct.

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/monitor/MonitorMetrics.java
ungreat: removed

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MeterName.java
ungreat: amended

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MetricManager.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+package com.oceanbase.odc.service.monitor;
+
+import java.util.function.Supplier;
+
+import io.micrometer.core.instrument.Counter;
+
+public interface MetricManager {
+
+    void registerGauge(MeterKey meterKey, Supplier<Number> f);
```
LioRoger, line 24: as we discussed before, Meter key is a inner concept, register api will return a object like a handler. user use this handler to interact with metricManager

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/InMemoryMetricManager.java
```diff
@@ -0,0 +1,175 @@
...
+        for (Tag tag : tags) {
+            builder.tag(tag.getKey(), tag.getValue());
+        }
+        builder.register(businessMeterRegistry);
+    }
+
+    public void incrementCounter(MeterKey meterKey) {
+        CounterWrapper wrapper = COUNTER_MAP.get(meterKey);
+        if (wrapper == null) {
+            if (COUNTER_MAP.size() >= monitorProperties.getMeter().getMaxCounterMeterNumber()) {
```
LioRoger, line 74: is this interface expected to be thread safe?
multi thread can entry wrapper == null logic to create same meter key

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/InMemoryMetricManager.java
```diff
@@ -0,0 +1,175 @@
...
+            wrapper = COUNTER_MAP.get(lockKey);
+            if (wrapper.getCounter() == null) {
+                Counter counter = registerCounter(meterKey);
+                wrapper.setCounter(counter);
+            }
+            wrapper.getCounter().increment();
+        }
+    }
+
+    private MeterKey getCounterLockKey(MeterKey meterKey) {
```
LioRoger, line 91: why not impl equals and hash to use key look up
find first is not stable

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/InMemoryMetricManager.java
```diff
@@ -0,0 +1,175 @@
...
+    public void startTimerSample(String sampleKey, MeterKey meterKey) {
+        Pair<String, MeterKey> sampleMapKey = Pair.of(sampleKey, meterKey);
+        TimerSampleHolder sampleHolder = this.TIMER_SAMPLE_MAP.get(sampleMapKey);
+        if (sampleHolder == null) {
+            if (this.TIMER_SAMPLE_MAP.size() >= monitorProperties.getMeter().getMaxTimerMeterNumber()) {
+                log.error("Too many Timer to register, meterKey={}", meterKey);
+                return;
+            }
+            // ensure lock same obj
+            this.TIMER_SAMPLE_MAP.computeIfAbsent(sampleMapKey,
```
LioRoger, line 114: if MeterKey not impl equals and hash, find will always return null, cause it's object compare

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/InMemoryMetricManager.java
```diff
@@ -0,0 +1,175 @@
...
+            if (timerSampleHolder == null) {
+                return;
+            }
+            Iterable<Tag> tags = meterKey.getTags();
+            Timer.Builder builder = Timer.builder(meterKey.getMeterName().getMeterName())
+                    .description(meterKey.getMeterName().getDescription());
+            for (Tag tag : tags) {
+                builder.tag(tag.getKey(), tag.getValue());
+            }
+            timerSampleHolder.stop(builder.register(businessMeterRegistry));
```
LioRoger, line 146: suggest change this method as endAndRecord semantics

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MeterKey.java
```diff
@@ -0,0 +1,106 @@
...
+    @Override
+    public int hashCode() {
+        int result = Objects.hash(meterName);
+        Set<Tag> tagSet = toTagSet(tags);
+        result = 31 * result + tagSet.hashCode();
+        return result;
+    }
+
+    private Set<Tag> toTagSet(Iterable<Tag> tags) {
+        Set<Tag> tagSet = new HashSet<>();
```
LioRoger, line 73: iterator to set may not create same order hashSet
in other words, why not use LinkedSet<Tag>  or other collection having order to maintains tags

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/InMemoryMetricManager.java
```diff
@@ -0,0 +1,176 @@
...
+                log.error("Too many Timer to register, meterKey={}", meterKey);
+                return;
+            }
+            // ensure lock same obj
+            this.TIMER_SAMPLE_MAP.computeIfAbsent(sampleMapKey,
+                    k -> new TimerSampleHolder(null));
+        }
+        Pair<String, MeterKey> lockKey = getTimerSampleLockKey(sampleMapKey);
+        synchronized (lockKey) {
+            TimerSampleHolder placeholder = this.TIMER_SAMPLE_MAP.get(lockKey);
```
LioRoger, line 119: lock value has same function for sync control

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/InMemoryMetricManager.java
```diff
@@ -0,0 +1,176 @@
...
+            builder.tag(tag.getKey(), tag.getValue());
+        }
+        return builder.register(businessMeterRegistry);
+    }
+
+    public void startTimerSample(String sampleKey, MeterKey meterKey) {
+        Pair<String, MeterKey> sampleMapKey = Pair.of(sampleKey, meterKey);
+        TimerSampleHolder sampleHolder = this.TIMER_SAMPLE_MAP.get(sampleMapKey);
+        if (sampleHolder == null) {
+            if (this.TIMER_SAMPLE_MAP.size() >= monitorProperties.getMeter().getMaxTimerMeterNumber()) {
```
LioRoger, line 109: As a code designer, if you are afraid of memory leak to add protection code instead of design a better mechanism to prevent memory leak, then the code is hard to maintain
As a user, I thought i have start timer sample, but nothing done, Is that what you expected?

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/InMemoryMetricManager.java
ungreat: Implemented equals and hash

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MeterKey.java
ungreat: LinkedSet only guarantees insertion order, but the requirement is actually order-independent. Now use Tags directly instead.

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/InMemoryMetricManager.java
ungreat: Setting a maximum allowed value is a common practice. For instance, WebMvcMetrics operates similarly, the only difference being the strategy for handling values that exceed the maximum, where, apart from logging, alarm are also added.

LioRoger (dismissed):
ac

yizhouxw (approved):

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MeterName.java
```diff
@@ -0,0 +1,78 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.oceanbase.odc.service.monitor;
+
+import lombok.Getter;
+
+@Getter
+public enum MeterName {
```
yizhouxw, line 21: MetricName ?

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MetricManager.java
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.oceanbase.odc.service.monitor;
+
+import java.util.function.Supplier;
+
+public interface MetricManager {
```
yizhouxw, line 20: MeterManager ?

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MonitorConfig.java
```diff
@@ -0,0 +1,33 @@
...
+package com.oceanbase.odc.service.monitor;
+
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+import com.oceanbase.odc.service.monitor.MonitorAutoConfiguration.NonEnabledMetricManager;
+
+@Configuration
+public class MonitorConfig {
```
yizhouxw, line 25: MonitorConfiguration

ungreat: done~

server/odc-service/src/main/java/com/oceanbase/odc/service/monitor/MeterName.java
ungreat: I think enums should not use the plural form, for instance, TaskType should not be written as TaskTypes.

yizhouxw (approved):
lgtm

LuckyPickleZZ (approved):

guowl3 (approved):
lgtm

yhilmare (approved):
lgtm

PR #3325, opened 2024-09-09
feat(schedule): adaption for support obcloud operation event

abigsaltyfish:
change response, add changeLog for schedule edit

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/model/ChangeScheduleResponse.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/model/ChangeScheduleResponse.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) 2023 OceanBase.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


yizhouxw (commented):
PR title should follow developer guide

server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/model/ChangeScheduleResponse.java
```diff
@@ -0,0 +1,30 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
+import lombok.Data;
+
+/**
+ * @author jingtian
+ * @date 2024/8/26
+ */
+@Data
+public class ChangeScheduleResponse extends Schedule {
```
yizhouxw, line 27: ChangeScheduleResp, please check RESTful API specfication

server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/model/ChangeScheduleResponse.java
```diff
@@ -0,0 +1,30 @@
...
+    @JsonIgnore
+    private ScheduleChangeLog changeLog;
```
yizhouxw, line 29: if @JsonIgnore, then will be skipped while return to client

yizhouxw (commented):
please confirm the target branch is correct

server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/model/ChangeScheduleResponse.java
abigsaltyfish: This field does not require client use

yizhouxw (approved):
lgtm

guowl3 (approved):
lgtm

PR #3361, opened 2024-09-12
fix(permission): failed to execute sql of renaming table

### LINKED ISSUE #3340: [Bug]: A table without table change permission needs to change the table name to display the change permission for a new table
### ODC version

odc432

### OB version

*

### What happened?

image

Create a table without change permission and modify the table name

_No response_

zijiacj:
#### What type of PR is this?
type-bug

image
Refer to the picture above ,we should display only 'ccc' without 'ccc1';

Fixes #3340

```docs

```


yhilmare (approved):
lgtm

PR #3392, opened 2024-09-13
fix(mock-data): failed to mock data at mysql 5.7

### LINKED ISSUE #3108: [Bug]: An error error is reported when constructing simulated data in the native MySQL 5.7 version
### ODC version

4.3.1

### OB version

no related

### What happened?

If you use the native MySQL 5.7 version to simulate data, an error message is reported

Simulation data can be constructed normally

If you use the native MySQL 5.7 version to simulate data, an error message is reported

Urcesr not carried useSSL=false

yhilmare:
#### What type of PR is this?

type-bug

i18n does not work on schedule task, the reason is that the local can not be set when we create a new ticket. we set an i18n key and translate it when we attempt to list or get it.

Fixes #3108

```docs

```


yizhouxw (approved):
lgtm

yizhouxw (approved):

PR #3399, opened 2024-09-13
fix(logicaldatabase): wrong generated ddl of creating and altering logical tables

### LINKED ISSUE #3375: [Bug]: When you create a logical table, you enter a table creation expression and a table creation statement is generated, but the logic database ticket fails to be executed
### ODC version

odc4.3.2

### OB version

no related 

### What happened?

image
image

When you create a logical table, you enter a table creation expression and a table creation statement is generated, but the logic database ticket fails to be executed

When creating logical tables, you should avoid incorrect expressions and table creation statements

When you create a logical table, you enter a table creation expression and a table creation statement is generated, but the logic database ticket fails to be executed

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug
When generating logical table ddl, it should ignore the logical database name, and use the user-defined logical table expression.

Fixes #3375 

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/db/DBTableService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/db/DBTableService.java
@@ -49,6 +49,7 @@ import com.oceanbase.odc.service.session.ConnectConsoleService;
 import com.oceanbase.odc.service.sqlcheck.SqlCheckUtil;
 import com.oceanbase.tools.dbbrowser.DBBrowser;
 import com.oceanbase.tools.dbbrowser.model.DBObjectIdentity;
+import com.oceanbase.tools.dbbrowser.model.DBSchema;
 import com.oceanbase.tools.dbbrowser.model.DBTable;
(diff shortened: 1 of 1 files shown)


yhilmare (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/db/DBTableService.java
```diff
@@ -209,4 +212,18 @@ private TableExtensionPoint getTableExtensionPoint(@NotNull ConnectionSession co
         return SchemaPluginUtil.getTableExtension(connectionSession.getDialectType());
     }
 
+    private void preHandleLogicalTable(DBTable table) {
+        String relation = table.getName();
+        if (StringUtils.isEmpty(relation)) {
+            throw new IllegalArgumentException("logical table expression should not be empty");
+        }
+        String[] segments = StringUtils.split(relation, ".");
```
yhilmare, line 220: how to handle such input: `table_name(0_9)` which does not contain schema

PR #3401, opened 2024-09-13
fix(table permission): index name displays as table name in sql windows check

### LINKED ISSUE #3341: [Bug]: When creating an index for a table without change permission, consider the index name as the table name
### ODC version

odc432

### OB version

*

### What happened?

image

Create a table without change permission and create an index

_No response_

zijiacj:
#### What type of PR is this?
type-bug

image
The sql check window does not display the index

Fixes #3341

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/DBSchemaExtractor.java
```diff
@@ -196,6 +197,16 @@ private static class OBMySQLRelationFactorVisitor extends OBParserBaseVisitor<Re
 
         private final Set<DBSchemaIdentity> identities = new HashSet<>();
 
+        @Override
+        public RelationFactor visitCreate_index_stmt(Create_index_stmtContext ctx) {
+            for (int i = 0; i < ctx.getChildCount(); i++) {
+                if (i != 2) {
```
yhilmare, line 203: magic number, which is not acceptable~
btw, how about alter table add index statement

server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/DBSchemaExtractor.java
```diff
@@ -352,6 +365,17 @@ public RelationFactor visitRename_table_stmt(OBParser.Rename_table_stmtContext c
...
         }
 
+        @Override
+        public RelationFactor visitCreate_index_stmt(OBParser.Create_index_stmtContext ctx) {
+            addRelationFactor(OracleFromReferenceFactory.getRelationFactor(ctx.relation_factor()));
+            return null;
+        }
+
+        @Override
+        public RelationFactor visitAlter_index_stmt(OBParser.Alter_index_stmtContext ctx) {
```
yhilmare, line 375: we should consider  of alter index option

server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/DBSchemaExtractor.java
```diff
@@ -213,6 +215,17 @@ public RelationFactor visitRename_table_stmt(Rename_table_stmtContext ctx) {
...
         }
 
+        @Override
+        public RelationFactor visitCreate_index_stmt(Create_index_stmtContext ctx) {
+            addRelationFactor(MySQLFromReferenceFactory.getRelationFactor(ctx.relation_factor()));
+            return null;
+        }
+
+        @Override
+        public RelationFactor visitAlter_index_option(Alter_index_optionContext ctx) {
```
yhilmare, line 225: is it necessary?

zijiacj: unnecessary, will remove it

yhilmare (approved):
lgtm

PR #3402, opened 2024-09-13
fix(table permission): partition name shouldn't display in the table permission check window

### LINKED ISSUE #3343: [Bug]: Create a partition table without permission to change. When creating a partition, the partition name is considered as the table name
### ODC version

odc432

### OB version

*

### What happened?

image

Create a partition table without permission to change and create a partition

_No response_

zijiacj:
#### What type of PR is this?
type-bug

Fixes #3343

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/DBSchemaExtractor.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/DBSchemaExtractor.java
@@ -43,6 +43,7 @@ import com.oceanbase.tools.sqlparser.adapter.oracle.OracleExpressionFactory;
 import com.oceanbase.tools.sqlparser.adapter.oracle.OracleFromReferenceFactory;
 import com.oceanbase.tools.sqlparser.obmysql.OBParser.Alter_table_actionContext;
 import com.oceanbase.tools.sqlparser.obmysql.OBParser.Create_database_stmtContext;
+import com.oceanbase.tools.sqlparser.obmysql.OBParser.Create_table_stmtContext;
 import com.oceanbase.tools.sqlparser.obmysql.OBParser.Database_factorContext;
 import com.oceanbase.tools.sqlparser.obmysql.OBParser.Dot_relation_factorContext;
 import com.oceanbase.tools.sqlparser.obmysql.OBParser.Normal_relation_factorContext;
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/DBSchemaExtractor.java
```diff
@@ -203,6 +203,9 @@ public RelationFactor visitAlter_table_action(Alter_table_actionContext ctx) {
             if (ctx.RENAME() != null) {
                 return null;
             }
+            if (ctx.alter_partition_option() != null) {
```
yhilmare, line 206: how about creating table ddl?
```sql
create table tbl(
    id int(11)
) partition by range(id) (
    partition p1 values less than(100)
)
```

yhilmare:
btw, this is not the sql check window, it is table permission check window actually

yhilmare:
It is not partition table name, please use partition name instead

server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/DBSchemaExtractor.java
zijiacj: have done

zijiacj:
> It is not partition table name, please use partition name instead

ok

zijiacj:
> btw, this is not the sql check window, it is table permission check window actually

ok

server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/DBSchemaExtractor.java
```diff
@@ -198,11 +199,23 @@ private static class OBMySQLRelationFactorVisitor extends OBParserBaseVisitor<Re
 
         private final Set<DBSchemaIdentity> identities = new HashSet<>();
 
+        @Override
+        public RelationFactor visitCreate_table_stmt(Create_table_stmtContext ctx) {
+            if (ctx.partition_option() != null) {
+                addRelationFactor(MySQLFromReferenceFactory.getRelationFactor(ctx.relation_factor()));
```
yhilmare, line 205: why should we add relation factor here?

zijiacj: Skipping over the ‘partition’

yhilmare (approved):
lgtm

PR #3417, opened 2024-09-14
fix(taskframework): cannot release resource when logfile upload failed

guowl3:
#### What type of PR is this?
type-bug

Release task resources whether the log upload is successful.

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/server/TaskMonitor.java
```diff
@@ -216,13 +218,14 @@ private void uploadLogFileToCloudStorage(DefaultTaskResult finalResult) {
...
-            Map<String, String> logMap = null;
+            Map<String, String> logMap = new HashMap<>();
             try {
                 logMap = biz.uploadLogFileToCloudStorage(finalResult.getJobIdentity(), cloudObjectStorageService);
+                logMap.put(JobAttributeKeyConstants.UPLOAD_LOG_STATUS, UploadLogStatus.SUCCESS);
             } catch (Throwable e) {
                 log.warn("Upload job log file to cloud storage occur error, jobId={}", getJobId(), e);
                 // putAll will throw NPE if it returns null.
-                logMap = new HashMap<>();
+                logMap.put(JobAttributeKeyConstants.UPLOAD_LOG_STATUS, UploadLogStatus.FAILED);
```
yizhouxw, line 228: may use exists LOG_STORAGE_FAILED_REASON attribute

guowl3: removed and fixed the issue of old fields were not effective

yizhouxw (approved):
lgtm

PR #3418, opened 2024-09-14
feat(text2sql): add a thread pool for copilot

abigsaltyfish:
Put the thread pool under ODC management

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/config/ScheduleConfiguration.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/config/ScheduleConfiguration.java
@@ -317,6 +317,22 @@ public class ScheduleConfiguration {
         return executor;
     }
 
+    @Bean(name = "copilotQueryExecutor")
+    public ThreadPoolTaskExecutor copilotQueryExecutor() {
+        ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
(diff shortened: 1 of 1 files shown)


yhilmare (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/config/ScheduleConfiguration.java
```diff
@@ -317,6 +317,22 @@ public ThreadPoolTaskExecutor queryProfileMonitorExecutor() {
         return executor;
     }
 
+    @Bean(name = "copilotQueryExecutor")
```
yizhouxw, line 320: 1. why only thread pool, no implementation?

2. if copilot is included in odc, we may add a switch configuration, e.g. odc.features.copilot.enabled

abigsaltyfish: implementation in odc-cloud, switch configuration will be added after the ODC supports Copilot in the future

yizhouxw: if implemented in odc-cloud, why create thread pool here ?

better add CloudScheduleConfiguration in odc-cloud instead.

abigsaltyfish: ok

PR #3419, opened 2024-09-14
fix(structure-compare): show incorrect precision of tinyint datatype in mysql mode

### LINKED ISSUE #3395: [Bug]: The table structure synchronization result is inconsistent
### ODC version

4.3.1

### OB version

mysql

### What happened?

The table structure synchronization result is inconsistent.

The table structure synchronization result is consistent.

1. create table in source database
```
create table `tinyint_test`(`del_flag` tinyint(4) UNSIGNED  ZEROFILL  DEFAULT '0' COMMENT '0 正常 1 删除' NOT NULL);
```
2. create structure comparison task
3. execute comparison result
4. create structure comparison task again

_No response_

### LINKED ISSUE #3415: [Bug]: the precision of the tinyint type is incorrect in mysql mode
### ODC version

all

### OB version

mysql 5.7

### What happened?

the precision of the tinyint type is incorrect in mysql mode:
table ddl:
image
odc show precision:
image

show the precision of the tinyint type correctly

create table with tinyint(n), n can be any, odc will always show precision 3:
```
create table `tinyint_test7`(`del_flag` tinyint(4) UNSIGNED  ZEROFILL  DEFAULT '0' COMMENT '0 正常 1 删除' NOT NULL);
```

_No response_

PeachThinking:
#### What type of PR is this?
type-bug
module-structure compare

Regardless of the precision defined for the `tinyint` field type in the table's ddl, the value of the `NUMERIC_PRECISION` field in the `information_schema.columns` table is always be 3.
table ddl:
```
create table `tinyint_test`(`del_flag` tinyint(4) UNSIGNED ZEROFILL DEFAULT '0' );
```
query result：
image
so we can get correctly precision by parse `COLUMN_TYPE`.
we already have a method called `fillPrecisionAndScale` which only used in `OBMySQLNoGreaterThan1479SchemaAccessor`, so i remove the method to `MySQLNoLessThan5700SchemaAccessor`

Fixes #3395
Fixes #3415

```docs

```


server/integration-test/src/test/java/com/oceanbase/odc/service/structurecompare/mysql/DefaultDBStructureComparatorTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+        jdbcTemplate.execute("use `" + sourceSchemaName + "`");
+        jdbcTemplate.execute(sourceDrop);
+        jdbcTemplate.execute(sourceSchemaDdl);
+        jdbcTemplate.execute("use `" + targetSchemaName + "`");
+        jdbcTemplate.execute(targetDrop);
+        jdbcTemplate.execute(targetSchemaDdl);
+        jdbcTemplate.execute("use `" + configuration.getDefaultDBName() + "`");
+
+        results = comparator.compare(getSourceConfig(srcConfiguration),
```
yizhouxw, line 87: under test method should called in test case, not in setUp

PeachThinking: In fact, all compare methods of the structure comparison class are implemented in setUp. This is because we only need to read the ddl script once and compare it once to obtain the comparison results of all unit tests, which can also improve the efficiency of unit testing.

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoLessThan5700SchemaAccessor.java
```diff
@@ -110,6 +110,18 @@ public class MySQLNoLessThan5700SchemaAccessor implements DBSchemaAccessor {
 
     protected JdbcOperations jdbcOperations;
     protected DBSchemaAccessorSqlMapper sqlMapper;
+    private static final Set<String> SPECIAL_TYPE_NAME = new HashSet<>();
```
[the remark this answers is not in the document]
PeachThinking: ok~

yhilmare (approved):
lgtm

server/integration-test/src/test/java/com/oceanbase/odc/service/structurecompare/mysql/DefaultDBStructureComparatorTest.java
yizhouxw: the basic concept is wrong here

yizhouxw: if only once comparing, the only one test case

PeachThinking: but each test case is written for a different test scenario, and it is not appropriate to write all of them in one test method. I think just consider the test class for structure comparison  as a special scenario, which compares all table structures at once.

yizhouxw (approved):
lgtm

yhilmare (approved):
lgtm

PR #3428, opened 2024-09-18
feat(response): add batch operation response

CHLK:
#### What type of PR is this?
type-feature

add batch operation response

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/common/response/batch/BatchOperationResponseData.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/common/response/batch/BatchOperationResponseData.java
@@ -26,25 +26,25 @@ import lombok.Data;
  * @since 4.3.2
  */
 @Data
-public class BatchOperateResponseData {
+public class BatchOperationResponseData {
     private int total;
     private int successCount;
-    private int                              failedCount;
-    private List<BatchOperateResponseDetail> details;
+    private int                                failedCount;
+    private List<BatchOperationResponseDetail> details;
 
-    public BatchOperateResponseData() {
+    public BatchOperationResponseData() {
         this.total = 0;
         this.successCount = 0;
         this.failedCount = 0;
     }
 
-    public void addSuccess(Collection<BatchOperateResponseDetail> details) {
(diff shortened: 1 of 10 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/common/response/batch/BatchOperationResponseData.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.List;
+
+import lombok.Data;
+
+/**
+ * @author keyang
+ * @date 2024/09/18
+ * @since 4.3.2
+ */
+@Data
```
yizhouxw, line 28: BatchOperationResult

server/odc-service/src/main/java/com/oceanbase/odc/service/common/response/batch/BatchOperationResponseDetail.java
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * @author keyang
+ * @date 2024/09/18
+ * @since 4.3.2
+ */
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+@Builder
+public class BatchOperationResponseDetail {
```
yizhouxw, line 35: ResourceOperationResult

server/odc-service/src/main/java/com/oceanbase/odc/service/common/response/batch/BatchOperationResponseData.java
CHLK: okay, i'll make this change

server/odc-service/src/main/java/com/oceanbase/odc/service/common/response/batch/BatchOperationResponseDetail.java
CHLK: okay, i'll make this change

CHLK:
@yizhouxw I've implemented your suggestion, please take a look when you have a chance

yizhouxw (approved):
lgtm

yhilmare (approved):
lgtm

PR #3431, opened 2024-09-18
fix(response): error dependency for Error in ResourceOperationResult

CHLK:
#### What type of PR is this?
type-bug

import com.oceanbase.odc.service.common.response.Error for Error in ResourceOperationResult;
add method add in BatchOperationResult

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #3437, opened 2024-09-19
fix(osc): new create table ddl not display when use alter table mode

LioRoger:
#### What type of PR is this?

type-bug
set ScheduleTask.OnlineSchemaChangeScheduleTaskParameters.newTableCreateDdlForDisplay to correct value when use alter table mode.

Fixes #
1. Front end fetch newTableDDL in ScheduleTask.OnlineSchemaChangeScheduleTaskResult.newTableDdl
2. newTableDdl is assigned by OnlineSchemaChangeScheduleTaskParameters.newTableCreateDdlForDisplay
3. if OnlineSchemaChangeScheduleTaskParameters.newTableCreateDdlForDisplay not set, OnlineSchemaChangeScheduleTaskResult.newTableDdl will be covered to null
4. this patch set newTableCreateDdlForDisplay to guarantee the following operation always fetch right value for newTableDdl

```docs

```


yhilmare (approved):
lgtm

LuckyPickleZZ (approved):

PR #3455, opened 2024-09-19
fix(schedule): restore deleted method

LuckyPickleZZ:
The `com.oceanbase.odc.service.schedule.ScheduleTaskService#listByJobNames` method was deleted in PR https://github.com/oceanbase/odc/pull/3335 .
But we still need this. So  restore it.


yhilmare (approved):
lgtm

PR #3467, opened 2024-09-20
feat(connection): add missing fields in OBInstance and ObTenant

CHLK:
#### What type of PR is this?
type-feature

add missing fields in OBInstance and ObTenant;
add WORKSPACE_COUNT_IN_PROJECT in LimitMetric;

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/OBTenantStatus.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/OBTenantStatus.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2023 OceanBase.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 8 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/OBTenantStatus.java
```diff
@@ -0,0 +1,55 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonValue;
+
+/**
+ * copy from obcloud-datasource, and package is
+ * com.oceanbase.obcloud.datasource.integration.ocp.model.OBTenantStatusResp <br>
+ * OB 租户的状态，来源于 OCP 的TenantStatus 枚举值<br>
+ * 这里将OCP的所有状态分为两类：ONLINE 和 NOT_CONNECTABLE
+ */
+public enum OBTenantStatus {
```
PeachThinking, line 30: Do you need to sense the tenant status? DSS needs to sense it to determine whether the tenant status supports connection.

PR #3490, opened 2024-09-20
fix(permission): logical database/table permission not shown

MarkPotato777:
#### What type of PR is this?

type-bug
make the view `list_user_database_permission_view` and `list_user_table_permission_view` DDL from `inner join connect_connection` to `left join connect_connection`, because the logical database does not belong to a connection, and the connection_id will be null.

Fixes #

```docs

```


yizhouxw (approved):
lgtm

yhilmare (approved):
lgtm

PR #3492, opened 2024-09-20
fix(logicaldatabase): physical databases who have already belonged to a logical database should not be able to configure again

### LINKED ISSUE #3219: [Bug]: Create a schema, select the referenced schema, and click Submit, but an error message is reported that it needs to be optimized, and it is not clear
### ODC version

4.3.2

### OB version

no related

### What happened?

Create a schema, select the referenced schema, and click Submit, but an error message is reported that it needs to be optimized, and it is not clear
image

When creating a logical library, select the physical library that has been referenced, and make the prompt clearer

Create a schema, select the referenced schema, and click Submit, but an error message is reported that it needs to be optimized, and it is not clear

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug
This PR add a precheck that guarantee  physical databases who have already belonged to a logical database should not be able to configure again/

Fixes #3219 

```docs

```


yizhouxw (approved):
lgtm

PR #3495, opened 2024-09-20
fix(logicaldatabase): generate wrong table topologies when logical expression is like db.tb_[00-10]

### LINKED ISSUE #3366: [Bug]: If the name of the physical database is test10, it will be changed to test010 when the table is created and parsed
### ODC version

odc 4.3.2

### OB version

no related

### What happened?

If the name of the physical database is test10, it will be changed to test010 when the table is created and parsed
image
image

The parsing should be correct

Create a physics library test_01~test10, select the physics library of 01~10 as the logic library, and create an expression that is sharded without table sharding, and test_table[01-10].testThe system prompts that the expression is incorrect

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug
Generate wrong table topologies when logical expression is like db.tb_[00-10]. 
This is because the wrong logic of calculating leading zero.

Fixes #3366 

```docs

```


yizhouxw (approved):
lgtm

PR #3496, opened 2024-09-20
fix(logicaldatabase): generate wrong ddl when alter table

### LINKED ISSUE #3448: [Bug]: An error message is reported in a field added to a logical table
### ODC version

odc 4.3.2

### OB version

no related

### What happened?

Select a logical table, click View Table Structure, and add fields, indexes, constraints, and system errors in the Table Information display box
image
Image

Logical tables can add fields, indexes, and constraints normally.

Add fields, indexes, and constraints to the logical table, click Next to confirm the SQL, and an error is reported

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug
When creating tables, users need to enter a logical table expression, so we need to preHandle the schema and table names; while altering tables, we should not pre handle it because we only need to treat the logical table as the physical table.

Fixes #3448 

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/logicaldatabase/LogicalTableExtractTask.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/logicaldatabase/LogicalTableExtractTask.java
@@ -143,7 +143,6 @@ public class LogicalTableExtractTask implements Runnable {
                     physicalTableEntity.setOrganizationId(logicalDatabase.getOrganizationId());
                     physicalTableEntity.setPhysicalDatabaseId(dataNode.getDatabaseId());
                     physicalTableEntity.setPhysicalDatabaseName(dataNode.getSchemaName());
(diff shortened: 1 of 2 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/connection/logicaldatabase/LogicalTableExtractTask.java
```diff
@@ -143,7 +143,6 @@ public void run() {
                     physicalTableEntity.setOrganizationId(logicalDatabase.getOrganizationId());
                     physicalTableEntity.setPhysicalDatabaseId(dataNode.getDatabaseId());
                     physicalTableEntity.setPhysicalDatabaseName(dataNode.getSchemaName());
-                    physicalTableEntity.setLogicalTableId(dataNode.getTableId());
```
MarkPotato777, line 146: Fix wrongly setting the logicalTableId which is a typo/mistake

PR #3499, opened 2024-09-20
fix(database): the databases list should be ordered by type and name by default

### LINKED ISSUE #3489: [Bug]: The logic base in the project should be displayed in the first column for easy operation
### ODC version

odc 4.3.2

### OB version

no related

### What happened?

image
If there are many physical libraries, the logical library needs to turn the page to operate, and the logical library should be displayed in front of the physical library

If there are many physical libraries, the logical library needs to turn the page to operate, and the logical library should be displayed in front of the physical library

Configure multiple schema databases that contain physical libraries and observe the sorting and display of logic databases

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug
The databases list should be order by type and name by default. As the LOGICAL_DATABSE > PHYSICAL_DATABASE，so we just need add type in the front of the name in sort, the direction is ASC.

Fixes #3489 

```docs

```


yhilmare (approved):
lgtm

PR #3500, opened 2024-09-20
fix(logicaldatabase): add max physical database count limit in one logical database

MarkPotato777:
#### What type of PR is this?

type-bug
Add a precheck that the physical database count should be less than or equals to 1000.

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #3501, opened 2024-09-20
fix(logicaldatabase): failed to recognize the wrong expression like db.tb.xxx

### LINKED ISSUE #3463: [Bug]: If you enter an incorrect logical table expression, there is no indication below the expression that the expression is incorrect, but when you click it, submit it, and confirm the SQL statement, the system reports an error
### ODC version

4.3.2

### OB version

no related

### What happened?

image
If you enter an incorrect logical table expression, there is no indication below the expression that the expression is incorrect, but when you click it, submit it, and confirm the SQL statement, the system reports an error

Fill in the wrong logical table creation expression, qym[1-2].test[35-56].test, click Submit and confirm the SQL, and click Submit to confirm that the SQL is not clickable, and there should be a hint that the syntax is incorrect

Fill in the wrong logical table creation expression, qym[1-2].test[35-56].test, click Submit and confirm the SQL, and click Submit to confirm that the SQL is not clickable, and there should be a hint that the syntax is incorrect

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug
`db.tb.xxx ` should be a bad syntax expression. I edited .g4 to fix it.

Fixes #3463 

```docs

```

### DIFF
--- a/server/integration-test/src/test/resources/connection/logicaldatabase/invalid_logical_table_expression_resolve.yaml
+++ b/server/integration-test/src/test/resources/connection/logicaldatabase/invalid_logical_table_expression_resolve.yaml
@@ -41,3 +41,11 @@
 - id: 11
   expression: "db.tb_[1-2147483648:1]"
   error_message_abstract: "In your expression, the range for [1-2147483648:1] must consist of valid positive integers."
+
+- id: 12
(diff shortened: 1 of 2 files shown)


yhilmare (approved):
lgtm

Issue #3514, opened 2024-09-23
[Bug]: The project administrator starts a schema change task and says that there are no permissions


qymsummer:
### ODC version

4.3.2

### OB version

no related

### What happened?

If you use the project administrator user to start a schema change task, you will be prompted that you have no permissions and the task has not been running
![image](https://github.com/user-attachments/assets/6dc9b03b-28a9-4267-b6b0-f0bbff83ca95)
![image](https://github.com/user-attachments/assets/4dfd02db-a85d-4789-9224-f53c90cbd82f)

### What did you expect to happen?

Using the Project Administrator user, start a database change task that can be executed successfully

### How can we reproduce it (as minimally and precisely as possible)?

A user who does not have any role in a working project, but is an administrator in the project, initiates a database change task.

sl01388797:
![Image](https://github.com/user-attachments/assets/d5d5cd73-e0e8-4708-9f80-66e43b91e7a5)
pass

PR #3515, opened 2024-09-23
fix(logicaldatabase): when sql contains physical databases that does not belong to the logical database, the front-end crashes

### LINKED ISSUE #3354: [Bug]: In the Logic Database Table Creation ticket, click Preview SQL page and an error is reported
### ODC version

odc 4.3.2

### OB version

no related

### What happened?

image

image
In the Logic Database Table Creation ticket, click Preview SQL page and an error is reported

The actual SQL statement can be displayed normally

In the Logic Database Table Creation ticket, click Preview SQL page and an error is reported

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug
1. When the user executes SQL containing physical databases that do not belong to the logical database, the front-end will crash. The solution is to throw the exception instead and notify users that the database is not found.
2. skip the permission check when publishing tasks to the task framework.

Fixes #3354 #3514 

```docs

```


yhilmare (approved):
lgtm

PR #3520, opened 2024-09-23
fix(logicaldatabase): terminal sub-tasks failed

### LINKED ISSUE #3517: [Bug]: Select a task and click Terminate, but the task is not terminated
### ODC version

4.3.2

### OB version

no related

### What happened?

image
Select a task and click Terminate, the task will continue to run until the run is completed

Click Terminate and the task should change to Terminating or Terminating Failed

Construct a database change task, 

_No response_

MarkPotato777:
Terminal sub-tasks failed. The root cause is to get the wrong parameters. And I optimized terminal logic, making it repeatable.

Fixes #3517 

```docs

```


yhilmare (approved):
lgtm

PR #3526, opened 2024-09-23
refactor(resource): refactor the logic of getting related pods of a deployment

yhilmare:
#### What type of PR is this?

type-refactor

The original method of obtaining the pod associated with a deployment is to use the name of the deployment as a prefix to filter, which is not rigorous.

The current filtering method is to use the deployment selector to filter, which can be more accurate

Fixes #

```docs

```


yizhouxw (approved):
lgtm

PR #3539, opened 2024-09-24
fix(resource): the status will be 'UNKNOWN' when resource is not created immediately

yhilmare:
#### What type of PR is this?

type-bug

the k8s can not create a resource immediately, the dag will be moved to 'UNKNOWN' if the current state is 'CREATING'

Fixes #

```docs

```


yizhouxw (dismissed):
lgtm

yizhouxw (approved):

MarkPotato777 (approved):

PR #3545, opened 2024-09-24
fix(task-framework): starting task fails when environment variables exceed 2MB in Process Mode

MarkPotato777:
#### What type of PR is this?

type-bug
Starting task fails when environment variables exceed 2MB in Process Mode.
The solution is that write `JobContext` into file, and set the file path into the environment variable, instead of setting the `JobContext` into the environment variable directly.
The solution is 

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/JobCallerBuilder.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/JobCallerBuilder.java
@@ -48,7 +48,8 @@ public class JobCallerBuilder {
         String jobContextFilePath = JobUtils.getExecutorDataPath() + "/" + StringUtils.uuid() + ".enc";
         try {
             FileUtils.writeStringToFile(new File(jobContextFilePath),
-                    environments.get(JobEnvKeyConstants.ODC_JOB_CONTEXT),
+                    JobUtils.encrypt(environments.get(JobEnvKeyConstants.ENCRYPT_KEY),
+                            environments.get(JobEnvKeyConstants.ENCRYPT_SALT), JobUtils.toJson(context)),
                     Charset.defaultCharset());
         } catch (Exception ex) {
             FileUtils.deleteQuietly(new File(jobContextFilePath));
@@ -57,8 +58,6 @@ public class JobCallerBuilder {
         environments.put(JobEnvKeyConstants.ODC_JOB_CONTEXT_FILE_PATH,
                 JobUtils.encrypt(environments.get(JobEnvKeyConstants.ENCRYPT_KEY),
                         environments.get(JobEnvKeyConstants.ENCRYPT_SALT), jobContextFilePath));
-        // remove JobContext from environments
-        environments.remove(JobEnvKeyConstants.ODC_JOB_CONTEXT);
         ProcessConfig config = new ProcessConfig();
         config.setEnvironments(environments);
 
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/JobCallerBuilder.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/JobCallerBuilder.java
@@ -15,8 +15,13 @@
  */
 package com.oceanbase.odc.service.task.caller;
 
+import java.io.File;
+import java.nio.charset.Charset;
 import java.util.Map;
 
+import org.apache.commons.io.FileUtils;
+
+import com.oceanbase.odc.common.util.StringUtils;
 import com.oceanbase.odc.service.task.config.JobConfigurationHolder;
 import com.oceanbase.odc.service.task.config.TaskFrameworkProperties;
 import com.oceanbase.odc.service.task.constants.JobEnvKeyConstants;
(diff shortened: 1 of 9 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/TaskApplication.java
```diff
@@ -24,6 +24,7 @@
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.core.LoggerContext;
+import org.eclipse.jgit.util.StringUtils;
```
LioRoger, line 27: is apache.common.StringUtils you wanted?

server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/JobCallerBuilder.java
```diff
@@ -36,7 +41,24 @@ public class JobCallerBuilder {
     public static JobCaller buildProcessCaller(JobContext context) {
         Map<String, String> environments = new JobEnvironmentFactory().build(context, TaskRunMode.PROCESS);
         JobUtils.encryptEnvironments(environments);
-
+        /**
+         * write JobContext to file in case of exceeding the environments size limit; set the file path in
+         * the environment instead
+         */
+        String jobContextFilePath = JobUtils.getExecutorDataPath() + "/" + StringUtils.uuid() + ".enc";
+        try {
```
LioRoger, line 49: Suggest create dir operation before write in case some platform may not support create file in dir not exits

server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/JobCallerBuilder.java
```diff
@@ -36,7 +41,24 @@ public class JobCallerBuilder {
...
+                    Charset.defaultCharset());
+        } catch (Exception ex) {
+            FileUtils.deleteQuietly(new File(jobContextFilePath));
+            throw new RuntimeException("Failed to write job context to file: " + jobContextFilePath, ex);
+        }
+        environments.put(JobEnvKeyConstants.ODC_JOB_CONTEXT_FILE_PATH,
+                JobUtils.encrypt(environments.get(JobEnvKeyConstants.ENCRYPT_KEY),
+                        environments.get(JobEnvKeyConstants.ENCRYPT_SALT), jobContextFilePath));
+        // remove JobContext from environments
+        environments.remove(JobEnvKeyConstants.ODC_JOB_CONTEXT);
```
LioRoger, line 61: this remove operation is wired, why not put it in env build function?

server/odc-service/src/main/java/com/oceanbase/odc/service/task/util/JobUtils.java
```diff
@@ -180,4 +180,12 @@ public static void putEnvToSysProperties(String environmentKey) {
     public static void encryptEnvironments(Map<String, String> environments) {
         new JobEnvironmentEncryptor().encrypt(environments);
     }
+
+    public static String encrypt(String key, String salt, String raw) {
```
LioRoger, line 184: JobEnvironmentEncryptor means encrypt. but is contains decrypt method. 
if we don't implement JobEnvironmentDecryptor,  
move  encrypt and decrypt method from JobEnvironmentEncryptor. to here is a better choice

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/context/ProcessJobContextProvider.java
```diff
@@ -0,0 +1,52 @@
...
+ */
+public class ProcessJobContextProvider implements JobContextProvider {
+    @Override
+    public JobContext provide() {
+        String encryptedJobContextJson;
+        try {
+            encryptedJobContextJson = FileUtils
+                    .readFileToString(new File(System.getProperty(JobEnvKeyConstants.ODC_JOB_CONTEXT_FILE_PATH)),
+                            Charset.defaultCharset());
+        } catch (Exception ex) {
```
LioRoger, line 43: this logic is same with   JobCallerBuilder, but they are different handle branch. try catch VS try catch finally. unit it

server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/JobCallerBuilder.java
MarkPotato777: `apache.commons.io.FileUtils#writeStringToFile` will help create parent directories if not exist, so it's safe here.

server/odc-service/src/main/java/com/oceanbase/odc/service/task/util/JobUtils.java
MarkPotato777: Yes, it's wired. But before I added this `decrypt` method, it has already existed a method `decrypt`, so I followed this. And you could also see `TextEncryptor`

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/context/ProcessJobContextProvider.java
MarkPotato777: The logic is not actually the same. We can always delete the file in `ProcessJobContextProvider` after reading the file, but we cannot delete the file in JobCallerBuilder if the exception not thrown. We only delete the file when writing files failed; if success, we keep the file.

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/TaskApplication.java
MarkPotato777: Yes I should use `com.oceanbase.odc.common.util.StringUtils`. Fixed it now.

server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/JobCallerBuilder.java
MarkPotato777: Fixed it.

yizhouxw (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/JobCallerBuilder.java
```diff
@@ -36,7 +41,23 @@ public class JobCallerBuilder {
     public static JobCaller buildProcessCaller(JobContext context) {
         Map<String, String> environments = new JobEnvironmentFactory().build(context, TaskRunMode.PROCESS);
         JobUtils.encryptEnvironments(environments);
-
+        /**
+         * write JobContext to file in case of exceeding the environments size limit; set the file path in
+         * the environment instead
+         */
+        String jobContextFilePath = JobUtils.getExecutorDataPath() + "/" + StringUtils.uuid() + ".enc";
```
yhilmare, line 48: .env?

server/odc-service/src/main/java/com/oceanbase/odc/service/task/caller/JobEnvironmentFactory.java
```diff
@@ -44,8 +44,9 @@ public class JobEnvironmentFactory {
     public Map<String, String> build(JobContext context, TaskRunMode runMode) {
         putEnv(JobEnvKeyConstants.ODC_BOOT_MODE, () -> JobConstants.ODC_BOOT_MODE_EXECUTOR);
         putEnv(JobEnvKeyConstants.ODC_TASK_RUN_MODE, runMode::name);
-        putEnv(JobEnvKeyConstants.ODC_JOB_CONTEXT, () -> JobUtils.toJson(context));
-
+        if (runMode.isK8s()) {
+            putEnv(JobEnvKeyConstants.ODC_JOB_CONTEXT, () -> JobUtils.toJson(context));
```
yhilmare, line 48: will k8s encounter the same problem that the process mode does?

PR #3559, opened 2024-09-25
fix(schema): enable syncing databases by datasource in sql console

### LINKED ISSUE #3442: [Bug]: Synchronous metadb function of object tree list has delay
### ODC version

ODC 4.3.2

### OB version

No related.

### What happened?

I created a database in the SQL console and clicked the synchronization metabase button of the object tree, but the object tree did not show the new database.
image

After clicking Synchronize Metabase, the database list is updated in time.

Follow what happened.

_No response_

zijiacj:
#### What type of PR is this?
type-bug

Fixes #3442

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/db/schema/DBSchemaIndexService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/db/schema/DBSchemaIndexService.java
@@ -64,6 +64,7 @@ import com.oceanbase.odc.service.iam.auth.AuthenticationFacade;
 import com.oceanbase.tools.dbbrowser.model.DBObjectType;
 
 import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
 
 /**
  * @author gaoda.xy
@@ -72,6 +73,7 @@ import lombok.NonNull;
 @Service
 @Validated
 @SkipAuthorize("permission check inside")
+@Slf4j
 public class DBSchemaIndexService {
 
     @Autowired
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/db/schema/DBSchemaIndexService.java
```diff
@@ -201,6 +203,12 @@ public Boolean syncDatabaseObjects(@NonNull @Valid SyncDBObjectReq req) {
         this.databaseService.refreshExpiredPendingDBObjectStatus();
         Set<Database> databases = new HashSet<>();
         if (req.getResourceType() == ResourceType.ODC_CONNECTION) {
+            try {
+                databaseService.internalSyncDataSourceSchemas(req.getResourceId());
+            } catch (InterruptedException ex) {
```
yhilmare, line 208: catch exception instead

server/odc-service/src/main/java/com/oceanbase/odc/service/db/schema/DBSchemaIndexService.java
```diff
@@ -201,6 +203,12 @@ public Boolean syncDatabaseObjects(@NonNull @Valid SyncDBObjectReq req) {
...
+                log.warn("sync all data bases of data base source: " + req.getResourceId() + " failed, errorMessage={}",
```
yhilmare, line 209: log.warn("sync data sources failed, datasourceId={}, errorMessage={}", req.getResourceID(), ex.getMsg())

server/odc-service/src/main/java/com/oceanbase/odc/service/db/schema/DBSchemaIndexService.java
zijiacj: It is recommended to return the error directly if not getting the lock, because if not getting the lock, it may not be the latest database collection in odc, so it is meaningless to synchronize the database object like table ,view etc later.
What's more, even if successful, the list of databases is still not up to date.

server/odc-service/src/main/java/com/oceanbase/odc/service/db/schema/DBSchemaIndexService.java
zijiacj: ok

PeachThinking (commented):
lgtm

PR #3560, opened 2024-09-25
fix(logicaldatabase): generated create logical table ddl is wrong

MarkPotato777:
#### What type of PR is this?

type-bug
When creating a table, the entire expression should be enclosed in backticks, rather than adding one pair for the database and another for the table. For example, it should be `create table db_[0-3].tb_[0-3](a varchar(32))` instead of `create table db_[0-3].tb_[0-3](a varchar(32))`.

Reasons:

It's more logical as the expression itself is a whole.
We are supporting enumeration syntax, for example, `create table db_1.tb_[1,2], db_2.tb_[7,8]`, where the expression must be treated as a whole before parsing.

Fixes #

```docs

```


PeachThinking (approved):

yhilmare (approved):
lgtm

PR #3565, opened 2024-09-25
fix(flow): filter out all ALTER_SCHEDULE flow instances in 'all tickets', 'created by me' and 'approving by me'

MarkPotato777:
#### What type of PR is this?

type-bug
Filter out all ALTER_SCHEDULE flow instances in 'all tickets', 'created by me' and 'approving by me'.
The flow instance that created by schedule should be not seen in the tickets. If users want to view them, go to the schedule module.

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
@@ -471,7 +471,6 @@ public class FlowInstanceService {
                     TaskType.SHADOWTABLE_SYNC,
                     TaskType.PARTITION_PLAN,
                     TaskType.ONLINE_SCHEMA_CHANGE,
-                    TaskType.ALTER_SCHEDULE,
                     TaskType.EXPORT_RESULT_SET,
                     TaskType.APPLY_PROJECT_PERMISSION,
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
```diff
@@ -471,7 +471,6 @@ public Page<FlowInstanceEntity> listAll(@NotNull Pageable pageable, @NotNull Que
                     TaskType.SHADOWTABLE_SYNC,
                     TaskType.PARTITION_PLAN,
                     TaskType.ONLINE_SCHEMA_CHANGE,
-                    TaskType.ALTER_SCHEDULE,
```
yhilmare, line 474: @guowl3 may check this

yhilmare: how can we process alter schedule ticket if you filter out them?

MarkPotato777: Users need to go to the schedule module to view the tickets.

yizhouxw (approved):
lgtm

PR #3592, opened 2024-09-26
fix(dlm): OOM occurs when a table's unique key contains nullable columns

guowl3:
#### What type of PR is this?
type-bug

validate all columns is no null for unique key in mysql.

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #3595, opened 2024-09-27
fix(notification): incorrent status of DLM task notification

### LINKED ISSUE #3591: [Bug]: The archiving task failed to execute, but the message was sent successfully
### ODC version

ODC432

### OB version

independent

### What happened?

The archiving task failed to execute, but the message was sent successfully
image

。

。

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

We've added a hook for notification in `JobTerminateNotifyListener` , and determine the type of notification to be sent based on the status in `job_job` table.
But for DLM tasks , the job status is Incorrect. The real status should be found in `schedule_task`.

Fixes #3591

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/task/listener/JobTerminateNotifyListener.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/task/listener/JobTerminateNotifyListener.java
@@ -19,13 +19,14 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 import com.oceanbase.odc.common.event.AbstractEventListener;
+import com.oceanbase.odc.core.shared.constant.TaskStatus;
 import com.oceanbase.odc.metadb.task.JobEntity;
+import com.oceanbase.odc.service.dlm.DLMService;
 import com.oceanbase.odc.service.notification.Broker;
 import com.oceanbase.odc.service.notification.NotificationProperties;
 import com.oceanbase.odc.service.notification.helper.EventBuilder;
 import com.oceanbase.odc.service.schedule.ScheduleService;
 import com.oceanbase.odc.service.schedule.ScheduleTaskService;
-import com.oceanbase.odc.service.task.enums.JobStatus;
 import com.oceanbase.odc.service.task.service.TaskFrameworkService;
 
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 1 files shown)


guowl3 (approved):
lgtm

yizhouxw (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/task/listener/JobTerminateNotifyListener.java
```diff
@@ -60,7 +63,11 @@ public void onEvent(JobTerminateEvent event) {
             JobEntity jobEntity = taskFrameworkService.find(event.getJi().getId());
             scheduleTaskService.findByJobId(jobEntity.getId())
                     .ifPresent(task -> {
-                        broker.enqueueEvent(event.getStatus() == JobStatus.DONE ? eventBuilder.ofSucceededTask(task)
+                        TaskStatus status =
+                                "DLM".equalsIgnoreCase(jobEntity.getJobType())
```
LioRoger, line 67: Is there any consistent definition of jobType, string constant is not easy typo

server/odc-service/src/main/java/com/oceanbase/odc/service/task/listener/JobTerminateNotifyListener.java
```diff
@@ -60,7 +63,11 @@ public void onEvent(JobTerminateEvent event) {
...
+                                        ? dlmService.getFinalTaskStatus(task.getId())
+                                        : jobEntity.getStatus().convertTaskStatus();
+                        broker.enqueueEvent(status == TaskStatus.DONE ? eventBuilder.ofSucceededTask(task)
```
LioRoger, line 70: is cancel a candidate state?

server/odc-service/src/main/java/com/oceanbase/odc/service/task/listener/JobTerminateNotifyListener.java
```diff
@@ -60,7 +63,11 @@ public void onEvent(JobTerminateEvent event) {
             JobEntity jobEntity = taskFrameworkService.find(event.getJi().getId());
             scheduleTaskService.findByJobId(jobEntity.getId())
                     .ifPresent(task -> {
-                        broker.enqueueEvent(event.getStatus() == JobStatus.DONE ? eventBuilder.ofSucceededTask(task)
+                        TaskStatus status =
+                                "DLM".equalsIgnoreCase(jobEntity.getJobType())
+                                        ? dlmService.getFinalTaskStatus(task.getId())
```
LioRoger, line 68: why dml maintains it's own status and may not same with job status?

PR #3602, opened 2024-09-29
feat(osc):  check ghost table not exist before dispatch flow task

LioRoger:
#### What type of PR is this?

type-feature
1. check ghost table not exist before dispatch flow task to guarantee ghost table must be created by osc task
2. drop ghost table when clean resources cause ghost table must be created by osc task
3. fix clean resource not handle exception correctly when release oms resource

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/OscTableUtil.java
```diff
@@ -0,0 +1,49 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * util to operate osc table
+ * 
+ * @author longpeng.zlp
+ * @date 2024/9/27 17:01
+ */
+@Slf4j
+public class OscTableUtil {
```
LuckyPickleZZ, line 36: I think this should be a private method instead of an util class

yizhouxw (approved):
lgtm

PR #3605, opened 2024-09-29
feat(osc): check ghost table not exist before dispatch flow task

LioRoger:
#### What type of PR is this?

type-feature
1. check ghost table not exist before dispatch flow task to guarantee ghost table must be created by osc task
2. drop ghost table when clean resources cause ghost table must be created by osc task
3. fix clean resource not handle exception correctly when release oms resource

Fixes #

```docs

```


LuckyPickleZZ (approved):
lgtm

yhilmare (approved):
lgtm

PR #3618, opened 2024-09-30
feat(i18n): change message of workspace and worksheet

CHLK:
#### What type of PR is this?
type-feature

change  i18n message of workspace and worksheet

Fixes #

```docs

```

### DIFF
--- a/server/odc-core/src/main/resources/i18n/BusinessMessages.properties
+++ b/server/odc-core/src/main/resources/i18n/BusinessMessages.properties
@@ -83,7 +83,7 @@ com.oceanbase.odc.ResourceType.ODC_NOTIFICATION_CHANNEL=Notification Channel
 com.oceanbase.odc.ResourceType.ODC_NOTIFICATION_POLICY=Notification Policy
 com.oceanbase.odc.ResourceType.ODC_NOTIFICATION_MESSAGE=Notification Message
 com.oceanbase.odc.ResourceType.ODC_GIT_REPOSITORY=Git Repository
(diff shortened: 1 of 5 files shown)


server/odc-core/src/main/resources/i18n/BusinessMessages.properties
```diff
@@ -83,7 +83,7 @@ com.oceanbase.odc.ResourceType.ODC_NOTIFICATION_CHANNEL=Notification Channel
 com.oceanbase.odc.ResourceType.ODC_NOTIFICATION_POLICY=Notification Policy
 com.oceanbase.odc.ResourceType.ODC_NOTIFICATION_MESSAGE=Notification Message
 com.oceanbase.odc.ResourceType.ODC_GIT_REPOSITORY=Git Repository
-com.oceanbase.odc.ResourceType.ODC_WORKSHEET=Worksheet
+com.oceanbase.odc.ResourceType.ODC_WORKSHEET=工作簿
```
yizhouxw, line 86: here should be 'Worksheet' the default file is designed for en_US language

yizhouxw (approved):

server/odc-core/src/main/resources/i18n/BusinessMessages.properties
CHLK: okay, i'll make this change.

yizhouxw (approved):

JessieWuJiexi (approved):

Issue #3616, opened 2024-09-30
[Bug]: the image built by odc did not decompress the obclient package


LuckyPickleZZ:
### ODC version

4.3.2

### OB version

-

### What happened?

the image built by odc did not decompress the obclient package.
There should be an obclient folder under /opt/odc.
![image](https://github.com/user-attachments/assets/9d9b857c-24db-4ee0-ac3f-07b1a100f994)

### What did you expect to happen?

-

### How can we reproduce it (as minimally and precisely as possible)?

build odc 4.3.2 docker image

PR #3625, opened 2024-10-08
fix(build): obclient installation is missing from the ODC image

LuckyPickleZZ:
#### What type of PR is this?
type-bug

There is an error in the bash script in the `server/odc-server/pom.xml`  file. This will result in the docker image built without installation of obclient

Fixes https://github.com/oceanbase/odc/issues/3616

```docs

```


yhilmare (approved):
lgtm

PR #3636, opened 2024-10-09
fix(load-data): add instanceId and tenantId into OBInstance

LuckyPickleZZ:
add region and cloudProvider into OBInstance.


yizhouxw (approved):
lgtm

PR #3640, opened 2024-10-10
fix(database-change): get object input stream error

kiko-art:
#### What type of PR is this?
type-bug

Return the input stream before closing it, and ensure to close the stream after its usage is complete.
Return the input stream before closing it, and ensure to close the stream after its usage is complete.

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/SqlPlanTask.java
```diff
@@ -190,8 +190,9 @@ private void initSqlIterator() {
         }
 
         for (String sqlObjectId : parameters.getSqlObjectIds()) {
+            InputStream current = null;
             try {
-                InputStream current = cloudObjectStorageService.getObject(sqlObjectId);
```
yizhouxw, line 194: user try .. resource for simplify code

guowl3 (approved):
lgtm

yizhouxw (approved):
lgtm

PR #3643, opened 2024-10-10
feat(authentication): add currentProjectId in AuthenticationFacade

CHLK:
#### What type of PR is this?
type-feature

add currentProjectId in AuthenticationFacade

Fixes #

```docs

```

### DIFF
--- a/server/starters/web-starter/src/main/java/com/oceanbase/odc/service/iam/auth/WebAuthenticationFacade.java
+++ b/server/starters/web-starter/src/main/java/com/oceanbase/odc/service/iam/auth/WebAuthenticationFacade.java
@@ -15,9 +15,10 @@
  */
 package com.oceanbase.odc.service.iam.auth;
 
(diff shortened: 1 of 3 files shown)


server/starters/web-starter/src/main/java/com/oceanbase/odc/service/iam/auth/WebAuthenticationFacade.java
```diff
@@ -24,5 +25,6 @@
  */
 @Profile("alipay")
 @Service
+@ConditionalOnMissingBean(AuthenticationFacade.class)
```
yhilmare, line 28: we may use some properties to control this bean's init

CHLK: okay, i'll  make this change

yhilmare (approved):
lgtm

PeachThinking (approved):
lgtm

PR #3644, opened 2024-10-10
fix(schedule): optimize the logic for creating scheduled tasks

guowl3:
#### What type of PR is this?
type-bug

1. Refresh the status of task units when restarting a task
2. Optimize the logic for creating scheduled tasks

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/job/DataArchiveDeleteJob.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/job/DataArchiveDeleteJob.java
@@ -15,15 +15,14 @@
  */
 package com.oceanbase.odc.service.schedule.job;
 
-import java.util.Optional;
-
 import org.quartz.JobExecutionContext;
 
-import com.oceanbase.odc.common.json.JsonUtils;
 import com.oceanbase.odc.core.shared.constant.TaskStatus;
-import com.oceanbase.odc.metadb.schedule.ScheduleTaskEntity;
+import com.oceanbase.odc.core.shared.exception.NotFoundException;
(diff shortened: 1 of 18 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/model/ScheduleTaskType.java
```diff
@@ -38,6 +42,29 @@ public enum ScheduleTaskType {
 
     LOGICAL_DATABASE_CHANGE,
 
-    LOAD_DATA
+    LOAD_DATA;
+
+    public boolean isExecuteInTaskFramework() {
```
LioRoger, line 47: if we can avoid this switch case, cause when new job type added, we will always search the code, and find where the code should be add and which value should be returned. If code not added here, default value returned, and code may not provided any waring. 
Restraint code for user should be designed, whether annotation or interface, make problem found on compile or runtime period.

server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/job/DataArchiveDeleteJob.java
```diff
@@ -37,42 +36,39 @@
 public class DataArchiveDeleteJob extends AbstractDlmJob {
     @Override
     public void executeJob(JobExecutionContext context) {
+        Long scheduleId = ScheduleTaskUtils.getScheduleId(context);
+        Long scheduleTaskId = ScheduleTaskUtils.getScheduleTaskId(context);
```
LioRoger, line 40: Control logic can be detached from job, cause they are similar code.

LioRoger (commented):
any test added?

guowl3:
move to next version

PR #3645, opened 2024-10-10
fix(database-change): sql plan task execute failed

kiko-art:
#### What type of PR is this?
type-bug

1. add log printf
2. resolve the issue with sql plan execution failure

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/SqlPlanTask.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/SqlPlanTask.java
@@ -134,6 +134,7 @@ public class SqlPlanTask extends BaseTask<SqlPlanTaskResult> {
 
         while (sqlIterator.hasNext()) {
             String sql = sqlIterator.next().getStr();
+            log.info("Start execute sql, sql={}", sql);
             index++;
             // The retry statement will write the result into the buffer, while executing a new SQL command will
             // clear the buffer.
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/SqlPlanTask.java
```diff
@@ -426,11 +431,10 @@ private String uploadToOSS(String filePath) {
             try {
                 String objectName = cloudObjectStorageService.upload(file.getName(), file);
                 ossAddress = String.valueOf(cloudObjectStorageService.generateDownloadUrl(objectName));
-                log.info("upload sql plan task result file to OSS, file name={}", file.getName());
+                log.info("succeed to upload sql plan task result file to OSS, file download url = {}", ossAddress);
```
yizhouxw, line 434: seems syntax error, may 

`Upload sql plan task result to cloud object storage successfully, objectName={}`

and, downloadUrl contains ak/sk/sts, we may not output it in log content.

so as below log output.

kiko-art: Have been changed.

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/SqlPlanTask.java
```diff
@@ -162,12 +163,13 @@ protected boolean doStart(JobContext context) throws Exception {
...
         }
         result.setTotalStatements(index);
-        log.info("The sql plan task execute finished,result={}", result);
 
         // all sql execute csv file list write to zip file
         writeZipFile();
         // upload file to OSS, also contains error record where is non-null
         upload();
+
+        log.info("The sql plan task execute finished,result={}", result);
```
guowl3, line 172: only print the SQL that failed to execute

kiko-art: Have been changed.

yizhouxw (approved):
lgtm

guowl3 (approved):

PR #3647, opened 2024-10-11
feat(task):introduce errorMessage to TaskResult interface

LioRoger:
…ocessors

#### What type of PR is this?

type-feature
1.introduce errorMessage to TaskResult interface
2.refactor terminate processor to hide type related implement

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/DataArchiveTask.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/DataArchiveTask.java
@@ -69,7 +69,7 @@ public class DataArchiveTask extends BaseTask<List<DlmTableUnit>> {
     }
 
     @Override
-    protected boolean doStart(JobContext context) throws Exception {
+    protected boolean doStart(JobContext context, TaskContext taskContext) throws Exception {
 
         jobStore.setJobParameters(getJobParameters());
         DLMJobReq parameters =
@@ -84,6 +84,7 @@ public class DataArchiveTask extends BaseTask<List<DlmTableUnit>> {
             jobStore.setDlmTableUnits(result);
         } catch (Exception e) {
             log.warn("Get dlm job failed!", e);
+            taskContext.getExceptionListener().onException(e);
             return false;
         }
         Set<String> dlmTableUnitIds = result.keySet();
@@ -136,6 +137,7 @@ public class DataArchiveTask extends BaseTask<List<DlmTableUnit>> {
                 } else {
                     finishTableUnit(dlmTableUnitId, TaskStatus.FAILED);
                 }
+                taskContext.getExceptionListener().onException(e);
             }
         }
         return true;
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/SqlPlanTask.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/SqlPlanTask.java
@@ -116,7 +116,7 @@ public class SqlPlanTask extends BaseTask<SqlPlanTaskResult> {
     }
 
     @Override
-    protected boolean doStart(JobContext context) throws Exception {
+    protected boolean doStart(JobContext context, TaskContext taskContext) throws Exception {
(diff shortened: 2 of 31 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/DataArchiveTask.java
```diff
@@ -84,6 +84,7 @@ protected boolean doStart(JobContext context) throws Exception {
             jobStore.setDlmTableUnits(result);
         } catch (Exception e) {
             log.warn("Get dlm job failed!", e);
+            taskContext.getExceptionListener().onException(e);
```
yizhouxw, line 87: how about remove try... block, throw exception straightly,
or, it is hard for Task Implemation to follow task framework's exception handling flow.

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/Task.java
```diff
@@ -33,7 +33,7 @@ public interface Task<RESULT> {
     /**
      * Start current task. This method will be called by {@link TaskExecutor} for fire a task
      */
-    void start(JobContext context);
+    void start(JobContext context, TaskContext taskContext);
```
yizhouxw, line 36: two context make it more complex, can we keep one context only?

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/DataArchiveTask.java
LioRoger: It's  task inner implement.  base class not know how sub class will handle exception.
Actually, this change should not involved in this PR, cause how exception thrown and notify is the duty of task code owner.

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/Task.java
LioRoger: it has been correct as make jobContext as a member of taskContext

yizhouxw (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/DataArchiveTask.java
```diff
@@ -136,6 +137,7 @@ protected boolean doStart(JobContext context) throws Exception {
                 } else {
                     finishTableUnit(dlmTableUnitId, TaskStatus.FAILED);
                 }
+                taskContext.getExceptionListener().onException(e);
```
guowl3, line 140: alarm is unnecessary if the task is canceled.

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/SqlPlanTask.java
```diff
@@ -152,6 +152,7 @@ protected boolean doStart(JobContext context) throws Exception {
                     break;
                 }
                 log.warn("Sql task execution failed, will continue to execute next statement.", e);
+                taskContext.getExceptionListener().onException(e);
```
guowl3, line 155: When there are many SQL, if each one fails the alerts become overwhelming

LioRoger: only the latest exception will be saved, if exception is fetched by TaskResult, exception will be reseted. So various exception may be recorded if Task throw different exception one by one

guowl3: get it

guowl3 (dismissed):
lgtm

yizhouxw (dismissed):

guowl3 (approved):

yizhouxw (approved):
lgtm

ungreat (commented):
LGTM

ungreat (approved):

yhilmare (approved):
lgtm

PR #3650, opened 2024-10-11
refator(schedule): add schedule task full log download url to OrganizationAuthenticationInterceptor whiteList

ysjemmm:
#### What type of PR is this?
type-factor

skip the OrganizationAuthenticationInterceptor when request schedule task full log download


yhilmare (approved):
lgtm

PR #3652, opened 2024-10-11
fix(databaseChange): load file from oss failed

guowl3:
load file from oss failed because of cloud provider is null

Fixes #

```docs

```


yizhouxw (approved):
lgtm

PR #3653, opened 2024-10-11
feat(common): sensitive data utils support mask json

ungreat:

Fixes #

```docs

```

### DIFF
--- a/server/odc-common/src/main/java/com/oceanbase/odc/common/security/SensitiveDataUtils.java
+++ b/server/odc-common/src/main/java/com/oceanbase/odc/common/security/SensitiveDataUtils.java
@@ -15,9 +15,16 @@
  */
 package com.oceanbase.odc.common.security;
 
+import java.util.Iterator;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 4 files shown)


server/odc-common/src/main/java/com/oceanbase/odc/common/security/SensitiveDataUtils.java
```diff
@@ -15,18 +15,77 @@
  */
 package com.oceanbase.odc.common.security;
 
+import java.util.Iterator;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import org.json.JSONObject;
```
yizhouxw, line 22: is it defined in fastjson ?

ungreat: Not fastjson, use JsonUtils instead for uniformity. In interest I also looked for the author, from [maven central repository](https://central.sonatype.com/artifact/org.json/json) know that the known contributor is Douglas Crockford, known as the father of json.

yizhouxw (approved):
lgtm

PR #3664, opened 2024-10-12
fix(datasource): missing tenantName and instanceName in data source details

guowl3:

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/ConnectionConfig.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/ConnectionConfig.java
@@ -72,6 +72,9 @@ public class ConnectionConfig
     private static final String JDBC_URL_PARAMETERS_KEY = "JDBC_URL_PARAMETERS";
     private static final String CLOUD_PROVIDER = "cloudProvider";
     private static final String REGION = "region";
+    private static final String OB_INSTANCE_ID = "instanceId";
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/ConnectionConfig.java
```diff
@@ -550,6 +553,22 @@ public String getRegion() {
         return o == null ? null : o.toString();
     }
 
+    public String getInstanceId() {
```
yhilmare, line 556: what's the different between instanceId and clusterName? same as below

yhilmare (dismissed):

yhilmare (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/ConnectionConfig.java
```diff
@@ -72,6 +72,9 @@ public class ConnectionConfig
     private static final String JDBC_URL_PARAMETERS_KEY = "JDBC_URL_PARAMETERS";
     private static final String CLOUD_PROVIDER = "cloudProvider";
     private static final String REGION = "region";
+    private static final String OB_INSTANCE_NICK_NAME = "instanceNickName";
```
yhilmare, line 75: we may use these static name in odc-cloud, I notice that you use primitive value in odc-cloud

PR #3671, opened 2024-10-12
fix(dlm): the status is incorrect if creating the pod timeout

guowl3:
#### What type of PR is this?
update final status to failed if create pod timeout

fix the issue of wrong status if create pod timeout

Fixes #

```docs

```


kiko-art (approved):
lgtm

PR #3677, opened 2024-10-14
fix(dlm): custom sharding strategy being ineffective

guowl3:
#### What type of PR is this?
type-bug

Set the default sharding strategy only when the request parameter is nul

Fixes #

```docs

```


kiko-art (approved):
lgtm

yhilmare (approved):
lgtm

PR #3678, opened 2024-10-14
fix(osc): osc task result not update when oms step info changed

LioRoger:
#### What type of PR is this?

type-bug
when osc task entry oms monitor step and full transfer progress is zero,  task result will not update, cause progress variable not changed. 
compare oms step info to force update task result.

Fixes #

```docs

```


LuckyPickleZZ (approved):
LGTM

yizhouxw (approved):
lgtm

PR #3687, opened 2024-10-15
fix(pl): PL debug will fail when there are databases with the same name under the data source (one exists and one does not)

### LINKED ISSUE #3649: [Bug]: the project was not updated synchronously after the data source modified the environment
### ODC version

any

### OB version

unrelated

### What happened?

The database environment information in the project was not updated synchronously after the data source modified the environment

After the data source changes the environment, the database environment information in the project is updated synchronously.

This can be achieved by adjusting the data source environment and subsequently verifying whether the database environment within the associated project has been updated.

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug
PL Debug will fail when there are databases with the same name under the data source (one exists and one does not).

The root cause is that the query sql does not add `is_existed=true` in the where condition.

Fixes #3649

```docs

```


yhilmare (approved):
lgtm

PR #3689, opened 2024-10-15
fix(scheduleTask): correct status on task restart

guowl3:
#### What type of PR is this?
type-bug

The schedule task shows the wrong status if the task is restarted.

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #3693, opened 2024-10-15
fix(load-data): fix start load task without region

LuckyPickleZZ:
#### What type of PR is this?
type-bug

fix start ob-loader without region


yhilmare (approved):
lgtm

PR #3694, opened 2024-10-15
fix(rule): Executing SQLs occurs NPE when the database belongs to a customized environment after upgrading to 4.3.2

### LINKED ISSUE #3651: [Bug]:open sql console report NPE when datasource configured as custom environment
### ODC version

any

### OB version

unrelated

### What happened?

Open sql console report NPE when datasource configured as custom environment.

When the data source environment is a custom environment, the sql console can be opened normally.

1. In ODC 4.3.1, create a customized environment
2. Upgrade to ODC 4.3.2
3. Open the SQL window and execute SQL or view the customized environment's SQL rules.

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug

The root cause is that the customized ruleset does not involve the new rules after upgrading ODC versions.
The solution is:
1. All the customized rules will be disabled by default.
2. All the customized rules' default values will follow the value of the built-in default environment.
Additionally, as some users have already upgraded to ODC 4.3.2, I added another script to migrate the rules that added in ODC 4.3.2, and the script is repeatable, so it's safe for other users who want to upgrate from a lower version to 4.3.2-bp1 directly.

Fixes #3651

```docs

```


yizhouxw (approved):
lgtm

yhilmare (approved):
lgtm

PR #3697, opened 2024-10-16
fix(schedule): the creatorIds field in QueryScheduleParams is not valid

LuckyPickleZZ:
#### What type of PR is this?
type-bug

The `creatorIds` field in QueryScheduleParams is not valid.


guowl3 (approved):

yizhouxw (approved):
lgtm

PR #3706, opened 2024-10-17
fix(database-change): sql content delimiter invalid 

kiko-art:
#### What type of PR is this?
type-bug

fix sql content delimiter invalid

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/model/DatabaseChangeParameters.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/model/DatabaseChangeParameters.java
@@ -18,6 +18,7 @@ package com.oceanbase.odc.service.flow.task.model;
 import java.io.Serializable;
 import java.util.List;
 
+import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.NotNull;
 
 import com.oceanbase.odc.core.flow.model.TaskParameters;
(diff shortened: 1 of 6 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/model/DatabaseChangeParameters.java
```diff
@@ -42,7 +43,8 @@ public class DatabaseChangeParameters implements Serializable, TaskParameters {
     private List<String> rollbackSqlObjectIds;
     private Long timeoutMillis = 172800000L;// 2d for default
     private TaskErrorStrategy errorStrategy;
-    private String delimiter = ";";
+    @NotBlank
+    private String delimiter;
```
yizhouxw, line 47: delimiter parameter should has default value, 
if not set, then use `;` instead

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/task/SqlPlanTask.java
```diff
@@ -120,6 +120,7 @@ protected void doInit(JobContext context) {
         this.result.setRegion(JobPropertiesUtils.getRegionName(jobProperties));
         this.result.setCloudProvider(JobPropertiesUtils.getCloudProvider(jobProperties));
         this.connectionSession = generateSession();
+        ConnectionSessionUtil.getSqlCommentProcessor(connectionSession).setDelimiter(parameters.getDelimiter());
```
yizhouxw, line 123: should set delimiter while construct CommentProcessor in `generateSession` method.

PR #3708, opened 2024-10-17
fix(database-change): sql content delimiter invalid

kiko-art:
#### What type of PR is this?
type-bug

fix sql content delimiter invalid

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #3709, opened 2024-10-17
fix(schedule): fail to filter by tenantId and clusterId

guowl3:

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduleService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduleService.java
@@ -731,9 +731,12 @@ public class ScheduleService {
             @NotNull QueryScheduleParams params) {
         log.info("List schedule overview req:{}", params);
         if (StringUtils.isNotBlank(params.getCreator())) {
-            params.setCreatorIds(userService.getUsersByFuzzyNameWithoutPermissionCheck(
-                    params.getCreator()).stream().map(User::getId).collect(Collectors.toSet()));
-            return Page.empty();
+            Set<Long> creatorIds = userService.getUsersByFuzzyNameWithoutPermissionCheck(
+                    params.getCreator()).stream().map(User::getId).collect(Collectors.toSet());
+            if (creatorIds.isEmpty()) {
+                return Page.empty();
+            }
+            params.setCreatorIds(creatorIds);
         }
         if (params.getDataSourceIds() == null) {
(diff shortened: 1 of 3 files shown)


yizhouxw (approved):
lgtm

kiko-art (dismissed):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduleService.java
```diff
@@ -789,8 +805,16 @@ public Page<ScheduleTaskListOverview> listScheduleTaskOverviewByScheduleType(@No
...
-                    authenticationFacade.currentOrganizationId(), params.getTenantId()));
+            List<Long> datasourceIds = connectionService.innerListIdByOrganizationIdAndTenantId(
+                    authenticationFacade.currentOrganizationId(), params.getTenantId());
+            if (datasourceIds.isEmpty()) {
+                return Page.empty();
+            }
+            if (params.getDataSourceIds().isEmpty()) {
+                params.getDataSourceIds().addAll(datasourceIds);
+            } else {
+                params.getDataSourceIds().retainAll(datasourceIds);
```
yhilmare, line 816: why should we retain all?
the familiar logic above does not do this

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/server/TaskMonitor.java
```diff
@@ -181,6 +181,7 @@ private void doFinal() {
         uploadLogFileToCloudStorage(finalResult);
 
         Map<String, String> latestLogMeta = finalResult.getLogMetadata();
+        log.info("Task id: {}, latest log meta: {}", getJobId(), latestLogMeta);
```
yhilmare, line 184: will it print a lot of log?

guowl3: just print it when the task is finished

server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduleService.java
guowl3: when the filter criteria include both tenantId and clusterId, the union of the two needs to be taken.

kiko-art (approved):
lgtm

yhilmare (approved):
lgtm

PR #3710, opened 2024-10-17
fix(task-framework): update task result before canceling

LuckyPickleZZ:
#### What type of PR is this?
type-bug

If the task is immediately stopped, its resultJson field will not be updated. So before changing the task status, update the task results first.


server/odc-service/src/main/java/com/oceanbase/odc/service/task/service/StdTaskFrameworkService.java
```diff
@@ -368,6 +368,13 @@ public boolean refreshLogMeta(Long id) {
         try {
             String executorEndpoint = executorEndpointManager.getExecutorEndpoint(je);
             DefaultTaskResult result = taskExecutorClient.getResult(executorEndpoint, JobIdentity.of(id));
+
+            // If the task is immediately stopped, its resultJson field would NOT be updated. So before changing
+            // the task status, update the task results first.
+            if (result != null) {
+                updateTaskResult(result, je);
```
guowl3, line 375: update result after logmeta saved

LuckyPickleZZ: done

guowl3 (approved):
lgtm

yhilmare (approved):
lgtm

PR #3712, opened 2024-10-18
fix(datasource): no return DSS datasource id

guowl3:
#### What type of PR is this?
type-bug

return dss datasource id when detail datasource

Fixes #

```docs

```


yizhouxw (approved):
lgtm

PR #3715, opened 2024-10-18
fix(database-chabge): allow delete finished schedule

kiko-art:
#### What type of PR is this?
type-bug

allow delete finished schedule

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduleService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduleService.java
@@ -303,8 +303,8 @@ public class ScheduleService {
                 throw new IllegalStateException("Update schedule is not allowed.");
             }
             if (req.getOperationType() == OperationType.DELETE
-                    && (targetSchedule.getStatus() != ScheduleStatus.TERMINATED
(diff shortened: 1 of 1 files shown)


yhilmare (dismissed):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduleService.java
```diff
@@ -303,9 +303,11 @@ public ChangeScheduleResp changeSchedule(ScheduleChangeParams req) {
                 throw new IllegalStateException("Update schedule is not allowed.");
             }
             if (req.getOperationType() == OperationType.DELETE
-                    && targetSchedule.getStatus() != ScheduleStatus.TERMINATED) {
+                    && (targetSchedule.getStatus() != ScheduleStatus.TERMINATED
+                            || targetSchedule.getStatus() != ScheduleStatus.COMPLETED)) {
```
yhilmare, line 307: the judgement here is not true

kiko-art: Already changed.

yhilmare (approved):
lgtm

PR #3717, opened 2024-10-18
fix(datasource): npe will be thrown when calling get connection concurrently

yhilmare:
#### What type of PR is this?

type-bug

npe will be thrown when calling get connection concurrently

Fixes #

```docs

```


yizhouxw (approved):
lgtm

PR #3720, opened 2024-10-19
fix(database-change): condition search invalid

kiko-art:
#### What type of PR is this?
type-bug

fix condition search invalid

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/model/QueryScheduleTaskParams.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/model/QueryScheduleTaskParams.java
@@ -32,7 +32,7 @@ public class QueryScheduleTaskParams {
     private List<Schedule> schedules;
     private String scheduleName;
     private Set<Long> dataSourceIds;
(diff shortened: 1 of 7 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/model/QueryScheduleTaskParams.java
```diff
@@ -32,7 +32,7 @@ public class QueryScheduleTaskParams {
     private List<Schedule> schedules;
     private String scheduleName;
     private Set<Long> dataSourceIds;
-    private List<TaskStatus> statuses;
+    private TaskStatus status;
```
yizhouxw, line 35: status in can cover status equals scenario, may not neccessary to change it ?

kiko-art: Changed back

yizhouxw (approved):
lgtm

yhilmare (approved):
lgtm

PR #3726, opened 2024-10-21
fix(taskframework): update load data task result only

LuckyPickleZZ:
fix https://github.com/oceanbase/odc/pull/3721
only update load data task result, To avoid possible state errors


yizhouxw (approved):
lgtm

yhilmare (approved):
lgtm

PR #3731, opened 2024-10-22
fix(global-search): failed to sync database which get 'PENDING' status and 'object_last_sync_time' is null

### LINKED ISSUE #3730: [Bug]: failed to sync database's schema when there are too much databases
### ODC version

4.3.2

### OB version

not related

### What happened?

1. sync database normally and the `object_sync_status` is set to 'PENDING' successfully (btw, the column `object_last_sync_time` is null for the first time sync)
截屏2024-10-22 15 02 34

2. shutdown odc and make sure that there are several databases in table `connect_database` still in 'PENDING' status and `object_last_sync_time` is null.

3. restart the odc and you will find that the database which get 'PENDING' status and `object_last_sync_time` will be never synced when you click 'refresh' button, the loading icon will never disappear:

截屏2024-10-22 15 07 28

there should not be any database never be synced~

follow 'what happened'

_No response_

yhilmare:
#### What type of PR is this?
type-bug

any exception happened is database syncing will lead to 'PENDING' status and `object_last_sync_time` is null

database will never be syncing when such situation happened, this pr fix this problem.

Fixes #3730

```docs

```


yizhouxw (approved):
lgtm

MarkPotato777 (approved):
LGTM

PeachThinking (approved):
lgtm

PR #3737, opened 2024-10-24
fix(query-profie): NPE occurred when `filter`'s value is null

LuckyPickleZZ:
#### What type of PR is this?
type-bug

When query from `gv$ob_sql_plan` view, the `fiter` value could be null. But we missed this, so there may be a NPE for some queries.


yizhouxw (approved):
lgtm

PR #3741, opened 2024-10-25
security:  follow a authenticated schedule id allows you to perform operations on any other schedule tasks

ysjemmm:
#### What type of PR is this?
security-bug

Because the server only authenticates and verifies the `scheduleId`, but does not verify whether a `scheduleTaskId` belongs to the `scheduleId`, the problem that arises is that if I have a verified `scheduleId`, I can access any other scheduleTask resource.
There is some code in the logging module that only fits `obcloud_202409`, so it is included this time

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduledTaskLoggerService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduledTaskLoggerService.java
@@ -156,7 +156,7 @@ public class ScheduledTaskLoggerService {
             return String.format(DOWNLOAD_LOG_URL_PATTERN, scheduleId, scheduleTaskId);
         } catch (Exception e) {
             log.warn("get download log url failed, scheduleId={}, scheduleTaskId={}", scheduleId, scheduleTaskId, e);
-            return StrUtil.EMPTY;
+            throw new RuntimeException("get download log url failed, scheduleId=" + scheduleId, e);
         }
     }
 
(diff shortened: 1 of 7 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/task/service/StdTaskFrameworkService.java
```diff
@@ -662,4 +663,15 @@ private void handleTaskResult(String jobType, TaskResult taskResult) {
         }
     }
 
+    @Override
+    public Map<String, String> getJobAttributes(Long jobId) {
+        List<JobAttributeEntity> attributeEntityList = jobAttributeRepository.findByJobId(jobId);
+        if (attributeEntityList == null) {
```
yizhouxw, line 669: attributeEntityList will never be null

server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduledTaskLoggerService.java
```diff
@@ -95,68 +103,74 @@ public class ScheduledTaskLoggerService {
...
+                String objId = jobAttributeMap.get(attributeKey);
+                String bucketName = jobAttributeMap.get(JobAttributeKeyConstants.LOG_STORAGE_BUCKET_NAME);
+                if (ObjectUtil.isNotNull(objId) && ObjectUtil.isNotNull(bucketName)) {
+                    return cloudObjectStorageService.generateDownloadUrl(objId).toString();
                 }
+                return StrUtil.EMPTY;
             }
+            return String.format(DOWNLOAD_LOG_URL_PATTERN, scheduleId, scheduleTaskId);
+        } catch (Exception e) {
+            log.warn("get download log url failed, scheduleId={}, scheduleTaskId={}", scheduleId, scheduleTaskId, e);
```
yizhouxw, line 158: throw exception instead

server/odc-service/src/main/java/com/oceanbase/odc/service/task/service/StdTaskFrameworkService.java
ysjemmm: I know

yhilmare (approved):
lgtm

PR #3747, opened 2024-10-28
fix(osc): remove ip mask when match user and session

LioRoger:
#### What type of PR is this?

type-bug
if user in user@'ip' format, list user and list session will not match any user, that will ignore lock user and kill session operation which leads to data loss
Notice: this pr has a amplify impact for kill user range. eg if user choose lock user1@ip1, and it has use1@ip2 account as well.   user@ip1 and user @ip2 will both be locked and killed

Fixes #

```docs

```


LuckyPickleZZ (approved):
lgtm

yhilmare (approved):
lgtm

PeachThinking (approved):
lgtm

yizhouxw (approved):
lgtm

PR #3766, opened 2024-10-31
feat(task): refactor task getStatus interface return TaskStatus instead of JobStatus

LioRoger:
#### What type of PR is this?

type-feature

let task get status return TaskStatus, That means task will not known nothing about job. job status will be refreshed depends on task status

Fixes #

```docs

```


guowl3 (approved):
lgtm

yhilmare (approved):
lgtm

PR #3767, opened 2024-10-31
feat(schedule): check the validity of the task each time it is scheduled

guowl3:
#### What type of PR is this?
type-feature

Check  the  validity  of  the  schedule  and  terminate  any  invalid  ones,  such  as:
1. The status of the schedule is not enabled.
2. The database is invalid or does not belong to the current project.
3. The project is archived.

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduleService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduleService.java
@@ -70,6 +70,7 @@ import com.oceanbase.odc.service.collaboration.project.ProjectService;
 import com.oceanbase.odc.service.common.util.SpringContextUtil;
 import com.oceanbase.odc.service.connection.ConnectionService;
 import com.oceanbase.odc.service.connection.database.DatabaseService;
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduleService.java
```diff
@@ -500,31 +500,39 @@ public void terminate(ScheduleEntity scheduleConfig) throws SchedulerException {
...
-                    terminate(entity);
-                } catch (Exception e) {
-                    log.warn("Terminate schedule failed,scheduleId={}", scheduleId);
-                }
+        Schedule schedule = null;
+        boolean isValidSchedule;
+        try {
+            schedule = nullSafeGetModelById(scheduleId);
+            isValidSchedule = isValidSchedule(schedule);
+        } catch (NotFoundException e) {
```
yhilmare, line 522: what about other exception?

guowl3: task will be executed if catch other exception

PR #3778, opened 2024-11-01
fix(data-transfer): read timout occurred when getting large for internal usage

### LINKED ISSUE #3759: [Bug]: the internal file transfer timeout is too short
### ODC version

4.2.4

### OB version

not related

### What happened?

1. upload a big file(more than 200MB) and startup a import task
2. use timer executor
截屏2024-10-30 11 08 50

3. make sure that the machine execute this task and the file to be uploaded is different
4. make sure that such exception is thrown
截屏2024-10-30 11 10 15
5. the user will find out that he/she will import nothing
lQLPJxIGn7mAduXNAhzNA3ewJPXrOoNpx3sHCHVwJJuWAA_887_540

import csv normally

follow what happened

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

When getting internal file for datatransfer, the default timeout was 10s . This is to small. ODC allows importing up to 2 GB of files.  So We estimate a 5-minute timeout based on transmission rate of 10 MB/s.

Fixes #3759

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/DataTransferRuntimeFlowableTask.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/DataTransferRuntimeFlowableTask.java
@@ -49,7 +49,7 @@ public class DataTransferRuntimeFlowableTask extends BaseODCFlowTaskDelegate<Voi
      * ODC allows importing up to 2 GB of files. We estimate a 5-minute timeout based on transmission
      * rate of 10 MB/s.
      */
-    private static final int DEFAULT_GET_IMPORT_FILE_TIMEOUT_SECONDS = 300;
+    private static final int DEFAULT_GET_IMPORT_FILE_TIMEOUT_MILLIS = 300 * 1000;
 
     @Autowired
     private DataTransferService dataTransferService;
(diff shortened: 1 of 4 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/flow/task/DataTransferRuntimeFlowableTask.java
```diff
@@ -45,6 +45,11 @@
 
 @Slf4j
 public class DataTransferRuntimeFlowableTask extends BaseODCFlowTaskDelegate<Void> {
+    /**
+     * ODC allows importing up to 2 GB of files. We estimate a 5-minute timeout based on transmission
+     * rate of 10 MB/s.
+     */
+    private static final int DEFAULT_GET_IMPORT_FILE_TIMEOUT_MILLIS = 300 * 1000;
```
yhilmare, line 52: shall we make this parameter editable?

LuckyPickleZZ: Under what circumstances do we need to modify this value?

yhilmare (approved):
lgtm

PR #3779, opened 2024-11-01
security(notification): filter illegal characters in notification webhook

LuckyPickleZZ:
#### What type of PR is this?
type-security

User would use some illegal characters to invoke SSRF attack. We should filter these characters.


yhilmare (approved):
lgtm

yizhouxw (approved):
lgtm

MarkPotato777 (approved):

PR #3780, opened 2024-11-01
fix(osc):  mdc context may not correct

LioRoger:
#### What type of PR is this?

type-bug
1.mdc is static variable and may be set by other thread. MDC not enduring. so different ODC node have different value maybe. resume and switch osc operation will lead to unknown result of MDC. use information stored to build MDC info
2. fix session leak
3. status change will always trigger result update  for task result column

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/oscfms/OscActionFsmBase.java
```diff
@@ -411,7 +414,7 @@ public ConnectionSession createConnectionSession() {
                     new DefaultConnectSessionFactory(connectionConfig).generateSession();
             ConnectionSessionUtil.setCurrentSchema(connectionSession,
                     dbName);
-            return new DefaultConnectSessionFactory(connectionConfig).generateSession();
+            return connectionSession;
```
yhilmare, line 417: remember that the connection session should be expired when you don't use it anymore otherwise It will be lead to resource leak

LuckyPickleZZ (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/oscfms/OscActionFsmBase.java
LioRoger: this is a wrapper, it will be called where the connection is needed and release connection after use

PR #3781, opened 2024-11-01
secure(framework): enable secure cookie for http session

yizhouxw:
#### What type of PR is this?

type-security

use secure cookie setting 'secure' and 'http-only' for http session


server/odc-service/src/main/java/com/oceanbase/odc/config/BeanConfiguration.java
```diff
@@ -111,14 +108,6 @@ public CloudMetadataClient cloudMetadataClient() {
         return new NullCloudMetadataClient();
     }
 
-    @Bean
-    @Primary
-    public CookieSerializer cookieSerializer() {
```
yizhouxw, line 116: if use selfdefined CookieSerializer, spring boot configuration will not works

yhilmare (approved):
lgtm

PR #3785, opened 2024-11-05
feat(object-storage): use s3 sdk for google cloud storage

LuckyPickleZZ:
#### What type of PR is this?
type-feature

This PR uses the S3 SDK to support for Google Cloud Storage (GCS).
Most of the object storage operations currently supported by ODC are compatible with the S3 protocol in GCS.

However, please note that batch deletion in GCS is not compatible with the S3 SDK. Therefore, a separate `deleteObject` method has been added for deleting objects.


yizhouxw (approved):
lgtm

yhilmare (approved):
lgtm

PR #3786, opened 2024-11-05
security(object-storage): add user id into object key

LuckyPickleZZ:
#### What type of PR is this?
type-security

When generating and using object key , we didn't validate whether the user id is consistent with the object . So in some scenarios, attackers may be able to directly access files after obtaining the object keys of certain users.


yizhouxw (approved):
lgtm

PR #3798, opened 2024-11-08
feat(web): session cookie secure default false

yizhouxw:
change default setting for cookie secure from true to false,
for avoid http depolyment not works by default.


yhilmare (approved):
lgtm

PR #3805, opened 2024-11-08
fix(partition-plan): task's status is not as expected when error strategy is set to 'ignore any errors'

### LINKED ISSUE #3784: [Bug]: the status for database change task is incorrect
### ODC version

4.3.x

### OB version

-

### What happened?

When executin database change task and configuring "continue on error". The task's status would ALWAYS be successful.
This will lead to user misunderstanding. 

We should update the status into failed when there are failed sqls.

⬆️

_No response_

yhilmare:
#### What type of PR is this?

type-bug

the task status will be set to 'sucess' even there are several errors in sql execution when we set 'ignore any errors' as the error process strategy which is not as expected.

Fixes #3784

```docs

```


yizhouxw (approved):
lgtm

guowl3 (approved):
lgtm

PR #3807, opened 2024-11-08
feat(task): refactor task interface, introduce task runtime component

LioRoger:
#### What type of PR is this?

type-feature
1. remove BaseTask, introduce TaskBase only contains jobContext variables
2. introduce TaskContainer to hold task runtime info. TaskStatus will not known by task anymore.
3. move cloudStorageService to TaskContext as a candidate component
4. move TaskMonitor/DefaultTaskBuilder eg to agent package

Fixes #

```docs

```


LuckyPickleZZ (approved):

server/odc-server/pom.xml
```diff
@@ -148,6 +148,11 @@
             <artifactId>mockito-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
```
yhilmare, line 152: why should we need this?

guowl3 (dismissed):
lgtm

server/odc-server/pom.xml
LioRoger: mock static and MockVerify need this package

yhilmare (dismissed):
lgtm

yizhouxw (dismissed):
lgtm

MarkPotato777 (commented):
The refactor looks good to me. I have another requirement but it's not in emergency: Can we have a mechanism to operate on sub-tasks? Currently, the logical database change tasks need to perform operations such as termination and skipping on a specific sub-task. Currently, we are achieving this through `modifyJobParameters` by passing sub-task ID and different parameters like `terminate` and `skip`, which is somewhat tricky.

LioRoger:
> The refactor looks good to me. I have another requirement but it's not in emergency: Can we have a mechanism to operate on sub-tasks? Currently, the logical database change tasks need to perform operations such as termination and skipping on a specific sub-task. Currently, we are achieving this through `modifyJobParameters` by passing sub-task ID and different parameters like `terminate` and `skip`, which is somewhat tricky.

that depends on Task design model. currently Task model hopes developer do all job in start method.  no sub task concept is
introduces.  
your requirement is more like a high level concept design.  Imagine if there is a object contains db change tasks, each task operate a instance,  operator delivered by this object to correspond task. but sadly this design is not so urgent.

yhilmare (approved):
lgtm

PR #3808, opened 2024-11-08
feat(dashboard): support dashboard

CHLK:
#### What type of PR is this?
type-feature

- feat(resourceLastAccess): add batchDeleteResourceIds to ResourceLastAccessService
- feat(ResourceType): add Chart and Dashboard to ResourceType

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/resourcehistory/ResourceLastAccessService.java
```diff
@@ -68,6 +69,18 @@ public ResourceLastAccessEntity add(Long organizationId, Long projectId, Long us
         return entity;
     }
 
+    public int batchDeleteResourceIds(Long organizationId, Long projectId, ResourceType resourceType,
```
yhilmare, line 72: `@NonNull` annotation may be better?
why should we judge null if we have done it at the `server/odc-service/src/main/java/com/oceanbase/odc/metadb/resourcehistory/ResourceLastAccessRepository.java`

server/odc-core/src/main/java/com/oceanbase/odc/core/shared/constant/ResourceType.java
```diff
@@ -143,4 +145,13 @@ public String getLocalizedMessage() {
         Locale locale = LocaleContextHolder.getLocale();
         return translate(null, locale);
     }
+
+    public static Optional<ResourceType> getByName(String name) {
```
yhilmare, line 149: what's the difference between this and `ResourceType.valueOf()`?

yhilmare (approved):
lgtm

yhilmare (approved):
lgtm

server/odc-core/src/main/resources/i18n/BusinessMessages.properties
```diff
@@ -86,6 +86,9 @@ com.oceanbase.odc.ResourceType.ODC_GIT_REPOSITORY=Git Repository
 com.oceanbase.odc.ResourceType.ODC_WORKSHEET=Worksheet
 com.oceanbase.odc.ResourceType.ODC_WORKSPACE=Workspace
 com.oceanbase.odc.ResourceType.ODC_WORKFLOW_CLUSTER=Workflow Cluster
```
JessieWuJiexi, line 88: 这个数仓开发术语讨论的时候定的是 Compute Node Pool

CHLK: okay,  I'll make this change.

JessieWuJiexi (approved):

yhilmare (approved):
lgtm

PR #3815, opened 2024-11-11
fix(schedule): customize description generator according to different environments

guowl3:
#### What type of PR is this?
type-bug

Customize description generator according to different environments

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #3819, opened 2024-11-11
fix(dlm): do not print logs if the data source is uninitialized

guowl3:

Fixes #

```docs

```


kiko-art (approved):
lgtm

PR #3820, opened 2024-11-11
feat(permission): support for global project roles

MarkPotato777:
#### What type of PR is this?

type-feature
This PR supports global project roles: `global_project_dba`, `global_project_owner`, `global_project_security_administrator`.
- This PR added a new implementation of `Permission` which is `ComposePermission`.  The `ComposePermission` is designed to fulfill the requirement where either `ResourceRolePermission` or `ResourcePermission` passing will result in a pass. 
- In addition, it added a new subclass of ResourcePermission which is `ProjectPermission` as the `ProjectPermission` is quite different from the ResourcePermission. `ProjectPermission` does not need a mask because all the actions are independent, and the method `implies`'s implementation is also different which will return true as long as there is an intersection between the two action lists.

Fixes #

```docs

```


server/odc-core/src/main/java/com/oceanbase/odc/core/authority/permission/ProjectPermission.java
```diff
@@ -0,0 +1,77 @@
...
+    @Override
+    public boolean implies(Permission permission) {
+        if (!(permission instanceof ProjectPermission)) {
+            return false;
+        }
+        ProjectPermission that = (ProjectPermission) permission;
+        if (this == that) {
+            return true;
+        }
+        boolean returnVal = (this.resourceId.equals(that.resourceId) || "*".equals(this.resourceId));
```
LioRoger, line 73: boolean expression with and logic should write inline, it will return when meet first false value.
or you can use if else branch  if you want save some compute resource

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/auth/BaseAuthorizer.java
```diff
@@ -43,6 +44,27 @@ public void checkPermission(Principal principal, Collection<Permission> permissi
...
 
+    @Override
+    public boolean isPermitted(Principal principal, Collection<Permission> permissions, SecurityContext context) {
+        User odcUser = (User) principal;
+        if (Objects.isNull(odcUser.getId())) {
+            return false;
+        }
+        for (Permission permission : permissions) {
+            boolean accessDenied = true;
+            for (Permission resourcePermission : listPermittedPermissions(principal)) {
```
LioRoger, line 55: will listPermittedPermissions access remote storage? 
if so can reuse the result first queried?
Is this logic want's to describe a rule that any resource permission not permitted by  listPermittedPermissions  will return false?

server/odc-migrate/src/main/resources/migrate/common/V_4_3_3_3__iam_role_permission.yaml
```diff
@@ -0,0 +1,66 @@
+kind: resource
```
yhilmare, line 1: will we release this on public aliyun? if so, there may be some transaction problems

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/auth/BaseAuthorizer.java
```diff
@@ -55,4 +77,6 @@ public void checkPermission(Principal principal, Collection<Permission> permissi
     public boolean supports(Class<? extends Principal> principal) {
         return User.class.isAssignableFrom(principal);
     }
+
+    public abstract List<Permission> listPermittedPermissions(Principal principal);
```
yhilmare, line 81: make it protected may be better

server/odc-core/src/main/java/com/oceanbase/odc/core/authority/permission/ComposedPermission.java
```diff
@@ -0,0 +1,59 @@
...
+    @Override
+    public boolean implies(Permission permission) {
+        if (!(permission instanceof ComposedPermission)) {
+            return false;
+        }
+        ComposedPermission composedPermission = (ComposedPermission) permission;
+        if (CollectionUtils.isEmpty(composedPermission.getPermissions())) {
+            return true;
+        }
+        if (CollectionUtils.isEmpty(this.permissions)) {
```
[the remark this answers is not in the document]
MarkPotato777: No, it `A.implies(B)` doesn't mean `B.implies(A)`, which is expected and by design.

server/odc-core/src/main/java/com/oceanbase/odc/core/authority/permission/ProjectPermission.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+
+    @Override
+    public boolean implies(Permission permission) {
+        if (!(permission instanceof ProjectPermission)) {
+            return false;
+        }
+        ProjectPermission that = (ProjectPermission) permission;
+        if (this == that) {
```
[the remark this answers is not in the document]
MarkPotato777: done.

server/odc-core/src/main/java/com/oceanbase/odc/core/authority/permission/ProjectPermission.java
MarkPotato777: done.

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/auth/BaseAuthorizer.java
MarkPotato777: > will listPermittedPermissions access remote storage?
> if so can reuse the result first queried?

Yes, it will access to the metadb. And you are right, if should be reused, and I updated it.

> Is this logic want's to describe a rule that any resource permission not permitted by listPermittedPermissions will return false?

Yes, your understanding is right.

server/odc-core/src/main/java/com/oceanbase/odc/core/authority/permission/PermissionProvider.java
```diff
@@ -48,4 +48,13 @@ public interface PermissionProvider {
      */
     Permission getPermissionByResourceRoles(SecurityResource resource, Collection<String> resourceRoles);
 
+    /**
+     * @param resource {@link SecurityResource}
+     * @param actions action collection
+     * @param resourceRoles, see {@link ResourceRoleName} enums
+     * @return permission collection
+     */
+    Permission getPermissionByActionsAndResourceRoles(SecurityResource resource, Collection<String> actions,
```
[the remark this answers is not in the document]
MarkPotato777: I think it is simpler to separate them. I do not see any benefits in combining them; on the contrary, it would increase the implementation cost of the this combined method.

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/ProjectPermissionValidator.java
```diff
@@ -73,9 +76,15 @@ public boolean hasProjectRole(@NonNull Collection<Long> projectIds, @NonNull Lis
...
         List<Permission> permissions = projectIds.stream().filter(Objects::nonNull)
-                .map(projectId -> new ResourceRoleBasedPermission(
-                        new DefaultSecurityResource(projectId.toString(), "ODC_PROJECT"), roleNames))
-                .collect(Collectors.toList());
+                .map(projectId -> {
+                    Permission resourceRolePermission = new ResourceRoleBasedPermission(
+                            new DefaultSecurityResource(projectId.toString(), "ODC_PROJECT"), roleNames);
+                    Permission projectResourcePermission =
+                            new ProjectPermission(new DefaultSecurityResource(projectId.toString(), "ODC_PROJECT"),
+                                    roleNames.stream().map(ResourceRoleName::name).collect(
```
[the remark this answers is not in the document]
MarkPotato777: A "Global Project Role" is a role within an RBAC, where the resource is ODC_PROJECT:* and the action corresponds to the project role name.

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/auth/BaseAuthorizer.java
MarkPotato777: done

server/odc-core/src/main/java/com/oceanbase/odc/core/authority/permission/ProjectPermission.java
```diff
@@ -0,0 +1,77 @@
...
+import lombok.Getter;
+import lombok.NonNull;
+
+/**
+ * @Author: Lebie
+ * @Date: 2024/11/11 19:26
+ * @Description: []
+ */
+@Getter
+public class ProjectPermission extends ResourcePermission {
```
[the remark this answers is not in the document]
MarkPotato777: Because the `actions` of "Global Project Role" differ significantly from the existing `ResourcePermission` `actions`, it is necessary to reimplement the `implies` method. `ResourcePermission` `actions` have implicit relationships, whereas Global Project Role actions are specific project roles like `DBA`, `OWNER`, `DEVELOPER`, etc., with no implicit relationships between them.

server/odc-migrate/src/main/resources/migrate/common/V_4_3_3_3__iam_role_permission.yaml
MarkPotato777: Yes, it will. Any suggestion for the transaction problem?

LioRoger (approved):
lgtm

PR #3821, opened 2024-11-12
feat(external table): external table support sensitive column

zijiacj:

Fixes #

```docs

```


libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/DBSchemaAccessor.java
```diff
@@ -218,6 +218,16 @@ default List<String> showExternalTables(String schemaName) {
      */
     List<DBTableColumn> listBasicViewColumns(String schemaName, String viewName);
 
+    /**
+     * Get all external table columns(hold only basic info) in the specified schema
+     */
+    Map<String, List<DBTableColumn>> listBasicExternalTableColumns(String schemaName);
```
LuckyPickleZZ, line 224: I think `listExternalTableColumns` is enough.

server/odc-service/src/main/java/com/oceanbase/odc/service/datasecurity/SensitiveColumnService.java
```diff
@@ -145,6 +147,12 @@ public List<DatabaseWithAllColumns> listColumns(@NotNull Long projectId, @NotEmp
                         accessor.listBasicTableColumns(database.getName()), exists));
                 databaseColumn.setView2Columns(getFilteringExistColumns(database.getId(),
                         accessor.listBasicViewColumns(database.getName()), exists));
+                DialectType dialectType = session.getDialectType();
```
LuckyPickleZZ, line 150: Have you tested whether the external table can recognize sensitive columns properly?

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/DBSchemaAccessor.java
zijiacj: image
I recommend keeping the current method name the same as table and view. Only a small portion of the column object's field information is retrieved.
In fact, the implementation of the above three methods is to share the same sql statement, but now it is drawn into three methods and repeated writing three of the same sql, is it necessary to make an interface method?

server/odc-service/src/main/java/com/oceanbase/odc/service/datasecurity/SensitiveColumnService.java
zijiacj: image
image
yes

LuckyPickleZZ (dismissed):

yhilmare (approved):
lgtm

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/constant/StatementsFiles.java
```diff
@@ -35,6 +35,8 @@ public final class StatementsFiles {
 
     public static final String MYSQL_5_6_x = "schema/sql/mysql/mysql_5_6_x.yaml";
 
+    public static final String OBORACLE_4_3_2 = "schema/sql/oboracle/oboracle_4_3_2.yaml";
```
yhilmare, line 38: prefer using 4.3.x, same as obmysql ode

yhilmare (approved):
lgtm

PR #3824, opened 2024-11-12
fix(session): put killCurrentQuerySupported into session

ungreat:

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/DBSessionResp.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/DBSessionResp.java
@@ -15,6 +15,9 @@
  */
 package com.oceanbase.odc.service.connection.model;
 
+import org.mapstruct.Mapper;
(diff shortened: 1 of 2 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/DBSessionResp.java
```diff
@@ -33,7 +36,25 @@
 @NoArgsConstructor
 @AllArgsConstructor
 public class DBSessionResp {
-    private DBSession session;
+    private DBSessionRespDelegate session;
     private SessionSettings settings;
-    private boolean killCurrentQuerySupported;
+
+    @Mapper
+    interface DbSessionRespMapper {
```
yhilmare, line 43: DbSessionRespMapper -> DBSessionRespMapper

ungreat: done~

yhilmare (approved):
lgtm

PR #3827, opened 2024-11-12
feat(task): remove DefaultTaskResult

LioRoger:
#### What type of PR is this?

type-feature
1. remove default task result, cause task result is inner concept, no interface needed.
2. fix JobEntity.resultJson not set correct value leads to Json deserialize failure

Fixes #

```docs

```


server/odc-common/src/main/java/com/oceanbase/odc/common/util/MapUtils.java
```diff
@@ -128,4 +129,39 @@ public static int size(final Map<?, ?> map) {
         return org.apache.commons.collections4.MapUtils.size(map);
     }
 
+
+    private int getMapSize(Map<String, String> map) {
```
yizhouxw, line 133: already has method `public static int size(final Map<?, ?> map) `

server/odc-common/src/main/java/com/oceanbase/odc/common/util/MapUtils.java
```diff
@@ -128,4 +129,39 @@ public static int size(final Map<?, ?> map) {
...
+     * compare map, null map and empty map consider as equals
+     * 
+     * @param src
+     * @param target
+     * @param <Key> key of map
+     * @param <Value> value of map
+     * @param equalFunction function with argument value to compare if value instance is equals
+     * @return true is map is equals
+     */
+    public static <Key, Value> boolean mapEquals(Map<Key, Value> src, Map<Key, Value> target,
```
yizhouxw, line 147: may naming as `isEqual` , as currrent class name 'MapUtils' already has 'Map' concept

server/odc-common/src/test/java/com/oceanbase/odc/common/util/MapUtilsTest.java
```diff
@@ -97,4 +98,37 @@ public void formatKvString_ValueContainsSeparator_IllegalArgumentException() {
         map.put("a", "2,");
         MapUtils.formatKvString(map);
     }
+
+    @Test
+    public void mapEqualsTest() {
```
yizhouxw, line 103: unit testing case naming follow style:
`{methodUnderTest}_{scenario}_{expectedResult}`

server/odc-service/src/main/java/com/oceanbase/odc/service/task/executor/TaskResult.java
```diff
@@ -17,27 +17,48 @@
...
+    private String errorMessage;
 
-    String getErrorMessage();
+    private double progress;
 
-    double getProgress();
+    private Map<String, String> logMetadata;
 
-    Map<String, String> getLogMetadata();
+    public boolean progressChanged(TaskResult previous) {
```
yizhouxw, line 49: progressChanged --> isProgressChanged

server/odc-common/src/main/java/com/oceanbase/odc/common/util/MapUtils.java
LioRoger: unused method， has removed

server/odc-common/src/main/java/com/oceanbase/odc/common/util/MapUtils.java
LioRoger: agree

server/odc-common/src/test/java/com/oceanbase/odc/common/util/MapUtilsTest.java
LioRoger: split in four methods

yizhouxw (approved):
lgtm

server/odc-common/src/main/java/com/oceanbase/odc/common/util/MapUtils.java
```diff
@@ -128,4 +129,34 @@ public static int size(final Map<?, ?> map) {
...
+     * compare map, null map and empty map consider as equals
+     * 
+     * @param src
+     * @param target
+     * @param <Key> key of map
+     * @param <Value> value of map
+     * @param equalFunction function with argument value to compare if value instance is equals
+     * @return true is map is equals
+     */
+    public static <Key, Value> boolean isEqual(Map<Key, Value> src, Map<Key, Value> target,
```
yhilmare, line 142: `equals` may be better?
just like `Objects.equals()`

yhilmare (approved):
lgtm

PR #3667, opened 2024-10-12
fix(sync-database): modify syncDatabaseTaskExecutor QueueCapacity to Integer.MAX_VALUE

PeachThinking:
#### What type of PR is this?
type-bug

#### What this PR does / why we need it:
modify `syncDatabaseTaskExecutor`QueueCapacity  to `Integer.MAX_VALUE` to avoid the situation where the data source cannot trigger synchronization forever after the submitted synchronization database task exceeds the maximum number of thread pools.

#### Which issue(s) this PR fixes:

Fixes #3670

#### Additional documentation e.g., usage docs, etc.:

```docs

```

PR #3829, opened 2024-11-12
fix(datasource): datasource sync failed

MarkPotato777:
#### What type of PR is this?

type-bug
This PR fixed two issues:
1. The `RejectedExecutionHandler` of the `syncDatabaseTaskExecutor` is `DiscardPolicy` which would cause to miss the datasources for sync. The solution is to use `CallerRunsPolicy`.
2. It didn't consider the situation where the tenant did not exist. This solution adds the handler by parsing the error message.

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/config/ScheduleConfiguration.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/config/ScheduleConfiguration.java
@@ -195,7 +195,7 @@ public class ScheduleConfiguration {
         executor.setWaitForTasksToCompleteOnShutdown(true);
         executor.setAwaitTerminationSeconds(5);
         executor.setTaskDecorator(new TraceDecorator<>());
-        executor.setRejectedExecutionHandler(new ThreadPoolExecutor.DiscardPolicy());
+        executor.setRejectedExecutionHandler(new ThreadPoolExecutor.CallerRunsPolicy());
         executor.initialize();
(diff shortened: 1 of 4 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/config/ScheduleConfiguration.java
```diff
@@ -195,7 +195,7 @@ public ThreadPoolTaskExecutor syncDatabaseTaskExecutor() {
         executor.setWaitForTasksToCompleteOnShutdown(true);
         executor.setAwaitTerminationSeconds(5);
         executor.setTaskDecorator(new TraceDecorator<>());
-        executor.setRejectedExecutionHandler(new ThreadPoolExecutor.DiscardPolicy());
```
yhilmare, line 198: this pr #3667 has already fix this problem, we may cherry-pick it

MarkPotato777: I feel the solution in pr #3667 is not the best solution. We submit the task in a for loop in the schedule, which is a fixed delay schedule. If we just make the queueSize to unlimited, the fixed delay schedule would actually  become a fixed rate schedule. If we set the rejected strategy to CallerRunsPolicy, this problem would be solved. @PeachThinking

yhilmare (approved):
lgtm

PR #3862, opened 2024-11-18
fix(schedule): list datasources with attributes

guowl3:
#### What type of PR is this?

type-bug

list datasources with attributes in the schedule list

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #3879, opened 2024-11-20
feat(datasource): supports creating and testing OSS data sources

guowl3:
#### What type of PR is this?
type-feature

1. Supports creating OSS datasources but is only used for the target datasource of the archive task
2. Supports testing OSS datasources

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionTesting.java
```diff
@@ -97,7 +99,9 @@ public ConnectionTestResult test(@NotNull @Valid TestConnectionReq req) {
         if (req.getAccountType() == ConnectionAccountType.SYS_READ) {
             connectionConfig.setDefaultSchema(null);
         }
-        return test(connectionConfig);
+        return connectionConfig.getDialectType() == DialectType.FILE_SYSTEM
```
LioRoger, line 102: this switch is not easy to extend,  if a new type involved, it may missed  to change this code.
can we designed a new way to  force the user feel this method?

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/FileSystemConnectionTesting.java
```diff
@@ -0,0 +1,153 @@
...
+import lombok.NonNull;
+
+/**
+ * @Author：tinker
+ * @Date: 2024/11/19 11:24
+ * @Descripition:
+ */
+
+@Component
+public class FileSystemConnectionTesting {
```
LioRoger, line 52: Testing means in do test process. it's a verb, use noun is more suitable

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/FileSystemConnectionTesting.java
```diff
@@ -0,0 +1,153 @@
...
+
+    private static final String COS_ENDPOINT_REGEX = "cos\\.(\\w+-\\w+)\\.myqcloud\\.com";
```
LioRoger, line 54: as this regex will be compiled to Pattern. declare pattern, please confirm Pattern is thread safe

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/FileSystemConnectionTesting.java
```diff
@@ -0,0 +1,153 @@
...
+            cloudClient.putObject(storageConfig.getBucketName(), objectKey,
+                    new ByteArrayInputStream(TMP_TEST_DATA.getBytes(StandardCharsets.UTF_8)), new ObjectMetadata());
+            DeleteObjectsRequest deleteObjectsRequest = new DeleteObjectsRequest();
+            deleteObjectsRequest.setBucketName(storageConfig.getBucketName());
+            deleteObjectsRequest.setKeys(Collections.singletonList(objectKey));
+            cloudClient.deleteObjects(deleteObjectsRequest);
+            return ConnectionTestResult.success(config.getType());
+        } catch (CloudException e) {
+            if (e.getCause() instanceof OSSException) {
+                OSSException cause = (OSSException) e.getCause();
```
LioRoger, line 87: aware of npe and class cast exception

server/odc-service/src/main/java/com/oceanbase/odc/service/objectstorage/cloud/client/AmazonCloudClient.java
```diff
@@ -182,6 +182,23 @@ public PutObjectResult putObject(String bucketName, String key, File file, Objec
         });
     }
 
+    @Override
+    public PutObjectResult putObject(String bucketName, String key, InputStream in, ObjectMetadata metadata)
+            throws CloudException {
+        return callAmazonMethod("Put object", () -> {
+            com.amazonaws.services.s3.model.ObjectMetadata objectMetadata = toS3(metadata);
```
LioRoger, line 189: is this package prefix needed?

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/FileSystemConnectionTesting.java
```diff
@@ -0,0 +1,153 @@
...
+                return CloudProvider.ALIBABA_CLOUD;
+            default:
+                throw new UnExpectedException();
+        }
+    }
+
+    private String getEndPointRegex(ConnectType type) {
+        switch (type) {
+            case COS:
+                return COS_ENDPOINT_REGEX;
```
LioRoger, line 138: as you have so many switch case around your code, reconstruct may needed.
let enum's member hold this fields. or put what programmer must implement to enums class to aware if you want to add some type ,you should do like this way

PR #3883, opened 2024-11-20
feat(dlm): data archive support generating dynamic target table name

kiko-art:
#### What type of PR is this?
type-feature

ata archive support generating dynamic target table name

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/job/DataArchiveJob.java
```diff
@@ -48,6 +48,11 @@ private void executeInTaskFramework(JobExecutionContext context) {
         parameters.setScheduleTaskId(taskEntity.getId());
         parameters.setJobType(JobType.MIGRATE);
         parameters.setTables(dataArchiveParameters.getTables());
+        for (DataArchiveTableConfig tableConfig : parameters.getTables()) {
```
guowl3, line 51: Why not write this code in the existing loop body?

guowl3 (approved):
lgtm

yhilmare (approved):
lgtm

PR #3891, opened 2024-11-21
fix(schedule): schedule can be edited while subtasks are still in progress 

guowl3:
#### What type of PR is this?
type-bug

FYI

Fixes #

```docs

```


yhilmare (approved):
lgtm

PR #3892, opened 2024-11-21
fix(sql-check): cannot get oracle affect sql rows

### LINKED ISSUE #3789: [Bug]: The insert statement does not take effect in Limit the number of lines affected by SQL
### ODC version

4.3.3

### OB version

Native Oracle

### What happened?

Set the number of affected rows to 0, select oracle, select Require approval, and execute an insert statement, in which case the insert statement is executed normally
image

Trigger the approval process

Set the number of affected rows to 0, select oracle, select Require approval, and execute an insert statement, in which case the insert statement is executed normally

_No response_

zijiacj:

Fixes #3789

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/OracleAffectedRowsExceedLimit.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/OracleAffectedRowsExceedLimit.java
@@ -115,7 +115,7 @@ public class OracleAffectedRowsExceedLimit extends BaseAffectedRowsExceedLimit {
                 "SELECT * FROM TABLE(DBMS_XPLAN.DISPLAY('PLAN_TABLE', '" + ODC_TEMP_EXPLAIN_STATEMENT_ID + "', 'ALL'))";
         List<String> queryResults = jdbcOperations.query(getPlanSql, (rs, rowNum) -> rs.getString("PLAN_TABLE_OUTPUT"));
         long estRowsValue = 0;
-        for (int rowNum = 5; rowNum < queryResults.size(); rowNum++) {
+        for (int rowNum = 2; rowNum < queryResults.size(); rowNum++) {
             String resultRow = queryResults.get(rowNum);
             estRowsValue = getEstRowsValue(resultRow);
             if (estRowsValue != 0) {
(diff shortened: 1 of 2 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/OracleAffectedRowsExceedLimit.java
```diff
@@ -115,7 +115,7 @@ private long getOracleAffectedRows(String originalSql, JdbcOperations jdbcOperat
                 "SELECT * FROM TABLE(DBMS_XPLAN.DISPLAY('PLAN_TABLE', '" + ODC_TEMP_EXPLAIN_STATEMENT_ID + "', 'ALL'))";
         List<String> queryResults = jdbcOperations.query(getPlanSql, (rs, rowNum) -> rs.getString("PLAN_TABLE_OUTPUT"));
         long estRowsValue = 0;
-        for (int rowNum = 5; rowNum < queryResults.size(); rowNum++) {
+        for (int rowNum = 2; rowNum < queryResults.size(); rowNum++) {
```
yhilmare, line 118: why 2?

zijiacj: image
4 or before 4 is fine

yhilmare: why do you set 5 before? does ob-mysql has the same problems?

zijiacj: no

zijiacj: image
The index will not get the value of column row from index 5

yhilmare: from row 0 may be better?

yhilmare (approved):
lgtm

PR #3894, opened 2024-11-22
fix(pl-edit): procedure name in drop procedure statements are recognized as the table name in DBSchemaExtractor

zijiacj:
#### What type of PR is this?
type-bug

image
When doing pl editing ,procedure name in drop procedure statement is recognized as the table name, which trigger the table permission check.
the improved approach is to have DBSchemaExtractor omit the procedure name when extracting the table names

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/DBSchemaExtractor.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/DBSchemaExtractor.java
@@ -234,15 +234,6 @@ public class DBSchemaExtractor {
         }
 
         @Override
-        public RelationFactor visitDrop_view_stmt(
-                com.oceanbase.tools.sqlparser.obmysql.OBParser.Drop_view_stmtContext ctx) {
-            ctx.table_list().relation_factor().stream()
-                    .map(e -> MySQLFromReferenceFactory.getRelationFactor(e).getSchema()).distinct()
-                    .forEach(e -> identities.add(new DBSchemaIdentity(e, null)));
-            return null;
-        }
-
-        @Override
         public RelationFactor visitPartition_option(Partition_optionContext ctx) {
             return null;
         }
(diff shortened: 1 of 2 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/DBSchemaExtractor.java
```diff
@@ -200,6 +200,36 @@ private static class OBMySQLRelationFactorVisitor extends OBParserBaseVisitor<Re
 
         private final Set<DBSchemaIdentity> identities = new HashSet<>();
 
+        public RelationFactor visitDrop_function_stmt(
```
yhilmare, line 203: this may be a common problem not just related to pl edition, please submit a issue first.

you may forget to process ob-oracle mode

zijiacj: For the editing pl of ob mysql, this function includes drop procedure. This pr can only solve the drop statement of ob mysql to ensure the normal operation of pl editing function. 
executing drop procedure statement in sql windows does not fall under the scope of pl editing and should fall under the scope of sql window

server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/DBSchemaExtractor.java
```diff
@@ -200,6 +200,48 @@ private static class OBMySQLRelationFactorVisitor extends OBParserBaseVisitor<Re
...
+            if (relationFactor.getSchema() != null) {
+                identities.add(new DBSchemaIdentity(relationFactor.getSchema(), null));
+            }
+            return null;
+        }
+
+        @Override
+        public RelationFactor visitDrop_view_stmt(
+                com.oceanbase.tools.sqlparser.obmysql.OBParser.Drop_view_stmtContext ctx) {
+            ctx.table_list().relation_factor().stream()
```
yhilmare, line 239: what about null value?

server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/DBSchemaExtractor.java
```diff
@@ -353,6 +395,78 @@ private static class OBOracleRelationFactorVisitor
 
         private final Set<DBSchemaIdentity> identities = new HashSet<>();
 
+        @Override
+        public RelationFactor visitDrop_package_stmt(OBParser.Drop_package_stmtContext ctx) {
+            String databaseName = ctx.relation_factor().normal_relation_factor().database_factor().getText();
```
yhilmare, line 400: we may use RelationFactor

server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/DBSchemaExtractor.java
zijiacj: remove it. view should be added in the permission control

server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/DBSchemaExtractor.java
zijiacj: ok

yhilmare (approved):
lgtm

PR #3897, opened 2024-11-22
feat(ConnectSession):  add param createConnectionSessionSupplier in nullSafeGet,  add param maxConcurrentTaskCount in createPhysicalConnectionSession 

CHLK:
#### What type of PR is this?
type-feature

support execute chart
- add param createConnectionSessionSupplier in nullSafeGet, for custom creating connectSession 
- add param maxConcurrentTaskCount in createPhysicalConnectionSession

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/session/ConnectSessionService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/session/ConnectSessionService.java
@@ -29,6 +29,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Supplier;
 
 import javax.annotation.PostConstruct;
 import javax.annotation.PreDestroy;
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/session/ConnectSessionService.java
```diff
@@ -382,6 +388,36 @@ public ConnectionSession nullSafeGet(@NotNull String sessionId, boolean autoCrea
...
     }
 
+    public ConnectionSession nullSafeGetSkipPermissionCheck(@NotNull String sessionId,
+            Supplier<ConnectionSession> createConnectionSessionSupplier) {
+        ConnectionSession session = connectionSessionManager.getSession(sessionId);
+        if (session == null) {
+            if (createConnectionSessionSupplier == null) {
+                throw new NotFoundException(ResourceType.ODC_SESSION, "ID", sessionId);
+            }
+            Lock lock = this.sessionId2Lock.computeIfAbsent(sessionId, s -> new ReentrantLock());
```
LioRoger, line 398: this lock logic can't protect session acquire in same behavior
1. if first attempt get session failed, code goto line 408, but session is created ,will this session should be cancelExpire?
2. each session id will create a lock in sessionId2Lock, this will cause memory leak, is that expected?

CHLK:
@LioRoger I've implemented your suggestion, please take a look when you have a chance.

server/odc-service/src/main/java/com/oceanbase/odc/service/session/ConnectSessionService.java
```diff
@@ -382,6 +389,37 @@ public ConnectionSession nullSafeGet(@NotNull String sessionId, boolean autoCrea
         return session;
     }
 
+    public ConnectionSession nullSafeGetSkipPermissionCheck(@NotNull String sessionId,
```
yhilmare, line 392: duplicated codes, can we reuse these codes with the method
`public ConnectionSession nullSafeGet(@NotNull String sessionId, boolean autoCreate)`?

CHLK: okay, i'll try to reuse this codes

CHLK:
@yhilmare I've implemented your suggestion, please take a look when you have a chance.

LuckyPickleZZ (approved):

PR #3898, opened 2024-11-22
feat(session): support Oracle kill session

MarkPotato777:
#### What type of PR is this?

type-feature
This PR implements the Oracle kill session feature. Additionally, it refactors the `DefaultDBSessionManage` with the following changes: 
1. Removed the confusing logic of adding comments in SQL to indicate `ServerAddress`; 
2. Unified the kill logic within `DefaultDBSessionManage`, consolidating it into a single class.

Fixes #

```docs

```

### DIFF
--- a/server/odc-core/src/main/java/com/oceanbase/odc/core/sql/execute/SessionOperations.java
+++ b/server/odc-core/src/main/java/com/oceanbase/odc/core/sql/execute/SessionOperations.java
@@ -56,7 +56,7 @@ public interface SessionOperations {
      * @param connectionIds
      * @return the map of connectionId to kill session SQL
      */
-    default Map<String, String> getKillSessionSqls(@NonNull Set<String> connectionIds) {
+    default Map<String, String> getKillSessionSqls(@NotEmpty Set<String> connectionIds) {
         return connectionIds.stream().collect(Collectors.toMap(id -> id, this::getKillSessionSql));
     }
 
--- a/server/odc-core/src/main/java/com/oceanbase/odc/core/sql/execute/SessionOperations.java
+++ b/server/odc-core/src/main/java/com/oceanbase/odc/core/sql/execute/SessionOperations.java
@@ -16,6 +16,8 @@
 package com.oceanbase.odc.core.sql.execute;
 
 import java.sql.Connection;
+import java.util.List;
+import java.util.Map;
 
 import lombok.NonNull;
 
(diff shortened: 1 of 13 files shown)


server/odc-core/src/main/java/com/oceanbase/odc/core/sql/execute/SessionOperations.java
```diff
@@ -30,4 +32,21 @@ public interface SessionOperations {
 
     void killQuery(@NonNull Connection connection, @NonNull String connectionId);
 
+    /**
+     * Get kill query SQL by connectionId
+     *
+     * @param connectionIds
+     * @return the map of connectionId to kill query SQL
+     */
+    Map<String, String> getKillQuerySqls(@NonNull List<String> connectionIds);
```
yhilmare, line 41: the return value `Map<String, String>` is not easy to understand, how about this:
```Java
String getKillQuerySql(@NonNull String connectionId);
String getKillSessionSql(@NonNull String connectionId);
```

MarkPotato777: Then the calling side needs to handle the scenario of getting SQLs in batch, which is not easy to use.
I think as the input parameter is a list, the return value, a map, does not seem that hard to understand.

yhilmare: SessionOperations.java is used in SessionExtensionPoint which is our plugin interface and it is not a internal interface~, this interface may be impled by out developer. We should not design interfaces only from the perspective of our own use.

server/odc-core/src/main/java/com/oceanbase/odc/core/sql/execute/SessionOperations.java
```diff
@@ -30,4 +32,21 @@ public interface SessionOperations {
...
+
+
+    /**
+     * Get kill session SQL by connectionId
+     *
+     * @param connectionIds
+     * @return the map of connectionId to kill session SQL
+     */
+    Map<String, String> getKillSessionSqls(@NonNull List<String> connectionIds);
```
LioRoger, line 50: actually, input as list, but output as map, there maybe some question for me.
1. if output map key is connectionId, will input list guarantee connectionIds will not duplicate? if so , set is more suitable
2. if batch is implement by single query, provide single query method is needed.

MarkPotato777: 👍🏻 Nice suggestions. I create two single methods:  `getKillQuerySql` and `getKillSessionSql`. The batch method will have a default implementation in the interface which relies on these two single methods. And the batch methods' input has been changed to a `Set`.

server/odc-core/src/main/java/com/oceanbase/odc/core/sql/execute/SessionOperations.java
MarkPotato777: I will change it by this way: 
I create two single methods: `getKillQuerySql` and `getKillSessionSql`. The batch method will have a default implementation in the interface which relies on these two single methods. And the batch methods' input has been changed to a `Set`.

yhilmare (approved):
lgtm

yhilmare (approved):
lgtm

yizhouxw (approved):
lgtm

PR #3921, opened 2024-11-27
fix(schedule): delete check sub task status

kiko-art:
#### What type of PR is this?
type-bug

delete check sub task status

Fixes #

```docs

```


guowl3 (approved):
lgtm

PR #3937, opened 2024-12-02
refactor(schedule): fill attribute at scheduleMapper

ungreat:

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/factory/ScheduleResponseMapperFactory.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/factory/ScheduleResponseMapperFactory.java
@@ -438,6 +438,7 @@ public class ScheduleResponseMapperFactory {
         Map<Long, ConnectionConfig> id2Connection =
                 dataSourceService.internalListSkipUserCheck(connectionIds, false, false)
                         .stream().collect(Collectors.toMap(ConnectionConfig::getId, o -> o));
+        dataSourceService.fullFillAttributes(id2Connection.values());
         databases.forEach(database -> {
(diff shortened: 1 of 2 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/factory/ScheduleResponseMapperFactory.java
```diff
@@ -90,6 +91,9 @@ public class ScheduleResponseMapperFactory {
     @Autowired
     private ConnectionService dataSourceService;
 
+    @Autowired
+    private ConnectionAttributeRepository attributeRepository;
```
LioRoger, line 95: I don't see any usage for this member.

ungreat: removed useless field

server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/factory/ScheduleResponseMapperFactory.java
```diff
@@ -438,6 +438,7 @@ private List<Database> getDatabaseByIds(Set<Long> ids) {
         Map<Long, ConnectionConfig> id2Connection =
                 dataSourceService.internalListSkipUserCheck(connectionIds, false, false)
```
guowl3, line 439: fullFillAttributes method is executed in dataSourceService.internalListSkipUserCheck

ungreat: may fix the same bug.

guowl3 (approved):
lgtm

PR #3947, opened 2024-12-03
fix(osc):add version whitelist  to enable lock table feature

LioRoger:
…range is 4.2.[5 - 9]

#### What type of PR is this?

type-bug
1. lock table feature for ObMySQL version is [4.2.5 ~ 4.3.0) for now.
2. previous logic will enable lock table for Ob version greater than 4.2.5 which have no rollback control for 4.3.x
3. introduce white list format in json array, default value is  [4.2.[5 - 9].*].  matched with java regex lib.
4. value is stored in config_system_configuration with key "odc.osc.support-lock-table-ob-version-json" , use sql like :"insert into config_system_configuration(`key`, `application`, `profile`, `label`, `description`, value) values('odc.osc.support-lock-table-ob-version-json', 'odc', 'default', 'master', 'control odc enable lock table version, default is ["4\\\\.2\\\\.5.*"]', '["4\\\\.2\\\\.5.*"]')"    to init it. 
5. if config_system_configuration has no rule, default value in memory is "4\.2\.[5-9].*."
6. ODC upgrade will not affect previous stored value.
 
Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/rename/LockTableSupportDecider.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/rename/LockTableSupportDecider.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2023 OceanBase.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.oceanbase.odc.service.onlineschemachange.rename;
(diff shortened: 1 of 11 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/onlineschemachange/rename/LockTableSupportDecider.java
```diff
@@ -0,0 +1,82 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * @author longpeng.zlp
+ * @date 2024/12/2 16:30
+ */
+@Slf4j
+public class LockTableSupportDecider {
+    public static final LockTableSupportDecider DEFAULT_LOCK_TABLE_DECIDER =
+            new LockTableSupportDecider(Arrays.asList("4\\.2\\.[5-9].*"));
```
MarkPotato777, line 36: We have `VersionUtils` to compare versions.

LioRoger: this is regex match, not accurate version compare

MarkPotato777 (approved):
LGTM

LuckyPickleZZ (commented):
lgtm

LuckyPickleZZ (approved):

yhilmare (approved):
lgtm

PR #3952, opened 2024-12-03
fix(task): rollback DestroyExecutorJob to do destroy job action

LioRoger:
#### What type of PR is this?

type-bug
job_job.executor_destroy_time is needed for judge if job has been destroyed and will be used calculate task throttle. 
if there exits more than 5 task with status not in preparing but executor_destroy_time is null. new task will not dispatched.
rollback DestroyExecutorJob class to destroy executor which will update executor_destroy_time as well.

Fixes #

```docs

```


guowl3 (approved):
lgtm

yhilmare (approved):
lgtm

PR #3954, opened 2024-12-04
feat(workflow): integration airflow rest api

LuckyPickleZZ:
add cluster endpoint adaptor


server/odc-service/src/main/java/com/oceanbase/odc/service/task/util/HttpClientUtils.java
```diff
@@ -59,6 +59,10 @@ public static void setHttpClient(OkHttpClient httpClient) {
         HttpClientUtils.httpClient = httpClient;
     }
 
+    public static OkHttpClient getHttpClient() {
+        return httpClient;
```
guowl3, line 63: used for what?

LuckyPickleZZ: need to use custom header.

guowl3 (approved):
lgtm

PR #3961, opened 2024-12-04
fix(session): add svrIp in session list

ungreat:

Fixes #

```docs

```


libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/stats/mysql/OBMySQLNoLessThan400StatsAccessor.java
```diff
@@ -41,6 +41,7 @@ public class OBMySQLNoLessThan400StatsAccessor extends OBMySQLStatsAccessor {
             + "  STATE, "
             + "  `USER_CLIENT_IP` as HOST, "
             + "  HOST as PROXY_HOST, "
+            + "  CONCAT(SVR_IP, ':', SVR_PORT) AS SVR_IP, "
```
PeachThinking, line 44: Does ob mysql and oracle support this field only in versions greater than 4.0?

ungreat: Yes， and only  ob versions greater than 4.0 will use svrIp as well.

PeachThinking (approved):
lgtm

PR #3967, opened 2024-12-05
fix(session): processlIst's sessionId not unique from ob4.2.5 and 4.3.x

ungreat:
#### What type of PR is this?
add svrIp into killSession API.
svrIp(with port) + sessionId is unique.

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/db/session/DBSessionService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/db/session/DBSessionService.java
@@ -37,7 +37,7 @@ import com.oceanbase.odc.core.shared.Verify;
 import com.oceanbase.odc.core.shared.model.OdcDBSession;
 import com.oceanbase.odc.core.sql.util.OdcDBSessionRowMapper;
 import com.oceanbase.odc.service.db.browser.DBStatsAccessors;
-import com.oceanbase.odc.service.db.session.KillSessionOrQueryReq.SessionIdSvpIp;
+import com.oceanbase.odc.service.db.session.KillSessionOrQueryReq.SessionIdSvrIp;
 import com.oceanbase.tools.dbbrowser.model.DBSession;
 import com.oceanbase.tools.dbbrowser.stats.DBStatsAccessor;
 
@@ -77,7 +77,7 @@ public class DBSessionService {
     }
 
     public List<SessionIdKillSql> getKillSql(@NonNull ConnectionSession session,
-            @NonNull List<SessionIdSvpIp> sessionIds,
+            @NonNull List<SessionIdSvrIp> sessionIds,
             String closeType) {
         List<OdcDBSession> allSession = list(session);
         Map<String, List<OdcDBSession>> sidMap = allSession.stream().collect(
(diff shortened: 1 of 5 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/db/session/DBSessionService.java
```diff
@@ -74,25 +76,31 @@ public List<OdcDBSession> list(@NonNull ConnectionSession session) {
...
-    public List<SessionIdKillSql> getKillSql(@NonNull ConnectionSession session, @NonNull List<String> sessionIds,
+    public List<SessionIdKillSql> getKillSql(@NonNull ConnectionSession session,
+            @NonNull List<SessionIdSvpIp> sessionIds,
             String closeType) {
         List<OdcDBSession> allSession = list(session);
-        Map<String, String> sessionId2SvrpIp =
-                allSession.stream().collect(
-                        Collectors.toMap(OdcDBSession::getSessionId,
-                                s -> MoreObjects.firstNonNull(s.getSvrIp(), "")));
+        Map<String, List<OdcDBSession>> sidMap = allSession.stream().collect(
```
MarkPotato777, line 83: If we can assume `sessionId` is unique, how about returning `Map<String, OdcDBSession>` which is more understandable, instead of `Map<String, List<OdcDBSession>>`. Then when you get `OdcDBSession` from the map, you do not need to `get(0)` from the session list which is kind of tricky.

ungreat: from ob4.2.5 and 4.3.x, sessionId here can't guaranteed to be unique.

server/odc-service/src/main/java/com/oceanbase/odc/service/db/session/DBSessionService.java
```diff
@@ -74,25 +76,31 @@ public List<OdcDBSession> list(@NonNull ConnectionSession session) {
...
             }
-            sqlBuilder.append(sid);
-            if (sessionId2SvrpIp.get(sid) != null) {
-                sqlBuilder.append(" /*").append(sessionId2SvrpIp.get(sid)).append("*/");
+            sqlBuilder.append(sid.getSessionId());
+            List<OdcDBSession> odcDBSessions = sidMap.get(sid.getSessionId());
+            if (CollectionUtils.isNotEmpty(odcDBSessions)) {
+                String svpIp = sid.getSvrIp();
+                if (svpIp != null) {
+                    Verify.verify(odcDBSessions.stream().anyMatch(s -> s.getSvrIp().equals(svpIp)), "Invalid SvpId");
```
MarkPotato777, line 97: typo: `svpIp`

server/odc-service/src/main/java/com/oceanbase/odc/service/db/session/DBSessionService.java
MarkPotato777: The logic is kind of tricky here, we may add some comments... Like why sessionId is not unique and explain why get(0) here makes sense.

ungreat:
Not support in current milestone

PR #3979, opened 2024-12-09
fix(session): add non support kill version

ungreat:
#### What type of PR is this?
"Starting from version OB4.2.5, OBServer supports client sessions, but the current version of ODC is not yet compatible with this feature.

Fixes #

```docs

```


server/odc-migrate/src/main/resources/migrate/common/V_4_3_2_21__add_kill_non_support_version.sql
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+INSERT INTO `config_system_configuration` (`key`, `value`, `application`, `profile`, `label`, `description`)
+VALUES ('non_support_kill_min_version', '4.2.5', 'odc', 'default', 'master', 'min version of non support kill session') ON DUPLICATE KEY UPDATE `id`=`id`;
```
MarkPotato777, line 18: 1. Follow config_system_configuration's key style, key name may be like `odc.session.kill-query-or-session.max-ob-version`?
2. Is it more appropriate to put the SQL in `data.sql`, in case your code relies on the refreshable config?

ungreat: 1、That's more reasonable ~
2、The absence of this configuration means that the OB version is not restricted, and placing this configuration in data.sql will always insert this configuration.

PR #3991, opened 2024-12-09
fix(actuator): diasble actuator by default

### LINKED ISSUE #3981: [Bug]: The actualtor occupies port 8089, which conflicts with the oms port and causes startup failure.
### ODC version

ODC4.3.2-bp2

### OB version

No related

### What happened?

The actualtor occupies port 8089, which conflicts with the oms port and causes startup failure.

The private cloud version cancels the cloud monitoring related features.

--

_No response_

ungreat:

Fixes #3981

```docs

```

### DIFF
--- a/server/odc-server/src/main/resources/config/application.yml
+++ b/server/odc-server/src/main/resources/config/application.yml
@@ -75,7 +75,7 @@ management:
     web:
       exposure:
         include: "business,application"
-    enabled-by-default: false
+    enabled-by-default: ${odc.system.monitor.actuator.enabled:false}
   metrics:
     export:
       prometheus:
--- a/server/odc-server/src/main/resources/config/application.yml
+++ b/server/odc-server/src/main/resources/config/application.yml
@@ -84,5 +84,3 @@ management:
   endpoint:
     prometheus:
       enabled: ${odc.system.monitor.actuator.enabled:false}
-  server:
-    port: ${odc.system.monitor.actuator.port:-1}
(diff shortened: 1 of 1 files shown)


server/odc-server/src/main/resources/config/application.yml
```diff
@@ -75,8 +75,14 @@ management:
     web:
       exposure:
         include: "business,application"
+    enabled-by-default: ${odc.system.monitor.actuator.enabled:false}
```
MarkPotato777, line 78: The config in `application.yml` could be directly insert into `config_system_configuration`. For example, `INSERT INTO config_system_configuration(`key`, `value`, `description`) VALUES('management.endpoints.enabled-by-default', 'false', 'xxxxxx') ON DUPLICATE KEY UPDATE `id`=`id`;`

server/odc-server/src/main/resources/config/application.yml
```diff
@@ -75,8 +75,14 @@ management:
...
+      prometheus:
+        enabled: ${odc.system.monitor.actuator.enabled:false}
+
   endpoint:
     prometheus:
-      enabled: true
+      enabled: ${odc.system.monitor.actuator.enabled:false}
   server:
-    port: 8089
+    port: ${odc.system.monitor.actuator.port:-1}
```
MarkPotato777, line 88: Is -1 a illegal port value?

server/odc-server/src/main/resources/config/application.yml
ungreat: in order to use same key to control

server/odc-server/src/main/resources/config/application.yml
ungreat: Illegal port value, removed default value. Insert value if need to

MarkPotato777 (approved):
LGTM

PR #3993, opened 2024-12-10
feat(session): SQL console connection keep alive

MarkPotato777:
#### What type of PR is this?

type-feature
This PR adds a new feature for `SingleConnectionDataSource` which is keep alive. When keepAlive set to true, there will be a schedule task to execute `SELECT 1 FROM DUAL` every 5 minutes to keep the connection alive. If it fails more than 5 times in a row, the keep alive task will be terminated.

Fixes #

```docs

```

### DIFF
--- a/server/odc-core/src/main/java/com/oceanbase/odc/core/datasource/SingleConnectionDataSource.java
+++ b/server/odc-core/src/main/java/com/oceanbase/odc/core/datasource/SingleConnectionDataSource.java
@@ -154,12 +154,12 @@ public class SingleConnectionDataSource extends BaseClassBasedDataSource impleme
      */
     public synchronized void resetConnection() throws SQLException {
         log.info("The connection will be reset soon");
-        close();
+        closeConnection();
         this.connection = null;
         this.lock = null;
+        this.failedKeepAliveAttempts.set(0);
         try (Connection conn = innerCreateConnection()) {
             onConnectionReset(conn);
-            initKeepAliveScheduler();
         }
     }
 
@@ -172,19 +172,24 @@ public class SingleConnectionDataSource extends BaseClassBasedDataSource impleme
         if (Objects.nonNull(keepAliveScheduler)) {
             shutdownKeepAliveScheduler();
         }
-        if (!Objects.isNull(this.connection)) {
-            try {
-                this.connection.close();
-            } catch (Throwable throwable) {
-                log.error("Failed to close the connection", throwable);
-            }
-        }
+        closeConnection();
     }
 
     protected void prepareConnection(Connection con) throws SQLException {
         Boolean autoCommit = getAutoCommit();
         if (autoCommit != null && con.getAutoCommit() != autoCommit) {
             con.setAutoCommit(autoCommit);
+        }
+    }
+
+
+    private void closeConnection() {
+        if (!Objects.isNull(this.connection)) {
+            try {
+                this.connection.close();
+            } catch (Throwable throwable) {
+                log.error("Failed to close the connection", throwable);
+            }
         }
     }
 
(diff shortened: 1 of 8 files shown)


server/odc-core/src/main/java/com/oceanbase/odc/core/datasource/SingleConnectionDataSource.java
```diff
@@ -235,6 +261,38 @@ private synchronized Connection innerCreateConnection() throws SQLException {
...
+                return;
+            }
+            try (Connection conn = getConnection()) {
+                try (Statement statement = conn.createStatement()) {
+                    statement.execute(keepAliveSql);
+                    failedKeepAliveAttempts.set(0);
+                    log.debug("Keep connection alive success");
+                }
+            } catch (SQLException e) {
+                log.warn("Failed to keep connection alive", e);
```
yhilmare, line 283: we may not print out the exception stack or lead to too much log printing

server/odc-core/src/main/java/com/oceanbase/odc/core/datasource/SingleConnectionDataSource.java
```diff
@@ -66,15 +71,32 @@ public class SingleConnectionDataSource extends BaseClassBasedDataSource impleme
     protected volatile Connection connection;
     private final List<ConnectionInitializer> initializerList = new LinkedList<>();
     private volatile Lock lock;
+    private final boolean keepAlive;
+    private long keepAliveIntervalMillis = 5 * 60 * 1000;
+    @Getter
+    private final String keepAliveSql = "SELECT 1 FROM DUAL";
```
yhilmare, line 77: this keep live sql should be allowed to set, this datasource is not just designed for ob

server/odc-core/src/main/java/com/oceanbase/odc/core/datasource/SingleConnectionDataSource.java
```diff
@@ -235,6 +261,38 @@ private synchronized Connection innerCreateConnection() throws SQLException {
         }
     }
 
+
+    private void initKeepAliveScheduler() {
+        if (!keepAlive) {
+            return;
+        }
+        failedKeepAliveAttempts.set(0);
```
yhilmare, line 269: this method seems not be re-entried?
will the `keepAliveScheduler` be shutdown when I try to call this method twice?

MarkPotato777: Will not reconstruct the scheduler when reset connection.

server/odc-core/src/main/java/com/oceanbase/odc/core/datasource/SingleConnectionDataSource.java
```diff
@@ -136,13 +169,10 @@ public void addInitializer(@NonNull ConnectionInitializer connectionInitializer)
...
     @Override
     public void close() {
-        if (!Objects.isNull(this.connection)) {
-            try {
-                this.connection.close();
-            } catch (Throwable throwable) {
-                log.error("Failed to close the connection", throwable);
-            }
+        if (Objects.nonNull(keepAliveScheduler) && !keepAliveScheduler.isTerminated()) {
+            shutdownKeepAliveScheduler();
```
yhilmare, line 173: we may try-catch this to avoid possible exception throwing

server/odc-core/src/main/java/com/oceanbase/odc/core/datasource/SingleConnectionDataSource.java
```diff
@@ -235,6 +276,38 @@ private synchronized Connection innerCreateConnection() throws SQLException {
         }
     }
 
+
+    private void initKeepAliveScheduler() {
+        if (!keepAlive) {
```
yhilmare, line 281:
```java
if (!keepAlive || this.keepAliveScheduler != null){
    return;
}
```

this may be more robust

server/odc-core/src/main/java/com/oceanbase/odc/core/datasource/SingleConnectionDataSource.java
MarkPotato777: done

server/odc-core/src/main/java/com/oceanbase/odc/core/datasource/SingleConnectionDataSource.java
MarkPotato777: done

yhilmare (dismissed):
lgtm

LioRoger (commented):
session count of odc should be counted, if it's a large amount like 100 or more. Executor may be need combined.

MarkPotato777:
> session count of odc should be counted, if it's a large amount like 100 or more. Executor may be need combined.

 The modification costs are relatively high. Let's stick to the current plan and run it in the testing environment. If there are performance issues, I will refactor it.

LioRoger (approved):
lgtm

PR #3998, opened 2024-12-11
fix(schdule):delete schedule with check sub running tasks

kiko-art:
check subtask when delete schedule.

#### What type of PR is this?
type-bug

delete schedule with check sub running tasks.

Fixes #

```docs

```


guowl3 (approved):
lgtm

PR #4026, opened 2024-12-16
fix(db-browser): failed to get table due to create extended stats in oracle mode

### LINKED ISSUE #3134: [Bug]: export and get table detail would fail if there are pseudocolumns in oracle table
### ODC version

4.x

### OB version

oracle

### What happened?

export would fail if there are pseudocolumns in oracle table

-

```
CREATE TABLE stats_test
   (    "X" NUMBER(*,0),
        "Y" NUMBER(*,0),
        "Z" CHAR(2 CHAR)
   );

SELECT DBMS_STATS.CREATE_EXTENDED_STATS('JINGTIAN2', 'STATS_TEST', '(X, Y)') FROM DUAL;
```

create a table with pseudocolumns.
you could refer https://www.askmac.cn/archives/oracle-virtual-column.html

_No response_

PeachThinking:
… in column

#### What type of PR is this?
type-bug

Fix failed to get table due to create extended stats in oracle mode.
All LONG or LONG RAW columns have to be retrieved from the ResultSet prior to all the other columns or oracle jdbc will throw “Stream has already been closed” Exception

Fixes #3134 

```docs

```


libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/model/DBTableColumn.java
```diff
@@ -122,7 +122,7 @@ public class DBTableColumn implements DBObject, DBObjectWarningDescriptor {
     private String collationName;
 
     /**
-     * MySQL special
+     * Default value for the column
```
MarkPotato777, line 125: From the name, `genExpression` refers to the generation column's expression, not the default value. Oracle uses `COL_DATA_DEFAULT` to represent `genExpression`, which is just an implementation detail in Oracle's data dictionary views, but it does not mean that this field represents a default value.

PeachThinking: OK，I have modified the comment.

MarkPotato777 (approved):
LGTM

PR #4046, opened 2024-12-17
fix(permission):time is missing when initiating apply database task again

### LINKED ISSUE #3953: [Bug]: When you apply for library permissions and submit them again, the validity period of the permissions is different from the first time
### ODC version

4.3.3.

### OB version

no related

### What happened?

image
image
When you apply for a library permission, the permission is valid for 30 days when you submit it, and the permission for re-initiating the permission is valid for 7 days

When re-initiated, the value of the permission validity period should be the same as the validity period selected for the first submission

Submit a database permission application, select a validity period of 30 days or another time, and initiate the request again to check whether the validity period of the newly generated ticket is the same as that of the first time

_No response_

zijiacj:
#### What type of PR is this?
type-bug

Fixes #3953

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/permission/database/model/ApplyDatabaseParameter.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/permission/database/model/ApplyDatabaseParameter.java
@@ -48,9 +48,10 @@ public class ApplyDatabaseParameter implements Serializable, TaskParameters {
      */
     private List<DatabasePermissionType> types;
     /**
-     * This property is required by the front end to initiate task again
+     * This field represents the duration of the permission. It is only used for the front end to
+     * re-initiate tasks, and the backend does not rely on it
      */
-    private String expireTimeValue;
+    private String validDuration;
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/permission/database/model/ApplyDatabaseParameter.java
```diff
@@ -47,6 +47,10 @@ public class ApplyDatabaseParameter implements Serializable, TaskParameters {
      * Permission types to be applied for, required
      */
     private List<DatabasePermissionType> types;
+    /**
+     * This property is required by the front end to initiate task again
+     */
+    private String expireTimeValue;
```
MarkPotato777, line 53: - `expireTimeValue` --> `validDuration`
- Clearly specify in the comment: This field represents the duration of the permission. It is only used for the front end to re-initiate tasks, and the backend does not rely on it.

MarkPotato777 (approved):
LGTM

PR #4047, opened 2024-12-17
fix(actuator): disable actuator autoconfiguration in client mode

ungreat:

Fixes #

```docs

```


ungreat:
Record the issue https://github.com/oceanbase/odc/issues/4051 and reconstruct it later.

MarkPotato777 (approved):
We may avoid importing spring actuator dependency in client mode later.

PR #4068, opened 2024-12-18
fix(task-framework): task log file is incomplete

guowl3:

We must store the task logs in a single file. If there are multiple files, some might be missing when uploading to OSS

Fixes #

```docs

```


ysjemmm (approved):
lgtm

PR #4073, opened 2024-12-19
fix(osc): remove distinct from osc query user sql

LioRoger:
#### What type of PR is this?

type-bug
distinct key word will be case  insensitive for MySQL select query.
distinct('odc', 'ODC') will only return one value.

Fixes #

```docs

```


LuckyPickleZZ (approved):
lgtm

PR #4084, opened 2024-12-23
fix(dbbrowser): cant get correct ddl of procedure

### LINKED ISSUE #4023: [Bug]: After creating a stored procedure with parameter types set and enum, there is no value
### ODC version

odc433

### OB version

*

### What happened?

image
image

There are values in ddl memory

Create a stored procedure and change the data type to set

_No response_

zijiacj:
#### What type of PR is this?
type-bug

image
image

Currently, procedure's ddl is pieced together, and the above types of ddl will be displayed correctly.
An improved approach is to return the ddl that is retrieved directly from the database

Fixes #4023 

```docs

```

### DIFF
--- a/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoLessThan5700SchemaAccessor.java
+++ b/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoLessThan5700SchemaAccessor.java
@@ -1343,13 +1343,6 @@ public class MySQLNoLessThan5700SchemaAccessor implements DBSchemaAccessor {
                 .append(" and ROUTINE_TYPE='PROCEDURE'");
         DBProcedure procedure = new DBProcedure();
         procedure.setProName(procedureName);
-        MySQLSqlBuilder parameters = new MySQLSqlBuilder();
-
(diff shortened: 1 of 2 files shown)


libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoLessThan5700SchemaAccessor.java
```diff
@@ -1350,14 +1350,22 @@ public DBProcedure getProcedure(String schemaName, String procedureName) {
                     .identifier(rs.getString("PARAMETER_NAME")).space()
                     .append(rs.getString("DTD_IDENTIFIER")).append(",");
         });
+        MySQLSqlBuilder getDDL = new MySQLSqlBuilder();
```
MarkPotato777, line 1353: 1. `parameters` is never used, the `queryForParameters` is unnecessary.
2. please add more unit tests for `getProcedure` to ensure the parser can handle the show create DDL.

zijiacj: ok

MarkPotato777 (approved):
LGTM

PR #4087, opened 2024-12-23
fix(execution-plan): avoid invalid number

### LINKED ISSUE #3243: [Bug]: failed to get sql execution plan
### ODC version

4.3.1

### OB version

ob 4217

### What happened?

failed to get sql execution plan，when parsing the estimated number of rows column, odc was reading it as a number, but there is an unexpected` value of more than 1.0e19`, which caused the parsing error.
image

get sql execution plan successfully

when parsing the estimated number of rows column, odc was reading it as a number, but there is an unexpected` value of more than 1.0e19`, which caused the parsing error.

_No response_

LuckyPickleZZ:
#### What type of PR is this?
type-bug

Sometimes there maybe incorrect number in sql execution plan from  observer. So we should check it before casting.

Fixes #3243 

```docs

```


MarkPotato777 (approved):
LGTM

PR #4090, opened 2024-12-23
fix(flow):view export function can be initiated without permission

### LINKED ISSUE #4066: [Bug]: Views without export permission can also be exported
### ODC version

odc433

### OB version

*

### What happened?

image
image

Cannot export views without export permission

1. Cannot export views without export permission
2. Select a view without export permission to export

_No response_

zijiacj:
#### What type of PR is this?
type-bug

Fixes #4066 

```docs

```


LioRoger (approved):
lgtm, but please spread  check range

MarkPotato777 (approved):
LGTM

PR #4093, opened 2024-12-23
fix(function):The function does not display properly if the return value type is year

### LINKED ISSUE #3942: [Bug]: The data type returned by the function is inconsistent with the actual type
### ODC version

odc433

### OB version

*

### What happened?

image

Should return to 2024

1. Create Function
create function `fun_year` (`p1` year) returns year
begin
  return p1;
end;
2. Execute function

_No response_

zijiacj:
#### What type of PR is this?
type-bug

create function `fun_year` (`p1` year) returns year begin
return p1;
end;

In ob mysql tenant. The input and return parameters of the above function are both of the year type. The function is executed in black screen. the input and output are both 2024. However, the value obtained by jdbc using getString is 2024-01-01, which is inconsistent with black screen.

image
Current  the jdbc of `ob` specifies that the `year` type corresponds to `Short` in java, so changed from `getString` to `getShort`

For getting function execution results from database which dialect_type is `mysql`, use `JdbcRowMapper` to handle the type conversion

Fixes #3942 

```docs

```


server/odc-core/src/main/java/com/oceanbase/odc/core/sql/util/JdbcDataTypeUtil.java
```diff
@@ -269,7 +269,8 @@ public static JDBCType parseDataType(String dataType) {
                 || "SIGNTYPE".equalsIgnoreCase(dataType)
                 || "SIMPLE_INTEGER".equalsIgnoreCase(dataType)
                 || "MEDIUMINT".equalsIgnoreCase(dataType)
-                || "INTEGER".equalsIgnoreCase(dataType)) {
+                || "INTEGER".equalsIgnoreCase(dataType)
+                || "YEAR".equalsIgnoreCase(dataType)) {
             dataType = "INTEGER";
```
LioRoger, line 274: 1. we'd better distinguish parseDataType with db type. cause same type name in different db may represent different JDBCType.
2. why not fist check if we have matched the type, if not match then do if else logic match

zijiacj: That is the problem of the jdbc layer, I do that is just an opportunistic practice. I will communicate with people   responsible for jdbc to decide  the specific practice

zijiacj: image
have been parsed to `SMALLINT `,consistent with the jdbc specification

PeachThinking (commented):
Have you discussed the feasibility of the solution to this problem with the jdbc teachers?

PeachThinking (approved):
lgtm

guowl3 (approved):
lgtm

server/odc-core/src/main/java/com/oceanbase/odc/core/sql/util/JdbcDataTypeUtil.java
```diff
@@ -201,6 +201,14 @@ public static Object getValueFromStatement(CallableStatement statement, int inde
 
     public static Object getValueFromResultSet(ResultSet resultSet, int index, String type)
             throws SQLException {
+        // YEAR type is used to represent the year in 4-digit format. The range is 1901 to 2155, and 0000
```
LioRoger, line 204: I don't think add if before switch is a good solution.
YEAR is a special MySQL datatype, although it stored in integer, but is have some special usage.
eg year start begin 1900, and if you use in where clause, value less than 1900 is  special number need to be handled. So why not create it as a type and process it by type referenced function

zijiacj: image
image
The enum type `JDBCType` belongs to the package `java.sql`  and cannot be modified. The `year` type is not included

zijiacj: Using` getShort` will only get the values from 1901 to 2155 and 0

zijiacj: For getting function execution results from database which dialect_type is mysql, use `JdbcRowMapper` to handle the type conversion

LuckyPickleZZ (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/db/util/OBMysqlCallFunctionCallBack.java
```diff
@@ -76,23 +81,39 @@ public CallFunctionResp doInConnection(Connection con) throws SQLException, Data
...
+                if (CollectionUtils.size(rows) == 1 && Objects.nonNull(rows.get(0))
+                        && CollectionUtils.size(rows.get(0)) == 1) {
+                    CallFunctionResp callFunctionResp = new CallFunctionResp();
+                    PLOutParam plOutParam = new PLOutParam();
+                    plOutParam.setValue(String.valueOf(rows.get(0).get(0)));
                     plOutParam.setDataType(function.getReturnType());
                     callFunctionResp.setReturnValue(plOutParam);
+                    callFunctionResp.setOutParams(null);
                     return callFunctionResp;
+                } else {
```
MarkPotato777, line 100: The behavior changed if the resultset's size was not 1. Is it expected?

server/odc-service/src/main/java/com/oceanbase/odc/service/db/util/OBMysqlCallFunctionCallBack.java
```diff
@@ -76,23 +81,39 @@ public CallFunctionResp doInConnection(Connection con) throws SQLException, Data
...
-            PLOutParam plOutParam = new PLOutParam();
             try (ResultSet res = stmt.executeQuery(sqlBuilder.toString())) {
                 if (!res.next()) {
-                    plOutParam.setValue(function.getReturnValue());
+                    return generateDefaultReturnValue();
+                }
+                JdbcQueryResult jdbcQueryResult = new JdbcQueryResult(res.getMetaData(), rowDataMapper);
+                jdbcQueryResult.addLine(res);
+                List<List<Object>> rows = jdbcQueryResult.getRows();
+                if (CollectionUtils.size(rows) == 1 && Objects.nonNull(rows.get(0))
```
MarkPotato777, line 91: Use org.apache.commons.collections4, then it can be simplified to 
`if (CollectionUtils.size(rows) == 1 && CollectionUtils.size(rows.get(0)) == 1)`

server/odc-service/src/main/java/com/oceanbase/odc/service/db/util/OBMysqlCallFunctionCallBack.java
zijiacj: Not expected, should throw an exception

server/odc-service/src/main/java/com/oceanbase/odc/service/db/util/OBMysqlCallFunctionCallBack.java
zijiacj: done

MarkPotato777 (approved):
LGTM

PeachThinking (approved):

guowl3 (approved):
lgtm

PR #4095, opened 2024-12-24
fix(task): correct region key name for resourceID

LioRoger:
#### What type of PR is this?

type-bug
correct region key name for K8S mode resource Id build

Fixes #

```docs

```


guowl3 (approved):
lgtm

PR #4096, opened 2024-12-24
fix(flow): add currentUserResourceRole in flow APIs

MarkPotato777:
#### What type of PR is this?

type-bug
Add `currentUserResourceRole` in the flow API for the front end to judge if the ticket could be operated by current user.

Fixes #

```docs

```


zijiacj (approved):
lgtm

PR #4103, opened 2024-12-25
fix(schedule): invalid  schedule annotation

LuckyPickleZZ:
#### What type of PR is this?
type-bug

In some scenarios, such as obcloud, schedule annotations written in the configuration class do not take effect. This will cause the system configuration to fail to refresh


yhilmare (approved):
lgtm

PR #4105, opened 2024-12-26
fix(migrate): newly added sql check rules in 4.3.2 will not be migrated successfully

### LINKED ISSUE #3651: [Bug]:open sql console report NPE when datasource configured as custom environment

MarkPotato777:
#### What type of PR is this?

type-bug
The newly added rules will not migrate successfully because of the wrong ruleset name.

Fixes #3651

```docs

```


guowl3 (approved):
lgtm

PR #4106, opened 2024-12-26
fix(integration): external approval integration will cause NPE if hyperlinkExpression is null

### LINKED ISSUE #2753: [Bug]: bastion intgration mode did not show internal schemas
### ODC version

v4.2.4/4.3.0

### OB version

not related

### What happened?

 bastion intgration mode did not show internal schemas.

e.g. only show 'SYS' schema for oceanbase mysql mode
image

in bastion integration mode, jump link expected see all databases

omit

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug
When users don't configure hyperlinkExpression, it will cause NPE.

Fixes #2753 

```docs

```


zijiacj (approved):
lgtm

PR #4108, opened 2024-12-27
fix(db): The method of determining whether opening the global client session is incorrect

zijiacj:
#### What type of PR is this?
type-bug

The current method for determining whether a global session is enabled applies only to `system` tenants, but not to `mysql` and `oralce` tenants. Instead, should adopt this way that determining whether the result set after running `show processlist` contains the column `time`.

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/db/session/DefaultDBSessionManage.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/db/session/DefaultDBSessionManage.java
@@ -310,6 +310,7 @@ public class DefaultDBSessionManage implements DBSessionManageFacade {
                         int columnIndex = rs.findColumn("time");
                         return columnIndex > 0;
                     } catch (SQLException e) {
+                        log.warn("Failed to find the column 'time' in the result set", e);
                         return false;
                     }
                 });
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/db/session/DefaultDBSessionManage.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/db/session/DefaultDBSessionManage.java
@@ -18,6 +18,7 @@ package com.oceanbase.odc.service.db.session;
 import static com.oceanbase.odc.core.shared.constant.DialectType.OB_MYSQL;
 
 import java.sql.Connection;
+import java.sql.SQLException;
 import java.sql.Statement;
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/db/session/DefaultDBSessionManage.java
```diff
@@ -300,38 +301,18 @@ private boolean isGlobalClientSessionEnabled(ConnectionSession connectionSession
...
+        // Check whether the global session is open
+        // If the global session is open, the "time" column will be displayed in the result set after
+        // executing the sql statement of "show processlist"
+        return connectionSession.getSyncJdbcExecutor(ConnectionSessionConstants.BACKEND_DS_KEY)
+                .query("show processlist", rs -> {
+                    try {
+                        int columnIndex = rs.findColumn("time");
+                        return columnIndex > 0;
+                    } catch (SQLException e) {
+                        return false;
```
PeachThinking, line 313: Is it necessary to add a log to print the cause of the exception in this code?

zijiacj: ok

server/odc-service/src/main/java/com/oceanbase/odc/service/db/session/DefaultDBSessionManage.java
```diff
@@ -300,38 +301,18 @@ private boolean isGlobalClientSessionEnabled(ConnectionSession connectionSession
...
-                    .query("show proxyconfig like '" + configName + "';",
-                            rs -> rs.next() ? rs.getInt("value") : null);
-        } catch (Exception e) {
-            log.warn("Failed to obtain the value of OBProxy's configuration variable: {}", e.getMessage());
-            return null;
-        }
+        // Check whether the global session is open
+        // If the global session is open, the "time" column will be displayed in the result set after
+        // executing the sql statement of "show processlist"
+        return connectionSession.getSyncJdbcExecutor(ConnectionSessionConstants.BACKEND_DS_KEY)
```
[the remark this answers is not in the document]
zijiacj: yes

PeachThinking (approved):
lgtm

LioRoger (approved):
lgtm

PR #4113, opened 2024-12-30
fix(migrate): complete data for connect_database.connect_type

### LINKED ISSUE #4082: [Bug]: After adding a connect_type field to the connect_database table, need to process the value of the stock data connect_type
### ODC version

ODC433

### OB version

any

### What happened?

Some records in the connect_database table have null values for connect_type

connect_type for all records in the connect_database table must have a value

Upgrading the odc from version 432 to 433 will repeat the problem

_No response_

MarkPotato777:
#### What type of PR is this?

type-bug
In the previous version, it added `connect_type` in `connect_database`, but `connect_type` was set to `default null`, which caused dirty data. This PR completed the `connect_type` whose value is null by setting it to `connect_connection.type.`

Fixes #4082

```docs

```


zijiacj (dismissed):
lgtm

LioRoger (dismissed):
lgtm

zijiacj (approved):
lgtm

PR #4115, opened 2024-12-31
fix(db):miss quote of pl name which contains @ causes pl edit failed

zijiacj:
#### What type of PR is this?
type-bug

When pl name contains `@`, need to add quote, otherwise the create and drop statements will fail

image

Fixes #

```docs

```


PeachThinking (approved):
lgtm

PR #4116, opened 2025-01-02
fix(permission): global project role cannot see approvable tickets

MarkPotato777:
#### What type of PR is this?

type-bug
1. I added `derivedFromGlobalProjectRole` in `ProjectMember` to identify if this project role is derived from the global project role for the front end to disable/enable some features.
2. I fixed the global role acquiring method. The previous is wrong which acquires from `iam_user_permission`. The correct way is from `iam_user_role`.

Fixes #

```docs

```


PeachThinking (approved):
lgtm

PR #4118, opened 2025-01-03
fix(workflow): add dfa transfer from unknown state

LuckyPickleZZ:
#### What type of PR is this?
type-bug

Now when k8s resource's state is UNKNOWN, it would be it's final state. And when the real resource is back to normal , the state could not be changed.


yhilmare (approved):
lgtm

PR #4122, opened 2025-01-06
fix(structure-compare):the structure synchronization task cannot be initiated when the structure comparison task is not created by yourself

PeachThinking:
#### What type of PR is this?
type-bug

fix the structure synchronization task cannot be initiated when the structure comparison task is not created by yourself, because the object storage is isolated by user id so that user1 can not see user2's object storage.

Fixes #

```docs

```


MarkPotato777 (approved):
LGTM

PR #4123, opened 2025-01-06
fix(session):the sql statement of 'select function' should need database change permission

zijiacj:

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/DBSchemaExtractor.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/DBSchemaExtractor.java
@@ -308,6 +308,8 @@ public class DBSchemaExtractor {
                 if (relationName.contains(".")) {
                     String[] names = relationName.split("\\.");
                     relationFactor.setSchema(names[0]);
+                } else {
+                    relationFactor.setSchema(this.defaultSchema);
                 }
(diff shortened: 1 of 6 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/session/util/DBSchemaExtractor.java
```diff
@@ -623,15 +631,40 @@ public Void visitCurrent_schema(Current_schemaContext ctx) {
...
     @Data
     @NoArgsConstructor
     @AllArgsConstructor
     @EqualsAndHashCode
     public static class DBSchemaIdentity {
-
         private String schema;
         private String table;
+        private String dbObjectName;
+        private DBObjectType dbObjectType;
```
MarkPotato777, line 662: It doesn't make sense that it adds `dbObjectName` and `dbObjectType` when it already contains `table`.

zijiacj: table is used to control the permissions of tables and views. However, for function, table must be null, and only schem permission control needs to be done.

ysjemmm (commented):
you can may edit code better

ysjemmm (commented):
you can may edit code better

PR #4124, opened 2025-01-06
fix(permission): global project roles cannot operate database/table permission apply and schedule tickets

MarkPotato777:
#### What type of PR is this?

type-bug

This PR fixed 2 bugs:
1. Global project roles cannot operate Database/Table permissions apply tickets
2. Global project roles cannot delete schedule tickets

The first bug is because the previous PR missed adapting to these two types of work orders.
The second bug is because the front end rely on the `project` field to display some buttons but schedule details doesn't include it.

Fixes #

```docs

```


LioRoger (approved):
lgtm

PeachThinking (approved):
lgtm

guowl3 (approved):

PR #4128, opened 2025-01-07
fix(permission): external approval integration doesn't take effect

MarkPotato777:
#### What type of PR is this?

type-bug
This PR fixed 2 bugs:
1. Even if the external approval has approved the flow instance, the approval process in ODC will not proceed. The reason is that when calling `approve` method, authentication fails due to a code modification introduced in version 4.3.3. The fix is to differentiate between scenarios that require authentication and those that do not.

2. The global project owner is unable to approve because of a logical error. When constructing `resourceRoleIdentifiers`, all project IDs should be enumerated instead of using `*` as a substitute.

Fixes #

```docs

```


PeachThinking (approved):
lgtm

zijiacj (approved):
lgtm

PR #4129, opened 2025-01-07
fix(db-browser): remove @JsonProperty in DBTableStats.dataSizeInBytes

PeachThinking:
#### What type of PR is this?
type-bug

remove @JsonProperty in DBTableStats.dataSizeInBytes so that we can provided the size of table in bytes to the front end

Fixes #

```docs

```


MarkPotato777 (approved):
LGTM

PR #4133, opened 2025-01-08
feat(resource): add double check when destroy resource

LioRoger:
#### What type of PR is this?
add double check, take most effort to avoid multi invoke of destroy resource operation

Fixes #

```docs

```


guowl3 (approved):
lgtm

PR #4143, opened 2025-01-09
fix(schedule):check has running task when delete

kiko-art:
#### What type of PR is this?
type-bug

check has running task when delete

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduleService.java
```diff
@@ -330,6 +330,10 @@ public ChangeScheduleResp changeSchedule(ScheduleChangeParams req) {
                         || targetSchedule.getStatus() == ScheduleStatus.COMPLETED, ErrorCodes.DeleteNotAllowed, null,
                         "Delete schedule is not allowed.");
             }
+            if (req.getOperationType() == OperationType.DELETE) {
```
guowl3, line 333: why not use the existing branch?

guowl3 (approved):

PR #4144, opened 2025-01-09
fix(session): kill session  may happen invalid server ip address

ungreat:
#### What type of PR is this?

type-bug

The original expression is used to match hint, so the regular pattern string is `/* <ip>:<port> */`, but it is no longer obtained from hint, so `/*` and `*/ `are not needed.

Fixes #

```docs

```


MarkPotato777 (approved):
LGTM

PeachThinking (approved):
lgtm

PR #4145, opened 2025-01-09
fix(k8s): k8s component status is incorrect

LuckyPickleZZ:
#### What type of PR is this?
type-bug

fix incorrect k8s status dfa


yhilmare (approved):
lgtm

PR #4146, opened 2025-01-09
fix(logicaldatabase): it may NPE when the logical database task starts

MarkPotato777:
#### What type of PR is this?

type-bug
The logical database task may have NPE in the log, but the task will success.
This is because when `TaskMonitor` call `getTaskResult`, the `executionGroupContext` may be null.
I fixed this by moving the process of init the `executionGroupContext` into `doInit` method.

Fixes #

```docs

```


guowl3 (approved):
lgtm

LioRoger (approved):
lgtm

PR #4147, opened 2025-01-09
fix(schedule): do not request external services when describe database

guowl3:
#### What type of PR is this?
type-bug

The operation of the schedule failed because the datasource is invalid.

Fixes #

```docs

```


MarkPotato777 (approved):
LGTM

PR #4148, opened 2025-01-10
fix(session): kill oracle session may cause sockettimeout

MarkPotato777:
#### What type of PR is this?

type-bug
Kill Oracle session may cause socket timeout because the kill SQL doesn't include `IMMEDIATE`.

Fixes #

The IMMEDIATE option forces the session to terminate immediately. It doesn't wait for the session to finish current operations or release resources; instead, it interrupts the session right away.

```docs

```


guowl3 (approved):

LioRoger (approved):
lgtm

PR #4151, opened 2025-01-10
feat(dlm): support filesystem

guowl3:
#### What type of PR is this?
type-feature

1. Supports creating OSS datasources but is only used for the target datasource of the archive task
2. Supports testing OSS datasources
3. Opt several error message
4. Record the save point for rollback task

Fixes #

The all comments is revolved in old merge request .https://github.com/oceanbase/odc/pull/3879

```docs

```


server/odc-server/src/main/resources/log4j2-task.xml
```diff
@@ -64,6 +64,11 @@
                             <!-- TRACE < DEBUG < INFO < WARN < ERROR < FATAL -->
                             <ThresholdFilter level="INFO" onMatch="ACCEPT" onMismatch="DENY"/>
                         </Filters>
+                        <Policies>
```
LioRoger, line 67: this policy has exited above

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/listener/UpdateDatasourceListener.java
```diff
@@ -0,0 +1,88 @@
...
+            }
+            Optional<DatabaseEntity> existed = byConnectionId.stream().filter(
+                    o -> connectionConfig.getHost().equals(o.getName())).findFirst();
+            if (existed.isPresent()) {
+                entity = existed.get();
+            }
+        }
+        // create or update
+        entity = entity == null ? new DatabaseEntity() : entity;
+        entity.setDatabaseId(com.oceanbase.odc.common.util.StringUtils.uuid());
```
LioRoger, line 74: this package prefix is not necessary maybe

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/listener/UpdateDatasourceListener.java
```diff
@@ -0,0 +1,88 @@
...
+        entity.setName(connectionConfig.getHost());
+        entity.setProjectId(connectionConfig.getProjectId());
+        entity.setConnectionId(connectionConfig.getId());
+        entity.setEnvironmentId(connectionConfig.getEnvironmentId());
+        entity.setSyncStatus(DatabaseSyncStatus.SUCCEEDED);
+        entity.setExisted(true);
+        entity.setObjectSyncStatus(DBObjectSyncStatus.SYNCED);
+        entity.setConnectType(connectionConfig.getType());
+        entity.setType(DatabaseType.PHYSICAL);
+        databaseRepository.save(entity);
```
LioRoger, line 85: what's save means， if entity has exists， update should be more accurate
anyway， if exception will be thrown if entity has exists

server/odc-service/src/main/java/com/oceanbase/odc/service/db/schema/DBSchemaSyncTaskManager.java
```diff
@@ -118,6 +119,9 @@ public void submitTaskByDatabases(@NonNull Collection<Database> databases) {
     }
 
     public void submitTaskByDataSource(@NonNull ConnectionConfig dataSource) {
+        if (dataSource.getDialectType() == DialectType.FILE_SYSTEM) {
```
LioRoger, line 122: use dialect type to judge if task can be submitted is not a general design, when user add a new dialect type, he will not fell which code should i add or changed, so can we design a more general tag to represent something should be done or should not be done

server/odc-service/src/main/java/com/oceanbase/odc/service/objectstorage/cloud/client/AlibabaCloudClient.java
```diff
@@ -170,6 +170,23 @@ public PutObjectResult putObject(String bucketName, String key, File file, Objec
         return putObject;
     }
 
+    @Override
+    public PutObjectResult putObject(String bucketName, String key, InputStream in, ObjectMetadata metadata)
+            throws CloudException {
+        PutObjectResult putObject = callOssMethod("Put object", () -> {
+            com.aliyun.oss.model.ObjectMetadata objectMetadata = toOss(metadata);
```
LioRoger, line 177: so as this package prefix

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/listener/UpdateDatasourceListener.java
guowl3: This is a built-in JPA method with the following behavior: If the new entity's primary key does not exist in the database, the save method will perform an insert operation. If the entity's primary key already exists in the database, the save method will perform an update operation.

server/odc-service/src/main/java/com/oceanbase/odc/service/objectstorage/cloud/client/AlibabaCloudClient.java
guowl3: This class also references a custom ObjectMetadata class with the same name from ODC, so a prefix needs to be added to distinguish between them.

server/odc-server/src/main/resources/log4j2-task.xml
guowl3: fix it in https://github.com/oceanbase/odc/pull/4165

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/listener/UpdateDatasourceListener.java
guowl3: fit it in https://github.com/oceanbase/odc/pull/4165

server/odc-service/src/main/java/com/oceanbase/odc/service/db/schema/DBSchemaSyncTaskManager.java
guowl3: Instead of using DialectType, a new method isFileSystem has been added to ConnectType to determine whether a connection is a file system.

PR #4152, opened 2025-01-10
fix(db-browser): the precision of a MySQL YEAR type can be 0

### LINKED ISSUE #3913: [Bug]: Native MySQL structure comparison error "Invalid display width. Use YEAR instead"
### ODC version

ODC433

### OB version

original mysql 8.0.22

### What happened?

Native MySQL structure comparison error "Invalid display width. Use YEAR instead"
image
image

。

```
CREATE TABLE `p111111` (
  `col1` int(11) NOT NULL AUTO_INCREMENT,
  `col2` decimal(10,2) DEFAULT NULL,
  `col3` decimal(10,2) DEFAULT NULL,
  `col4` bit(8) DEFAULT NULL,
  `col5` tinyint(4) DEFAULT NULL,
  `col6` smallint(6) DEFAULT NULL,
  `col7` mediumint(9) DEFAULT NULL,
  `col8` bigint(20) DEFAULT NULL,
  `col9` float(10,2) DEFAULT NULL,
  `col10` double(10,2) DEFAULT NULL,
  `col11` varchar(10) DEFAULT NULL,
  `col12` char(10) DEFAULT NULL,
  `col13` text DEFAULT NULL,
  `col14` tinytext DEFAULT NULL,
  `col24` time DEFAULT NULL,
  `col25` date DEFAULT NULL,
  `col26` datetime DEFAULT NULL,
  `col27` year(4) DEFAULT NULL,
  PRIMARY KEY (`col1`)
)
 partition by range(col1)
(partition `p1` values less than (10000),
partition `p2` values less than (50000),
partition `p3` values less than (100000),
partition `p4` values less than (MAXVALUE))

```

_No response_

zhijiamu:

Fixes #3913

```docs

```

### DIFF
--- a/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoLessThan5700SchemaAccessor.java
+++ b/libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoLessThan5700SchemaAccessor.java
@@ -121,6 +121,7 @@ public class MySQLNoLessThan5700SchemaAccessor implements DBSchemaAccessor {
         SPECIAL_TYPE_NAMES.add("bigint");
         SPECIAL_TYPE_NAMES.add("float");
         SPECIAL_TYPE_NAMES.add("double");
+        SPECIAL_TYPE_NAMES.add("year");
     }
 
     public MySQLNoLessThan5700SchemaAccessor(@NonNull JdbcOperations jdbcOperations) {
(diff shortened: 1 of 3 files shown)


libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoLessThan5700SchemaAccessor.java
```diff
@@ -605,6 +606,11 @@ protected RowMapper<DBTableColumn> listTableRowMapper() {
 
     protected void fillPrecisionAndScale(DBTableColumn column) {
         String typeName = column.getTypeName();
+
+        // 在mysql较高版本中，ddl和表结构对比中，year的长度会被强制置为4。如果columntype=yaer(2)时，这里的year的长度会是2
```
MarkPotato777, line 610: Nice comments! It's better to use English for comments :)

MarkPotato777:
It would be even better if you could clearly explain the cause of the bug and the fix solution in the PR title and the description. You can find more information in `oceanbase/odc/docs/zh-CN/CONTRIBUTION.md`. 

If you have any questions, feel free to ask me at any time.

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/mysql/MySQLNoLessThan5700SchemaAccessor.java
zhijiamu: ok, get, i will recorrect the note to english next time

MarkPotato777 (commented):
PR title suggestion: fix(db-browser): the precision of a MySQL YEAR type can be 0

PR #4154, opened 2025-01-10
fix(sql-parser):  failed to parse table ddl when create fulltext key with parser

PeachThinking:
#### What type of PR is this?
type-bug

ob mysql can specify a word breaker using WITH PARSER {SPACE | NGRAM | BENG} only when creating a full-text index.

native mysql:https://dev.mysql.com/doc/refman/8.4/en/create-index.html
```
WITH PARSER parser_name

This option can be used only with FULLTEXT indexes. It associates a parser plugin with the index if full-text indexing and searching operations need special handling. [InnoDB](https://dev.mysql.com/doc/refman/8.4/en/innodb-storage-engine.html) and [MyISAM](https://dev.mysql.com/doc/refman/8.4/en/myisam-storage-engine.html) support full-text parser plugins. If you have a [MyISAM](https://dev.mysql.com/doc/refman/8.4/en/myisam-storage-engine.html) table with an associated full-text parser plugin, you can convert the table to InnoDB using ALTER TABLE.
```
so i modify create index with parser from `STRING_VALUE` to `relation_name` according to ob parser yacc in OBParser.g4:
image

Fixes #

```docs

```


MarkPotato777 (approved):
LGTM

PR #4161, opened 2025-01-13
fix(migrate): vectorydb only load on server mode

ungreat:
#### What type of PR is this?
type-bug

VectorDBMigrate only load on server mode.

Fixes #

```docs

```


MarkPotato777 (approved):
LGTM

PR #4165, opened 2025-01-14
feat(dlm): response the comments on #4151

guowl3:
#### What type of PR is this?
type-feature

Response the comments on #4151 

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionTesting.java
```diff
@@ -104,7 +104,7 @@ public ConnectionTestResult test(@NotNull @Valid TestConnectionReq req) {
 
     public ConnectionTestResult test(@NonNull ConnectionConfig config) {
         ConnectType type = config.getType();
-        if (type.getDialectType() == DialectType.FILE_SYSTEM) {
+        if (type != null && type.isFileSystem()) {
```
LioRoger, line 107: this code is not union with following change, if you can't determinate if type is null, use common function to wrap this judge, union the call entry.

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/DefaultConnectionAdapter.java
```diff
@@ -36,6 +38,13 @@ public class DefaultConnectionAdapter implements ConnectionAdapter {
      */
     public <T extends CloudConnectionConfig> T adaptConfig(T connectionConfig) {
         PreConditions.notNull(connectionConfig, "connectionConfig");
+        if (connectionConfig instanceof ConnectionConfig
```
LioRoger, line 41: type may be a common attribute for Connection config

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionTesting.java
guowl3: removed this validation now because type already has a not null validation in ConnectionConfig

LioRoger (approved):
lgtm

MarkPotato777 (approved):
LGTM

PR #4166, opened 2025-01-14
fix(session): killing sessions failed in OSC tasks

MarkPotato777:
#### What type of PR is this?

type-bug
Kill sessions failed in OSC tasks. The root cause is the kill SQL is wrongly set to `kill query session_id`, which should be `kill session_id`

Fixes #

```docs

```


LioRoger (approved):
lgtm

PR #4169, opened 2025-01-15
fix(session): killing session may fail due to a wrong regex

MarkPotato777:
#### What type of PR is this?

type-bug
The previous regex to extract observer IP and port is wrong. This regular expression has two issues:
1. It allows any characters before and after, which can cause logical errors. For example, with an input like 1.1.1.1:8888, the IP will be captured as 1.1 instead of 1.1.1.1 because .* is a greedy pattern that tries to match as many characters as possible. 
3. The dot (.) is not escaped, which means it matches any character in regex, but it should actually represent the dot (.) itself.

Fixes #

```docs

```

### DIFF
--- a/server/odc-common/src/main/java/com/oceanbase/odc/common/util/HostUtils.java
+++ b/server/odc-common/src/main/java/com/oceanbase/odc/common/util/HostUtils.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright (c) 2023 OceanBase.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


server/odc-common/src/main/java/com/oceanbase/odc/common/util/HostUtils.java
```diff
@@ -0,0 +1,71 @@
...
+
+
+    @NotNull
+    public static ServerAddress extractServerAddress(String ipAndPort) {
+        String trimmed = StringUtils.trim(ipAndPort);
+        if (StringUtils.isBlank(trimmed)) {
+            log.info("unable to extract server address, text is empty");
+            throw new IllegalArgumentException("Empty server address!");
+        }
+        Matcher matcher = SERVER_PATTERN.matcher(trimmed);
```
LioRoger, line 45: use split  to distract ip and port is more accurate maybe

MarkPotato777: done

guowl3 (approved):

ungreat (approved):

PeachThinking (approved):
lgtm

PR #4170, opened 2025-01-15
feat(sql-check):support part of modify data type online ddl check

LioRoger:
#### What type of PR is this?

type-feature
if OBMySQL version greater than 4.0.0, check if data type change statement is online ddl
1. modify （add/remove/change）default value
2. change null flag null / not null
3. increase precision of char/varchar/text/number
collate/charset modify is offline ddl for now.

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLOfflineDdlExists.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLOfflineDdlExists.java
@@ -16,18 +16,25 @@
 package com.oceanbase.odc.service.sqlcheck.rule;
 
 import java.io.StringReader;
+import java.math.BigDecimal;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Function;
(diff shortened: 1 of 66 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/MySQLOfflineDdlExists.java
```diff
@@ -112,18 +163,233 @@ protected List<CheckViolation> addColumnInLocation(Statement statement, AlterTab
         });
     }
 
-    protected List<CheckViolation> changeColumnDataType(Statement statement, CreateTable target,
+    protected List<CheckViolation> changeColumnDataType(String dbVersion, Statement statement, CreateTable target,
             AlterTableAction action) {
+        boolean isOb4x = VersionUtils.isGreaterThan(dbVersion, "4.0.0");
```
PeachThinking, line 168: `getDBVersion()` may return null and `VersionUtils.isGreaterThan` will throw exception

LioRoger: good advice, i have correct the version judge logic to StringUtils.isNotEmpty(dbVersion) and VersionUtils.isGreaterThan(dbVersion, "4.0.0").

guowl3 (approved):
lgtm

zijiacj (approved):
lgtm

PR #4171, opened 2025-01-15
fix(session): get wrong server port 

ungreat:
#### What type of PR is this?
type-bug

DBSession's svr_ip should use sql_port,  instead of svr_port. svr_port is the port of the rpc，while sql_port for sql connection.

Fixes #

```docs

```


LioRoger (approved):
lgtm
but for this case, any integrate test covered this code will meet the problem. so please add test to guarantee your code work correct, special for this common package, cause you don't know who will use this code and what problem you will cause

MarkPotato777 (approved):
LGTM

PR #4174, opened 2025-01-16
fix(session): DBSession does not involve svrIp when OB version less than 4.0

MarkPotato777:
#### What type of PR is this?

type-bug
When the observer version is less than 4.0, there is no `oceanbase.gv$full_processlist`, so it lacks the observer Ip & port as the `DBSession.svrIp`. I added the info by executing `show full processlist`.

Fixes #

```docs

```


libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/stats/DBStatsAccessorFactory.java
```diff
@@ -70,7 +70,7 @@ public DBStatsAccessor buildForOBMySQL() {
             // OB 版本 >= 4.0.0
             return new OBMySQLNoLessThan400StatsAccessor(getJdbcOperations());
         } else {
-            return new OBMySQLStatsAccessor(getJdbcOperations());
+            return new OBMySQLLessThan400StatsAccessor(getJdbcOperations());
```
PeachThinking, line 73: why OBMySQLStatsAccessor is not used？

libs/db-browser/src/test/java/com/oceanbase/tools/dbbrowser/env/BaseTestEnv.java
```diff
@@ -88,7 +88,7 @@ public abstract class BaseTestEnv extends BasePropertiesEnv {
             log.info("Clear OB Oracle user succeed, user={}", TEST_OB_ORACLE_DATABASE_NAME);
             clear(mysqlParseResult, MYSQL_DS_KEY);
             log.info("Clear MySQL database succeed, database={}", TEST_MYSQL_DATABASE_NAME);
-            clear(oracleConfig);
+            // clear(oracleConfig);
```
PeachThinking, line 91: remove "//"

libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/stats/DBStatsAccessorFactory.java
MarkPotato777: done

libs/db-browser/src/test/java/com/oceanbase/tools/dbbrowser/env/BaseTestEnv.java
MarkPotato777: done

PeachThinking (approved):
lgtm

PR #4175, opened 2025-01-16
fix(session): kill console query may failed cause session occupied

ungreat:
#### What type of PR is this?
type-bug

DBStatsAccessors use console session, may lead  to session occupied. Now use new ConnectionSession to list all session.

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/db/session/DefaultDBSessionManage.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/db/session/DefaultDBSessionManage.java
@@ -209,7 +209,7 @@ public class DefaultDBSessionManage implements DBSessionManageFacade {
         ConnectionSession creteConnectionSession = null;
         try {
             creteConnectionSession = factory.generateSession();
-            return DBStatsAccessors.create(connectionSession)
+            return DBStatsAccessors.create(creteConnectionSession)
                     .listAllSessions()
                     .stream()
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/db/session/DefaultDBSessionManage.java
```diff
@@ -200,12 +202,25 @@ private List<KillResult> doKill(ConnectionSession session, Map<String, String> c
     }
 
     private List<OdcDBSession> getSessionList(ConnectionSession connectionSession, Predicate<OdcDBSession> filter) {
-        return DBStatsAccessors.create(connectionSession)
-                .listAllSessions()
-                .stream()
-                .map(OdcDBSession::from)
-                .filter(filter == null ? a -> true : filter)
-                .collect(Collectors.toList());
+        CreateSessionReq keyFromId = new DefaultConnectSessionIdGenerator().getKeyFromId(connectionSession.getId());
```
LioRoger, line 205: in my opinion, the core point is if connectionSession has been in use,  if for kill session in sql console, the datasource may be in use, but other situation, user has new a connection for it, it must not be in use.

MarkPotato777: 👍🏻 done

LioRoger (approved):
lgtm

PR #4176, opened 2025-01-16
fix(session): add retry logic when get session list

MarkPotato777:
#### What type of PR is this?

type-bug
The `oceanbase.gv$ob_processlist` is not in-time after kill sessions. So I added a retry logic when checking the session list.

Fixes #

```docs

```


LioRoger (approved):
lgtm

PR #4179, opened 2025-01-17
fix(migrate): 4331 migrate script may failed on ob1479

ungreat:
#### What type of PR is this?
type-bug

modify column may not effect on ob1479

Fixes #

```docs

```


MarkPotato777 (approved):
LGTM

PR #4186, opened 2025-01-20
fix(migrate): upgrade to 4.3.3 will fail if it exists custom environments before

MarkPotato777:
#### What type of PR is this?

type-bug
It doesn't distinguish between updating and creating scenarios when migrating the SQL rules. 
We should only handle the creating scenarios, and ignore the updating scenarios. If not (as the bug's root cause), it will cause a unique key conflict.

Fixes #

```docs

```


LioRoger (approved):
lgtm

PR #4187, opened 2025-01-20
fix(sql-check): it may occur NPE when doing SQL check

MarkPotato777:
#### What type of PR is this?

type-bug
In the SQL check rule 'TooLongCharLength', it may occur NPE as the dataType is nullable, but the caller does not consider it.

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/TooLongCharLength.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/TooLongCharLength.java
@@ -19,6 +19,7 @@ import java.math.BigDecimal;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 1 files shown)


zijiacj (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/rule/TooLongCharLength.java
```diff
@@ -74,19 +75,20 @@ public List<DialectType> getSupportsDialectTypes() {
...
-                return false;
-            }
-            CharacterType type = (CharacterType) d;
-            return type.getLength().compareTo(new BigDecimal(maxCharLength)) > 0;
-        }).map(d -> {
-            CharacterType type = (CharacterType) d;
-            return SqlCheckUtil.buildViolation(sql, d, getType(),
-                    new Object[] {maxCharLength, type.getLength().toString()});
-        }).collect(Collectors.toList());
+        return stream.filter(d -> Objects.nonNull(d) && Objects.nonNull(d.getDataType()))
```
LioRoger, line 78: which situation that d is null or d.getDataType is null
i think the parse give more information about unrecognized ColumnDefinition maybe a better choice

LioRoger (approved):

PR #4188, opened 2025-01-21
fix(schedule): detail schedule in the individual space may fail

MarkPotato777:
#### What type of PR is this?

type-bug
In the individual space, there is no project related to the schedule. This PR adds a judgement if the projected is null, then skill detail the project.

Fixes #

```docs

```


guowl3 (approved):
lgtm

PR #4190, opened 2025-01-21
fix(task): change DefaultResourceOperatorBuilder default resource match type

LioRoger:
…nativeK8sPod

#### What type of PR is this?

type-bug
DefaultResourceOperatorBuilder should use nativeK8sPod type match to let cloud k8s matcher work correct

Fixes #

```docs

```


guowl3 (dismissed):
lgtm

guowl3 (approved):

PR #4196, opened 2025-01-23
fix(changelog): the status does not progress to 'Failed' when a change error occurs

guowl3:
#### What type of PR is this?
type-bug

Catch exception when a change error occurs and update status to failed.

Fixes #

```docs

```


LioRoger (approved):
lgtm

PR #4197, opened 2025-01-23
fix(connection): failed to connect to ob sharding mysql

PeachThinking:
#### What type of PR is this?
type-bug

We do not verify whether the `ConnectType`  selected by the user is consistent with the actual `ConnectType`, because for ob sharding type, it is not possible to obtain it‘s real `ConnectType` through the `ConnectTypeUtil#getDialectType` method.

Fixes #

```docs

```


guowl3 (dismissed):

guowl3 (approved):

LioRoger (approved):
lgtm

PR #4198, opened 2025-01-23
fix(changelog): cannot rollback update changelog status

guowl3:
#### What type of PR is this?
type-bug

Transaction will rollback if update status in executeScheduleChange method, so we need to update status in call method

Fixes #

```docs

```


MarkPotato777 (approved):
LGTM

PR #4199, opened 2025-01-24
security: remove useless sshd-common dependence

ungreat:
#### What type of PR is this?
type-bug

Remove useless sshd-common dependence to avoid some security vulnerabilities.

Fixes #

```docs

```


MarkPotato777 (commented):
LGTM

LioRoger (approved):
maven dependency can help us analysis pom effect or conflict, my question is why this dependency should be removed explicit? is there any other dependency should be excluded or upgraded?

PR #4202, opened 2025-02-05
feat(taskframework): customized task submission

guowl3:
#### What type of PR is this?
type-feature

Customized task submission for different environments

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/submitter/DefaultJobSubmitter.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/submitter/DefaultJobSubmitter.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright (c) 2023 OceanBase.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/submitter/DefaultJobSubmitter.java
```diff
@@ -0,0 +1,90 @@
...
+    @Override
+    public Long submit(String parametersJson, String type, Long timeoutMillis, CloudProvider provider, String region) {
+        TaskDescription taskDescription = TaskDescription.valueOf(type);
+        Map<String, String> jobData = new HashMap<>();
+        jobData.put(JobParametersKeyConstants.META_TASK_PARAMETER_JSON, parametersJson);
+        if (timeoutMillis != null) {
+            jobData.put(JobParametersKeyConstants.TASK_EXECUTION_TIMEOUT_MILLIS, timeoutMillis.toString());
+        }
+        Map<String, String> jobProperties = new HashMap<>();
+        if (provider != null && StringUtils.isNotEmpty(region)) {
```
LioRoger, line 58: if provider is not null, but region is null, program may not work correct, add some check here

guowl3: checked in cloud

PR #4203, opened 2025-02-06
fix(sql-execute): nls_format invalidation

LuckyPickleZZ:
#### What type of PR is this?
type-bug

The NlsFormatInterceptor failed to recognize statement like `set variables='xxx'`.
image

image

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/session/interceptor/NlsFormatInterceptor.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/session/interceptor/NlsFormatInterceptor.java
@@ -195,7 +195,7 @@ public class NlsFormatInterceptor extends BaseTimeConsumingInterceptor {
                             .generate();
                 }
                 name = ctx.sys_var_and_val().obj_access_ref_normal().getText();
-            } else if (ctx.set_expr_or_default() != null ) {
+            } else if (ctx.set_expr_or_default() != null) {
                 if (ctx.set_expr_or_default().bit_expr() != null) {
                     value = new OracleExpressionFactory(ctx.set_expr_or_default().bit_expr()).generate();
                 }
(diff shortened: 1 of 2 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/session/interceptor/NlsFormatInterceptor.java
```diff
@@ -187,15 +187,28 @@ private static class NlsFormatVariableVisitor extends OBParserBaseVisitor<Variab
...
         public VariableAssign visitVar_and_val(Var_and_valContext ctx) {
-            if (ctx.scope_or_scope_alias() == null) {
-                return null;
-            }
             Expression value = null;
-            if (ctx.set_expr_or_default().bit_expr() != null) {
-                value = new OracleExpressionFactory(ctx.set_expr_or_default().bit_expr()).generate();
+            String name;
+            if (ctx.sys_var_and_val() != null) {
+                if (ctx.sys_var_and_val().set_expr_or_default().bit_expr() != null) {
```
LioRoger, line 193: is ctx.sys_var_and_val().set_expr_or_default() must has a value not null?
in other words,  if chained value access should have null check

LuckyPickleZZ: Chain access should indeed undergo non null checks. The reason for skipping the check here is that if it is an SQL statement with 'set xxx=xxx', then 'sys_dar_andnval(). set-expr_or_default()' will definitely have a non empty value. In other cases, NPE does not affect operation. If an exception is caught externally, it can be ignored.

MarkPotato777 (approved):
LGTM

LioRoger (approved):
lgtm

PR #4212, opened 2025-02-07
fix(schedule): unreasonable parameter validation

guowl3:
#### What type of PR is this?
type-bug

1. Allow choosing minutes and seconds as the offset unit
2. Only check permission for OBInstance

Fixes #

```docs

```


kiko-art (approved):
lgtm

PR #4214, opened 2025-02-08
fix(task): process mode not invoke destroy action when job is canceled

LioRoger:
#### What type of PR is this?

type-bug
when task run in process mode, JobCaller.finish not called cause job_job.destroyed_time column is set unexpected,
remove this code to let destroyExecutor daemon task handle remain clean action.

Fixes #

```docs

```


guowl3 (approved):
lgtm

PR #4217, opened 2025-02-08
fix(task): sync report taskResult call

LioRoger:
…port

#### What type of PR is this?

type-bug
report task result in TaskMonitor is not synchronized. that may cause unexpected result save.
this patch sync report call, if doFinal called, the following report will be ignored

Fixes #

```docs

```


guowl3 (approved):
lgtm

PR #4219, opened 2025-02-08
fix(dlm): correct status when task is terminated

guowl3:
#### What type of PR is this?
type-bug

1. correct subtask status when task is terminated

Fixes #

```docs

```


LioRoger (approved):
lgtm

PR #4220, opened 2025-02-10
fix(osc): updateRateLimiterConfig npe issue

LioRoger:
#### What type of PR is this?

type-bug
if rate limiter config not provide in ScheduleEntity.OnlineSchemaChangeParameters, npe will be caused.
OscActionBase access for this variable has handled the issue in OscActionFsmBase.parseOnlineSchemaChangeParameters

Fixes #

```docs

```


LuckyPickleZZ (approved):
Item

PR #4221, opened 2025-02-10
fix(flowable): flowable task may be blocked

ungreat:
#### What type of PR is this?
type-bug

The default waiting queue of the flowTaskExecutor thread pool is the largest, causing the number of threads in the pool to not increase. This PR sets the coreThreadSize and maxThreadSize to be the same and makes this number configurable.

Fixes #

```docs

```


guowl3 (approved):
lgtm

PR #4224, opened 2025-02-11
fix(osc): add uid for oms updateConfigRequest

LioRoger:
#### What type of PR is this?

type-bug
add uid for osc when invoke UpdateOmsConfig open API

Fixes #

```docs

```


LuckyPickleZZ (approved):
lgtm

PR #4225, opened 2025-02-11
fix(dlm): upgrade sdk to 1.2.1

guowl3:
#### What type of PR is this?
type-bug

1. upgrade sdk version to 1.2.1
2. update limit config for all table
3. record partition save point

Fixes #

```docs

```


tonghan-gsh (approved):
lgtm

PR #4230, opened 2025-02-14
fix(notification): fail to generate notification event

LuckyPickleZZ:
#### What type of PR is this?
type-bug

ResourceRoleService#listByResourceIdentifierIn requires organization. And when calling it asynchronously, it would throw an AccessDeniedException.


MarkPotato777 (approved):
LGTM

PR #4231, opened 2025-02-14
build: fail to build artifact

LuckyPickleZZ:
#### What type of PR is this?
type-bug

The github artifact would failed like https://github.com/oceanbase/odc/actions/runs/13306599104/job/37158911752 .
According to https://stackoverflow.com/questions/75652635/a-required-class-was-missing-while-executing-maven-plugin-with-github-action , fixed it.


guowl3 (approved):
lgtm

PR #4232, opened 2025-02-14
feat(database): supports query database by tenantName、clusterNam and dataSourceName

ysjemmm:
#### What type of PR is this?
type-feature

Databases can be queried by tenantName, clusterName and dataSourceName

### DIFF
--- a/server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DataBaseController.java
+++ b/server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DataBaseController.java
@@ -69,8 +69,8 @@ public class DataBaseController {
     @RequestMapping(value = "/databases", method = RequestMethod.GET)
     public PaginatedResponse<Database> listDatabases(
             @RequestParam(required = false, name = "name") String name,
-            @RequestParam(required = false, name = "tenant") String tenantName,
-            @RequestParam(required = false, name = "cluster") String clusterName,
+            @RequestParam(required = false, name = "tenantId") String tenantName,
+            @RequestParam(required = false, name = "clusterId") String clusterName,
             @RequestParam(required = false, name = "type") List<DatabaseType> types,
             @RequestParam(required = false, name = "existed") Boolean existed,
             @RequestParam(required = false, name = "dataSourceName") String dataSourceName,
--- a/server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DataBaseController.java
+++ b/server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DataBaseController.java
@@ -69,8 +69,8 @@ public class DataBaseController {
     @RequestMapping(value = "/databases", method = RequestMethod.GET)
     public PaginatedResponse<Database> listDatabases(
             @RequestParam(required = false, name = "name") String name,
-            @RequestParam(required = false, name = "tenant") String tenantId,
-            @RequestParam(required = false, name = "cluster") String clusterId,
+            @RequestParam(required = false, name = "tenant") String tenantName,
+            @RequestParam(required = false, name = "cluster") String clusterName,
             @RequestParam(required = false, name = "type") List<DatabaseType> types,
             @RequestParam(required = false, name = "existed") Boolean existed,
             @RequestParam(required = false, name = "dataSourceName") String dataSourceName,
@@ -92,8 +92,8 @@ public class DataBaseController {
                 .existed(existed)
                 .environmentId(environmentId)
                 .schemaName(name)
-                .tenantId(tenantId)
-                .clusterId(clusterId)
+                .tenantName(tenantName)
+                .clusterName(clusterName)
                 .dataSourceName(dataSourceName)
                 .containsUnassigned(containsUnassigned)
                 .includesPermittedAction(includesPermittedAction)
--- a/server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DataBaseController.java
+++ b/server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DataBaseController.java
@@ -69,8 +69,8 @@ public class DataBaseController {
     @RequestMapping(value = "/databases", method = RequestMethod.GET)
     public PaginatedResponse<Database> listDatabases(
             @RequestParam(required = false, name = "name") String name,
-            @RequestParam(required = false, name = "tenantId") String tenantName,
-            @RequestParam(required = false, name = "clusterId") String clusterName,
+            @RequestParam(required = false, name = "tenantName") String tenantName,
+            @RequestParam(required = false, name = "clusterName") String clusterName,
             @RequestParam(required = false, name = "type") List<DatabaseType> types,
             @RequestParam(required = false, name = "existed") Boolean existed,
             @RequestParam(required = false, name = "dataSourceName") String dataSourceName,
--- a/server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DataBaseController.java
+++ b/server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DataBaseController.java
@@ -69,6 +69,8 @@ public class DataBaseController {
     @RequestMapping(value = "/databases", method = RequestMethod.GET)
     public PaginatedResponse<Database> listDatabases(
             @RequestParam(required = false, name = "name") String name,
+            @RequestParam(required = false, name = "tenant") String tenantId,
+            @RequestParam(required = false, name = "cluster") String clusterId,
             @RequestParam(required = false, name = "type") List<DatabaseType> types,
             @RequestParam(required = false, name = "existed") Boolean existed,
             @RequestParam(required = false, name = "dataSourceName") String dataSourceName,
@@ -90,6 +92,9 @@ public class DataBaseController {
                 .existed(existed)
                 .environmentId(environmentId)
                 .schemaName(name)
+                .tenantId(tenantId)
+                .clusterId(clusterId)
+                .dataSourceName(dataSourceName)
                 .containsUnassigned(containsUnassigned)
                 .includesPermittedAction(includesPermittedAction)
                 .projectId(projectId).build();
--- a/server/odc-service/src/main/java/com/oceanbase/odc/metadb/connection/DatabaseSpecs.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/metadb/connection/DatabaseSpecs.java
@@ -57,8 +57,7 @@ public class DatabaseSpecs {
 
     public static Specification<DatabaseEntity> connectionIdIn(Collection<Long> connectionIds) {
         Set<Long> finalConnectionIds = connectionIds.stream().filter(Objects::nonNull).collect(Collectors.toSet());
-        return (root, query, builder) -> CollectionUtils.isEmpty(connectionIds) ? builder.conjunction()
-                : root.get("connectionId").in(finalConnectionIds);
+        return SpecificationUtil.columnIn("connectionId", finalConnectionIds);
     }
 
     public static Specification<DatabaseEntity> organizationIdEquals(Long organizationId) {
--- a/server/odc-service/src/main/java/com/oceanbase/odc/metadb/connection/DatabaseSpecs.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/metadb/connection/DatabaseSpecs.java
@@ -19,6 +19,7 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.apache.commons.collections4.CollectionUtils;
 import org.springframework.data.jpa.domain.Specification;
@@ -52,6 +53,12 @@ public class DatabaseSpecs {
     public static Specification<DatabaseEntity> connectionIdEquals(Long connectionId) {
         return (root, query, builder) -> Objects.isNull(connectionId) ? builder.conjunction()
                 : builder.equal(root.get("connectionId"), connectionId);
+    }
+
+    public static Specification<DatabaseEntity> connectionIdIn(Collection<Long> connectionIds) {
+        Set<Long> finalConnectionIds = connectionIds.stream().filter(Objects::nonNull).collect(Collectors.toSet());
+        return (root, query, builder) -> CollectionUtils.isEmpty(connectionIds) ? builder.conjunction()
+                : root.get("connectionId").in(finalConnectionIds);
     }
 
     public static Specification<DatabaseEntity> organizationIdEquals(Long organizationId) {
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
@@ -851,11 +851,20 @@ public class ConnectionService {
         return innerList(params, Pageable.unpaged()).toList();
     }
 
-    public Set<Long> innerGetIdsByNameAndTenantAndCluster(String dataSourceName, String tenantName,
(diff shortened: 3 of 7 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/model/QueryDatabaseParams.java
```diff
@@ -51,4 +51,10 @@ public class QueryDatabaseParams {
     private Boolean containsUnassigned;
 
     private Boolean includesPermittedAction;
+
+    private String clusterId;
```
MarkPotato777, line 55: What does `clusterId` mean? Should it be `clusterName`? So as the `tenantId`.

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
```diff
@@ -135,6 +137,8 @@
 import com.oceanbase.odc.service.task.base.precheck.PreCheckTaskParameters.AuthorizedDatabase;
 import com.oceanbase.tools.dbbrowser.model.DBDatabase;
 
+import cn.hutool.core.collection.CollUtil;
+import cn.hutool.core.util.StrUtil;
```
MarkPotato777, line 141: Use `com.oceanbase.odc.common.util.StringUtils` instead.

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
```diff
@@ -135,6 +137,8 @@
 import com.oceanbase.odc.service.task.base.precheck.PreCheckTaskParameters.AuthorizedDatabase;
 import com.oceanbase.tools.dbbrowser.model.DBDatabase;
 
+import cn.hutool.core.collection.CollUtil;
```
MarkPotato777, line 140: Use `org.apache.commons.collections.CollectionUtils` instead.

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
ysjemmm: ok

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
ysjemmm: ok

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/model/QueryDatabaseParams.java
ysjemmm: yes, it be clusterName and tenantName

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
```diff
@@ -340,9 +344,27 @@ public Page<Database> list(@NonNull QueryDatabaseParams params, @NotNull Pageabl
...
+                    .name(params.getDataSourceName())
+                    .build();
+            orDataSourceIds = connectionService.list(queryConnectionParams, Pageable.unpaged()).getPage().stream().map(
+                    ConnectionConfig::getId).collect(Collectors.toSet());
+        }
         if (Objects.nonNull(params.getDataSourceId())) {
             specs = specs.and(DatabaseSpecs.connectionIdEquals(params.getDataSourceId()));
         }
+        if (CollUtil.isNotEmpty(orDataSourceIds)) {
+            specs = specs.or(DatabaseSpecs.connectionIdIn(orDataSourceIds));
```
MarkPotato777, line 366: Where is this 'or' condition added? Is it added with an 'or' to all previous conditions, or to a specific condition? What does the final SQL generated by Spring Data JPA look like? Can you provide an example?

server/odc-service/src/main/java/com/oceanbase/odc/metadb/connection/DatabaseSpecs.java
```diff
@@ -54,6 +55,12 @@ public static Specification<DatabaseEntity> connectionIdEquals(Long connectionId
                 : builder.equal(root.get("connectionId"), connectionId);
     }
 
+    public static Specification<DatabaseEntity> connectionIdIn(Collection<Long> connectionIds) {
```
MarkPotato777, line 58: You could use `SpecificationUtil` to simplify your code.

ysjemmm: I will try

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
```diff
@@ -340,9 +344,27 @@ public Page<Database> list(@NonNull QueryDatabaseParams params, @NotNull Pageabl
...
+            QueryConnectionParams queryConnectionParams = QueryConnectionParams.builder()
+                    .clusterNames(
+                            StrUtil.isNotBlank(params.getClusterId()) ? Collections.singletonList(params.getClusterId())
+                                    : null)
+                    .tenantNames(
+                            StrUtil.isNotBlank(params.getTenantId()) ? Collections.singletonList(params.getTenantId())
+                                    : null)
+                    .name(params.getDataSourceName())
+                    .build();
+            orDataSourceIds = connectionService.list(queryConnectionParams, Pageable.unpaged()).getPage().stream().map(
```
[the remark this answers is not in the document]
ysjemmm: The user wants to retrieve a list of databases, and I think the user should also have access to the dataSource that the database belongs to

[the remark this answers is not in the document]
ysjemmm: Otherwise, users can access dataSourceName through tenantName, clusterName, and datasourcename

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
ysjemmm: The new dataSourceName, tenantName, and clusterName are added Parameters are used in conjunction with the packet database.
For example:

- The user has the data source `ds1`, the cluster name is `c1`, and underneath it are the databases `c1_da1`, `c1_da2`,

- The user has the data source `ds2`, the cluster name is `c2`, and the following databases are `da`, `da1`,

When I select the front end select "Group by cluster" and enter the search field "2", I get a list of databases as `c1_da2`, `da`, `da1`, because the cluster names that `da` and `da2` belong to can also be vaguely matched by "2"

ysjemmm: > Where is this 'or' condition added? Is it added with an 'or' to all previous conditions, or to a specific condition? What does the final SQL generated by Spring Data JPA look like? Can you provide an example?

for example, group databases by cluster: 
query text = keywords

- query connection_id by clusterName: `select connection_id from connect_connection where cluster_name like %keywords%`, then save in connectionIds
- query databases by name or connectionIds: `select * from connect_database where name like %keywords% or connection_id in connectionIds`

ysjemmm: I will change,  that is `... and ('name' like name or connection_id in (connectionIds)`, it may change to `specs = specs.and(
                 DatabaseSpecs.nameLike(params.getSchemaName()).or(DatabaseSpecs.connectionIdIn(orDataSourceIds)));`

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DataBaseController.java
```diff
@@ -69,6 +69,8 @@ public SuccessResponse<Database> getDatabase(@PathVariable Long id) {
     @RequestMapping(value = "/databases", method = RequestMethod.GET)
     public PaginatedResponse<Database> listDatabases(
             @RequestParam(required = false, name = "name") String name,
+            @RequestParam(required = false, name = "tenant") String tenantName,
```
guowl3, line 72: The "tenant" and "cluster" are ambiguous, there might be a need to support queries using "tenantId" and "clusterId" in the future

ysjemmm: > The "tenant" and "cluster" are ambiguous, there might be a need to support queries using "tenantId" and "clusterId" in the future

done

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
```diff
@@ -845,6 +845,15 @@ public List<ConnectionConfig> listSkipPermissionCheck(@NotNull QueryConnectionPa
         return innerList(params, Pageable.unpaged()).toList();
     }
 
+    public Set<Long> innerGetIdsByNameAndTenantAndCluster(String dataSourceName, String tenantName,
+            String clusterName) {
+        Specification<ConnectionEntity> spec = Specification.where(ConnectionSpecs.nameLike(dataSourceName))
```
LioRoger, line 850: like match is full table match, is there any possible that we use accurate match?

ysjemmm: > like match is full table match, is there any possible that we use accurate match?

it is fuzzy match

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
```diff
@@ -845,6 +845,15 @@ public List<ConnectionConfig> listSkipPermissionCheck(@NotNull QueryConnectionPa
...
+                .and(ConnectionSpecs.tenantNameLike(tenantName))
+                .and(ConnectionSpecs.clusterNameLike(clusterName));
+        return this.repository.findAll(spec, Pageable.unpaged()).stream().map(ConnectionEntity::getId)
```
LioRoger, line 853: this function should handle if any of the argument is not passed correctly, eg tenantName not passed in fact, how can query make expected query

guowl3 (dismissed):

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
```diff
@@ -851,6 +851,24 @@ public List<ConnectionConfig> listSkipPermissionCheck(@NotNull QueryConnectionPa
...
     }
 
+    public Set<Long> innerGetIdsIfAnyOfNameTenantCluster(String dataSourceName, String tenantName,
+            String clusterName) {
+        if (StringUtils.isBlank(dataSourceName) && StringUtils.isBlank(tenantName)
+                && StringUtils.isBlank(clusterName)) {
+            return Collections.emptySet();
+        }
+        Specification<ConnectionEntity> spec;
+        if (StringUtils.isNotBlank(dataSourceName)) {
```
PeachThinking, line 861: There should not be an `if...else if...else` condition here？ for example, if both `dataSourceName` and `tenantName` are not blank, then you only query `dataSourceName`.

ysjemmm: Yes, according to requirements，I only need one of those at most, instead of having an "and" or "or" relationship

ysjemmm: When more than one condition is passed, it is no longer compatible with the function of the method

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
```diff
@@ -851,6 +852,25 @@ public List<ConnectionConfig> listSkipPermissionCheck(@NotNull QueryConnectionPa
         return innerList(params, Pageable.unpaged()).toList();
     }
 
+    public Set<Long> innerGetIdsIfAnyOfNameTenantCluster(String dataSourceName, String tenantName,
+            String clusterName) {
+        long conditionCount = Stream.of(dataSourceName, tenantName, clusterName).filter(StringUtils::isNotBlank)
+                .count();
+        if (conditionCount == 0L || conditionCount >= 2L) {
```
MarkPotato777, line 859: Why return empty set when `conditionCount >= 2`?
It's tricky to use magic numbers 0L and 2L.

ysjemmm: I only need one condition at most, otherwise it doesn't fit the function

MarkPotato777: Then just verify the conditionCount == 1, and throw an exception if not.

MarkPotato777: BTW, this method is not readable. The caller is not aware that you only accept one of the parameters. Think about what you will do when you are required to accept more parameters in future features.

ysjemmm: > BTW, this method is not readable. The caller is not aware that you only accept one of the parameters. Think about what you will do when you are required to accept more parameters in future features.

If there are multiple conditions, all I can think of is using Map, or using `Function`, leaving the decision to the caller

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DataBaseController.java
```diff
@@ -69,6 +69,8 @@ public SuccessResponse<Database> getDatabase(@PathVariable Long id) {
     @RequestMapping(value = "/databases", method = RequestMethod.GET)
     public PaginatedResponse<Database> listDatabases(
             @RequestParam(required = false, name = "name") String name,
+            @RequestParam(required = false, name = "tenantId") String tenantName,
```
MarkPotato777, line 72: why 'name' is 'tenantId' while the parameter is 'tenantName' ?

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/model/InnerQueryConnectionParams.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * @Author: ysj
+ * @Date: 2025/2/28 10:30
+ * @Since: 4.3.4
+ * @Description:
+ */
+@Data
+@Builder
+public class InnerQueryConnectionParams {
```
MarkPotato777, line 29: Why do we need an inner one?

ysjemmm: To avoid function queries passing in too many parameters, we need some special combination of query parameter classes for internal queries, because it is difficult to extend with `QueryConnectionParams`

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DataBaseController.java
ysjemmm: may see https://github.com/oceanbase/odc/pull/4232#discussion_r1957536596

ysjemmm: > I assume that you misunderstood @guowl3 's comments... `tenant` is ambiguous, you should use `tenantName` instead. `tenantId` is just an example...

oo，I know😂

MarkPotato777 (dismissed):
LGTM

PeachThinking (dismissed):
lgtm

guowl3 (approved):

LioRoger (approved):
lgtm

guowl3 (approved):

PeachThinking (approved):

PR #4234, opened 2025-02-18
feat(task): introduce supervisor agent

LioRoger:
#### What type of PR is this?

type-feature
introduce task supervisor component to manager task
1. task will not be distinguished on local or k8s, supervisor endpoint will be the only things that job caller needed to known
2. introduce supervisor agent to do command call, supervisor agent will be bind to a task resource. it's the entity of supervisor endpoint
3. supervisor net command call is shield by local and remote supervisor task proxy.
4. resource can be managed flexible by ResourceManageStrategy
5. reconstruct task daemon driver, let job modified status safe
6. rate limiter for process is redesigned.

Fixes #

```docs

```


server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_1__add_supervisor_agent.sql
```diff
@@ -0,0 +1,18 @@
+--
+-- Add supervisor_endpoint table to maintain supervisor agent info, this table will bind to resource_resource
+--
+CREATE TABLE `supervisor_endpoint` (
```
MarkPotato777, line 4: ODC tends to use the module name as the table name prefix. How about `task_supervisor_endpoint`?

server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_1__add_supervisor_agent.sql
```diff
@@ -0,0 +1,18 @@
...
+    `id` bigint(20) NOT NULL AUTO_INCREMENT,
+    `host` varchar(256)  NOT NULL COMMENT 'host of supervisor',
+    `port` int(11)  NOT NULL COMMENT 'port of supervisor',
+    `status` varchar(64)  NOT NULL COMMENT 'status of supervisor, contains PREPAREING,AVAILABLE,DESTROYED,UNAVAILABLE，ABANDON',
+    `loads` int(11)  NOT NULL COMMENT 'load of supervisor',
```
MarkPotato777, line 9: what do `loads` represent for?

server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_2__add_resource_allocate_info.sql
```diff
@@ -0,0 +1,20 @@
+--
+-- Add resource_allocate_info table to maintain task resource allocate info
+--
+CREATE TABLE IF NOT EXISTS `resource_allocate_info` (
```
MarkPotato777, line 4: How about `task_resource_allocate_info`?

server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_2__add_resource_allocate_info.sql
```diff
@@ -0,0 +1,20 @@
...
+    `id` bigint(20) NOT NULL AUTO_INCREMENT,
+    `task_id` bigint(20) NOT NULL COMMENT 'task id allocate to this allocate info, equals to job_job.id',
+    `resource_allocate_state` varchar(20) NOT NULL COMMENT 'resource allocate state, update by resource allocator, including PREPARING, AVAILABLE, FAILED, FINISHED',
+    `resource_usage_state` varchar(20) NOT NULL COMMENT 'resource usage state update by resource user, including PREPARING, USING, FINISHED',
+    `endpoint` varchar(512) DEFAULT NULL COMMENT 'supervisor endpoint, in format host:port',
```
MarkPotato777, line 9: why do we need `endpoint` as we already have `supervisor_endpoint_id`? Is it for duplication and convenience? If yes, how about naming it to `supervisor_endpoint_address` for clearness?

server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_2__add_resource_allocate_info.sql
```diff
@@ -0,0 +1,20 @@
...
+    `supervisor_endpoint_id` bigint(20) DEFAULT NULL  COMMENT 'supervisor endpoint id related to supervisor_endpoint.id',
+    `resource_region` varchar(128) NOT NULL COMMENT 'resource region to filter endpoint',
+    `resource_group` varchar(128) NOT NULL COMMENT 'resource group to filter endpoint',
+    `resource_applier_name`  varchar(128) NOT NULL COMMENT 'resource applier name, eg PROCESS/K8S',
+    `create_time` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP COMMENT 'create time',
+    `update_time` datetime DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP COMMENT 'update time',
+    PRIMARY KEY (`id`),
+    UNIQUE KEY `task_index` (`task_id`),
```
MarkPotato777, line 17: ODC tends to use `idx_${table_name}_${column_names}` as the index name.

server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_1__add_supervisor_agent.sql
LioRoger: good suggest, i will correct name with task prefix

server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_1__add_supervisor_agent.sql
LioRoger: currently one task takes one load, so loads means how many task has run on this supervisor.
In future  release different task may have different load

server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_2__add_resource_allocate_info.sql
LioRoger: introduce this field is to divide reference between user and task resource mode, if we only provide resource_id, resource user will have feel the exists of supervisor endpoint table, that's what i intend to avoided.
rename endpoint to supervisor_endpoint is more clear

server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_2__add_resource_allocate_info.sql
LioRoger: ok, i will correct index name to following the name rule

server/odc-server/src/main/java/com/oceanbase/odc/supervisor/SupervisorAgent.java
```diff
@@ -0,0 +1,88 @@
...
+ * @author longpeng.zlp
+ * @date 2024/11/26 15:43
+ */
+@Slf4j
+@Evolving
+public class SupervisorAgent {
+    public static void main(String[] args) {
+        BootAgentUtil.setLogPathSysProperty();
+        // set log4j xml
+        BootAgentUtil.setLog4JConfigXml(SupervisorAgent.class.getClassLoader(), "log4j2-supervisor.xml");
```
guowl3, line 40: After the task is completed, has the log been uploaded to OSS? Similarly, what about the executor logs?

server/odc-service/src/main/java/com/oceanbase/odc/service/dlm/DLMJobStore.java
```diff
@@ -69,9 +69,11 @@ public void destroy() {
             return;
         }
         try {
-            dataSource.close();
+            if (null != dataSource) {
```
guowl3, line 72: non-null validation is already in place

server/odc-service/src/main/java/com/oceanbase/odc/service/dlm/DLMJobStore.java
```diff
@@ -69,9 +69,11 @@ public void destroy() {
...
+                dataSource.close();
+            }
         } catch (Exception e) {
-            log.warn("Close meta datasource failed,errorMsg={}", e.getMessage());
+            log.warn("Close meta datasource failed,errorMsg={}", e);
```
guowl3, line 76: maybe the error message is already sufficient

server/odc-service/src/main/java/com/oceanbase/odc/service/resource/ResourceManager.java
```diff
@@ -80,7 +82,7 @@ public void registerResourceOperator(ResourceOperatorBuilder<?, ?> operator) {
      * @return
      */
     @SuppressWarnings("all")
-    @Transactional
+    @Transactional(propagation = Propagation.REQUIRES_NEW)
```
guowl3, line 85: Do these methods not need to consider the behavior of the caller?

server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduledTaskLoggerService.java
```diff
@@ -320,4 +321,28 @@ private CloudObjectStorageService getCloudObjectStorageService(JobEntity jobEnti
         return cloudObjectServiceMap.computeIfAbsent(objectStorageConfiguration,
                 k -> CloudObjectStorageServiceBuilder.build(objectStorageConfiguration));
     }
+
+    public String getLogContent(@NonNull String executorEndpoint, @NonNull Long jobId, @NonNull OdcTaskLogLevel level) {
```
guowl3, line 325: can this method be extended to query task logs, execution logs, and supervision logs?

server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/job/SqlPlanJob.java
```diff
@@ -166,6 +166,9 @@ private void executeInTaskFramework(JobExecutionContext context) {
             JobPropertiesUtils.setCloudProvider(jobProperties,
                     CloudProvider.fromValue(attributes.get("cloudProvider").toString()));
             JobPropertiesUtils.setRegionName(jobProperties, attributes.get("region").toString());
+        } else {
+            JobPropertiesUtils.setDefaultCloudProvider(jobProperties);
```
guowl3, line 170: Inheriting from AbstractOdcJob and calling the submit method is more generic

server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_2__add_resource_allocate_info.sql
LioRoger: same as above

server/odc-service/src/main/java/com/oceanbase/odc/service/dlm/DLMJobStore.java
LioRoger: that's maybe the code merge from 4.3.3 make this logic, has removed

server/odc-service/src/main/java/com/oceanbase/odc/service/dlm/DLMJobStore.java
LioRoger: e may be a wrapped message, real cause should be printed

server/odc-service/src/main/java/com/oceanbase/odc/metadb/task/JobRepository.java
```diff
@@ -45,6 +45,14 @@ public interface JobRepository extends JpaRepository<JobEntity, Long>,
     int updateExecutorEndpoint(@Param("id") Long id, @Param("executorEndpoint") String executorEndpoint,
             @Param("oldStatus") JobStatus oldStatus);
 
+    @Transactional
+    @Query(value = "update job_job set "
+            + " executor_endpoint=:executorEndpoint,  executor_identifier = :executorIdentifier"
```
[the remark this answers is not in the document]
LioRoger: for pull mode, endpoint should be updated with executor identifier, previous logic not handle this column

server/odc-service/src/main/java/com/oceanbase/odc/service/schedule/ScheduledTaskLoggerService.java
LioRoger: this is a method move, we can extend it in future release

server/odc-server/src/main/java/com/oceanbase/odc/supervisor/SupervisorAgent.java
LioRoger: this has  record as a new issue to handle

server/odc-service/src/main/java/com/oceanbase/odc/service/resource/ResourceManager.java
LioRoger: this method is idempotent， so no matter how many times it is called, the result will not change

server/odc-service/src/main/java/com/oceanbase/odc/service/task/config/DefaultTaskFrameworkProperties.java
```diff
@@ -96,8 +96,40 @@ public class DefaultTaskFrameworkProperties implements TaskFrameworkProperties {
     private String destroyExecutorJobCronExpression;
 
     private String pullTaskResultJobCronExpression;
+
+    // start and manage resource should invoke frequent to reduce job schedule time
+    private String startPreparingJobV2CronExpression = "0/1 * * * * ?";
```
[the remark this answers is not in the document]
LioRoger: has add init sql to data.sql

guowl3 (approved):
lgtm

MarkPotato777 (approved):
LGTM

LuckyPickleZZ (approved):

PR #4235, opened 2025-02-18
feat(osc): support drop column sql

LioRoger:
#### What type of PR is this?

type-feature

drop column not blocked any more, but drop column and add column can't do in same time

Fixes #

```docs

```


LuckyPickleZZ (approved):
lgtm

PR #4236, opened 2025-02-20
feat(database): Sync current user database objects when grouping databases

ysjemmm:
#### What type of PR is this?
type-feature

At present, the section that groups databases in the resource tree will obtain all databases that the user has the right to operate. In this case, all data objects under the project that the user has joined need to be synchronized

### DIFF
--- a/server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DBObjectIndexController.java
+++ b/server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DBObjectIndexController.java
@@ -70,7 +70,7 @@ public class DBObjectIndexController {
 
     @ApiOperation(value = "syncCurrentUserVisibleDatabases",
             notes = "Sync all databases visible to the current user")
-    @RequestMapping(value = "/sync/currentUser", method = RequestMethod.POST)
+    @RequestMapping(value = "/syncAll", method = RequestMethod.POST)
     public SuccessResponse<Boolean> syncCurrentUserVisibleDatabases() {
         return Responses.success(dbSchemaIndexService.syncCurrentUserVisibleDatabases());
     }
--- a/server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DBObjectIndexController.java
+++ b/server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DBObjectIndexController.java
@@ -68,5 +68,12 @@ public class DBObjectIndexController {
         return Responses.success(dbSchemaIndexService.syncDatabaseObjects(req));
     }
 
+    @ApiOperation(value = "syncCurrentUserDatabaseObjectsWithProject",
+            notes = "Sync all database objects under the project that the current user is joining")
+    @RequestMapping(value = "/syncAllWithProject", method = RequestMethod.POST)
+    public SuccessResponse<Boolean> syncCurrentUserDatabaseObjectsWithProject() {
+        return Responses.success(dbSchemaIndexService.syncCurrentUserDatabaseObjectsWithProject());
+    }
+
 
 }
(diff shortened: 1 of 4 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/connection/database/DatabaseService.java
```diff
@@ -459,6 +459,17 @@ public Set<Database> listExistDatabasesByProjectId(@NonNull Long projectId) {
                 .map(databaseMapper::entityToModel).collect(Collectors.toSet());
     }
 
+    @SkipAuthorize("internal usage")
+    public Set<Database> listExistAndNotPendingDatabasesByProjectIdIn(@NonNull Collection<Long> projectIds) {
+        if (CollectionUtils.isEmpty(projectIds)) {
+            return Collections.emptySet();
+        }
+        return databaseRepository
+                .findByProjectIdInAndExistedAndObjectSyncStatusNot(projectIds, true, DBObjectSyncStatus.PENDING)
```
MarkPotato777, line 468: Why exclude databases with a sync status of PENDING?

ysjemmm: `PENDING` indicates that data synchronization is pending. To prevent repeated synchronization, see `DBSchemaIndexService#syncDatabaseObjects()`.
image

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DBObjectIndexController.java
```diff
@@ -68,5 +68,12 @@ public SuccessResponse<Boolean> syncDatabaseObjects(@RequestBody SyncDBObjectReq
         return Responses.success(dbSchemaIndexService.syncDatabaseObjects(req));
     }
 
+    @ApiOperation(value = "syncCurrentUserDatabaseObjectsWithProject",
+            notes = "Sync all database objects under the project that the current user is joining")
+    @RequestMapping(value = "/syncAllWithProject", method = RequestMethod.POST)
+    public SuccessResponse<Boolean> syncCurrentUserDatabaseObjectsWithProject() {
```
MarkPotato777, line 74: Why not put this API in `DBObjectIndexController` as the `syncDatabaseObjects` did?

MarkPotato777: I think it's not about the 'project', but 'authorized'.

ysjemmm: > Why not put this API in `DBObjectIndexController` as the `syncDatabaseObjects` did?

If you use the `SyncDBObjectReq` parameter passed in `syncDatabaseObjects`, it is difficult to extend support for synchronizing all data objects connected by the user

ysjemmm: > I think it's not about the 'project', but 'authorized'.

Synchronized idea with PD: Synchronize all data objects that the current user has joined the project

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/DBObjectIndexController.java
```diff
@@ -68,5 +68,12 @@ public SuccessResponse<Boolean> syncDatabaseObjects(@RequestBody SyncDBObjectReq
         return Responses.success(dbSchemaIndexService.syncDatabaseObjects(req));
     }
 
+    @ApiOperation(value = "syncCurrentUserVisibleDatabases",
+            notes = "Sync all databases visible to the current user")
+    @RequestMapping(value = "/sync/currentUser", method = RequestMethod.POST)
```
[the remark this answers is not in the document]
ysjemmm: > it's not a RESTful API. `/currentUser` is meaningless, because all the API is related to the current user. In addition, 'currentUser' is not a resource. How about `/syncAll` ? User action(here is `syncAll`) + POST to represent a special API.

Is api/v2/database/object/syncAll？I think ok!

MarkPotato777 (approved):
LGTM

guowl3 (approved):
lgtm

PR #4237, opened 2025-02-20
fix(cloud):failed to deserialize OBInstanceType

PeachThinking:
#### What type of PR is this?
type-bug

fix failed to deserialize OBInstanceType when value is "CLUSTER", because the value of OBInstanceType.CLUSTER is "cluster", so i change value to lowercase uniformly when deserializing.

Fixes #

```docs

```


guowl3 (dismissed):

MarkPotato777 (approved):
LGTM

LioRoger (approved):

PR #4238, opened 2025-02-20
feat(configarchive): support acvhive shchedule config

ungreat:
#### What type of PR is this?
type-feature

0、 add config export and extractor base on jackson
1、support export shchedule config

Fixes #

```docs

```


LioRoger (commented):
first, archive is not suitable for this module
second, when export or import data, transactional should be considered, especial for import task process

ungreat:
> first, archive is not suitable for this module second, when export or import data, transactional should be considered, especial for import task process

Constructive comment！
Rename module from achive to export.
Add rowid in shcheduleRowData for idempotent operations In import.

server/odc-common/src/main/java/com/oceanbase/odc/common/util/FileZipper.java
```diff
@@ -0,0 +1,56 @@
...
+                    int length;
+                    while ((length = fis.read(buffer)) >= 0) {
+                        zos.write(buffer, 0, length);
+                    }
+                    zos.closeEntry();
+                }
+            }
+        }
+
+        for (File file : files) {
```
LioRoger, line 52: one function should have other side effect, in this case, delete file is an side effect. Delete operation should be an explicit call for method caller

server/odc-service/src/main/java/com/oceanbase/odc/service/common/util/OdcFileUtil.java
```diff
@@ -181,15 +181,15 @@ public static String getAbsolutePath() {
         return absolutePath;
     }
 
+    public static void setAbsolutePath(String absolutePath) {
```
LioRoger, line 184: as for util, it's not good design to use a static variable.
 first it's not thread safe, 
second, this hide the factor that should be known by the caller

server/odc-service/src/main/java/com/oceanbase/odc/service/common/util/OdcFileUtil.java
```diff
@@ -277,4 +277,19 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IO
             out.flush();
         }
     }
+
+    public static boolean createFileWithDirectories(File file) {
+        File parentDirectory = file.getParentFile();
```
LioRoger, line 282: this logic is massive.
the parent will not exists if file not exists, 
if you want make dir, just divide mkdir and create file into different functions

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/streamprovider/RemoteUriStreamProvider.java
```diff
@@ -0,0 +1,44 @@
...
+    private final String remoteUrl;
+
+    public RemoteUriStreamProvider(String remoteUrl) {
+        this.remoteUrl = remoteUrl;
+    }
+
+    @Override
+    public InputStream getInputStream() throws Exception {
+        URL url = new URL(remoteUrl);
+        HttpURLConnection connection = (HttpURLConnection) url.openConnection();
```
LioRoger, line 33: why class cast is needed here. if you use http connection to retrieve a input stream, http attribute should be considered, eg socket timeout,  heartbeat timeout, that may cause stream read error if data fetch in stream mode.

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/model/ExportedFile.java
```diff
@@ -0,0 +1,91 @@
...
+    /**
+     * AES secret， null means decrypt data
+     */
+    @Exclude
+    private String secret;
+
+    private boolean checkConfigJsonSignature;
+
+    @JsonIgnore
+    private transient ExportedDataStreamProvider provider;
```
LioRoger, line 62: this variable should be not part of ExportedFile, it's how to read it

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/model/ExportedFile.java
```diff
@@ -0,0 +1,91 @@
...
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+import lombok.ToString.Exclude;
+
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public class ExportedFile {
```
LioRoger, line 36: ExportedFile is wired for me, but i don't have better advice.
In my opinion, the core point for this file is format.
so, this class is more like a context for export file, it provide where to store the content,  how to encrypt it.

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/model/ExportedFile.java
```diff
@@ -0,0 +1,91 @@
...
+        exportedFile.setCheckConfigJsonSignature(secret != null);
+        exportedFile.setUri(file.getPath());
+        ExportedDataStreamProvider provider = new FileStreamProvider(file);
+        exportedFile.setProvider(provider);
+        exportedFile.setSecret(secret);
+        return exportedFile;
+    }
+
+    public File toFile() {
+        Verify.equals(fileType, LOCAL_FILE, "file type not supported");
```
LioRoger, line 88: this method call should not provide here

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/model/ExportedZipFileFactory.java
```diff
@@ -0,0 +1,78 @@
...
+import java.util.Map;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipOutputStream;
+
+import javax.annotation.Nullable;
+
+import com.oceanbase.odc.core.shared.Verify;
+import com.oceanbase.odc.service.common.util.OdcFileUtil;
+
+public class ExportedZipFileFactory {
```
LioRoger, line 32: this class has nothing about export, so that's simply a ZipFile util, and it's not a factory

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/model/ExportedZipFileFactory.java
```diff
@@ -0,0 +1,78 @@
...
+
+    private final File configFile;
+
+    private final Map<String, InputStream> additionFiles;
+
+    public ExportedZipFileFactory(File configFile, Map<String, InputStream> additionFiles) {
```
LioRoger, line 38: It's looks like only provide local file add input, then you can change Map values define InputStream as File
and by the way, if you define InputStream, there is a question, if this class should close InputStream

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/model/ExportedZipFileFactory.java
```diff
@@ -0,0 +1,78 @@
...
+        this.configFile = configFile;
+        this.additionFiles = additionFiles;
+    }
+
+    public ExportedFile build(String outputZipFileName, @Nullable String secret) throws IOException {
+        Verify.notNull(configFile, "configFile");
```
LioRoger, line 44: verify should move to constructor check

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/model/ExportedZipFileFactory.java
```diff
@@ -0,0 +1,78 @@
...
+        try (FileOutputStream fos = new FileOutputStream(outputZipFileName);
+                ZipOutputStream zos = new ZipOutputStream(fos)) {
+            try (InputStream configInputStream = Files.newInputStream(configFile.toPath())) {
+                addFileToZip("config.json", configInputStream, zos);
+                OdcFileUtil.deleteFiles(configFile);
```
LioRoger, line 49: it should be delete by who created it,

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/model/ExportProperties.java
```diff
@@ -0,0 +1,120 @@
...
+import com.oceanbase.odc.core.shared.Verify;
+import com.oceanbase.odc.service.common.util.SpringContextUtil;
+
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Ensure that the keys are in the same order before and after serialization, so use linkedHashMap
+ */
+@Slf4j
+public class ExportProperties extends LinkedHashMap<String, Object> {
```
LioRoger, line 44: why not use LinkedHashMap directly and use a Util to get expected values

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/model/ExportConstants.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public static final String SCHEDULE_ARCHIVE_TYPE = "scheduleArchive";
+
+    /**
+     * filePath, include filename and filetype
+     */
+    public static final String FILE_NAME = "fileName";
+    public static final String FILE_PATH = "filePath";
+
+    public static final String FILE_ZIP_SUFFER = ".zip";
```
LioRoger, line 34: suffix ?

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/model/Encryptable.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.oceanbase.odc.service.exporter.model;
+
+public interface Encryptable {
```
LioRoger, line 18: if decrypt method provided, Encryptable is not suitable for this class

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/Exporter.java
```diff
@@ -0,0 +1,35 @@
...
+
+import java.io.IOException;
+
+import com.oceanbase.odc.service.exporter.model.ExportProperties;
+import com.oceanbase.odc.service.exporter.model.ExportRowDataAppender;
+import com.oceanbase.odc.service.exporter.model.ExportedFile;
+
+public interface Exporter {
+
+    ExportedFile archiveFullData(Object data, ExportProperties metaData, String encryptKey) throws Exception;
```
LioRoger, line 26: Full is not necessary for this method name, cause you don't provide part archieve

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/Exporter.java
```diff
@@ -0,0 +1,35 @@
...
+
+    ExportedFile archiveFullData(Object data, ExportProperties metaData) throws Exception;
+
+    ExportRowDataAppender buildRowDataAppender(ExportProperties metaData,
```
LioRoger, line 30: this name is not necessary for interface

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/Extractor.java
```diff
@@ -0,0 +1,35 @@
...
+import java.io.Closeable;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.oceanbase.odc.service.exporter.model.ExportRowDataReader;
+import com.oceanbase.odc.service.exporter.model.ExportedData;
+import com.oceanbase.odc.service.exporter.model.ExportedFile;
+
+public interface Extractor<R> extends Closeable {
+
+    boolean checkSignature();
```
LioRoger, line 27: interface should only provide what you want other seen, not what you implemented

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/impl/JsonExtractor.java
```diff
@@ -0,0 +1,342 @@
...
+    @Getter
+    private ExportedFile exportedFile;
+
+    private JsonExtractor() {}
+
+    @SneakyThrows
+    public static JsonExtractor buildJsonExtractor(ExportedFile exportedFile, String tempPath) {
+        JsonExtractor jsonExtractor = new JsonExtractor();
+        // Create a random directory within the specified destination path
+        Path randomDir = Files.createTempDirectory(new File(tempPath).toPath(), "unzipped-");
```
LioRoger, line 73: as for random, i think this method should create unique name for path,  and double check maybe needed

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/impl/JsonExtractor.java
```diff
@@ -0,0 +1,342 @@
...
+        try (FileOutputStream fos = new FileOutputStream(tempZipFile);
+                InputStream inputStream = exportedFile.getProvider().getInputStream()) {
+            byte[] buffer = new byte[1024];
+            int len;
+            while ((len = inputStream.read(buffer)) != -1) {
+                fos.write(buffer, 0, len);
+            }
+        }
+        // Unzip the temporary file into the random directory
+        try (ZipInputStream zis = new ZipInputStream(Files.newInputStream(tempZipFile.toPath()))) {
```
LioRoger, line 90: this is how zip util should provided

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/impl/JsonExtractor.java
```diff
@@ -0,0 +1,342 @@
...
+        try (BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream, StandardCharsets.UTF_8))) {
+            while ((bytesRead = reader.read(buffer, 0, buffer.length)) != -1) {
+                stringBuilder.append(buffer, 0, bytesRead);
+            }
+        }
+
+        return stringBuilder.toString();
+    }
+
+    public static class JsonRowDataReader implements ExportRowDataReader<JsonNode> {
```
LioRoger, line 283: mark as private static final

server/odc-service/src/main/java/com/oceanbase/odc/service/common/util/OdcFileUtil.java
ungreat: This method was not added by me, nor was it used, so it can be deleted.

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/model/ExportProperties.java
ungreat: cause some proerties juse in runtime, not necessary add to metadata in final file

server/odc-common/src/main/java/com/oceanbase/odc/common/util/FileZipper.java
ungreat: Perhaps it is because mergeTo does not reflect the semantics of deleting the original file, seprate method instance

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/streamprovider/RemoteUriStreamProvider.java
ungreat: This was designed in advance and has now been removed.

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/model/ExportedFile.java
ungreat: The class was redefined to handle it based on the local File, and only the description information was added

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/model/ExportedZipFileFactory.java
ungreat: may builder is more suitable

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/model/ExportedZipFileFactory.java
ungreat: Indeed, it should not be used to directly access inputStream; instead, use file.👍🏻

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/model/ExportConstants.java
ungreat: may extendtion is more suitable

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/Extractor.java
ungreat: checkSignature in this interface can make sence

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/impl/JsonExtractor.java
ungreat: Files.createTempDirectory could ensure create  a unique path

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/impl/JsonExtractor.java
ungreat: move to utils

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/Exporter.java
ungreat: Stream reading is necessary. Delete the full export interface, as it is not currently in use.

guowl3 (approved):
lgtm

CHLK (approved):
lgtm

PR #4240, opened 2025-02-24
fix(pl debug):the username configured for new odp connections does not contain cluster names.

zijiacj:
#### What type of PR is this?
type-bug

Fixes #

In pl debug, the username configured for new odp connections does not contain cluster names. If one odp reversed-proxy multiple clusters, an error message is displayed. Therefore, should add the cluster name for odp connection.

```docs

```


MarkPotato777 (approved):
LGTM

PR #4246, opened 2025-02-28
fix(sql-parser): fail to recognize float in pl

LuckyPickleZZ:
#### What type of PR is this?
type-bug

The previous PlSqlLexer.g4 file was unable to recognize floating-point numbers like "123." 

Fixes #

```docs

```


MarkPotato777 (approved):
LGTM

LuckyPickleZZ:
This modification will affect the parsing of statements such as `for a in 1..b loop`. Therefore, PR is temporarily closed

PR #4250, opened 2025-03-03
feat(flow): add cancel flow method without permission

ungreat:
#### What type of PR is this?
type-feature

add cancel flow method without permission

Fixes #

```docs

```


MarkPotato777 (approved):
LGTM

PR #4251, opened 2025-03-03
feat(log): add route log callable

ungreat:
#### What type of PR is this?
type-feature

Instead of having to configure its own log4j2 configuration for each task, provide a routeLog base class

Fixes #

```docs

```


MarkPotato777 (approved):
LGTM

PR #4259, opened 2025-03-04
feat(export):schedule export support sqlPlain and PartitionPlan

ungreat:
#### What type of PR is this?
type-feature

schedule export support sqlPlain and PartitionPlan

Fixes #

```docs

```


server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/ScheduleController.java
```diff
@@ -41,6 +41,7 @@
 import com.oceanbase.odc.service.common.util.WebResponseUtils;
 import com.oceanbase.odc.service.dlm.model.RateLimitConfiguration;
 import com.oceanbase.odc.service.schedule.ScheduleService;
+import com.oceanbase.odc.service.schedule.archiverist.ScheduleTaskExporter;
```
guowl3, line 44: It seems like it hasn’t been used.

ungreat: Removed

guowl3 (approved):
lgtm

PR #4264, opened 2025-03-05
feat(flowInstance): Supports to batch approval flowInstances 

ysjemmm:
#### What type of PR is this?
type-feature

To batch approval flowInstances

### DIFF
--- a/server/odc-common/src/main/java/com/oceanbase/odc/common/json/JsonUtils.java
+++ b/server/odc-common/src/main/java/com/oceanbase/odc/common/json/JsonUtils.java
@@ -173,6 +173,17 @@ public class JsonUtils {
         return innerToJson(OBJECT_MAPPER, obj);
     }
 
+    public static byte[] toBytes(Object obj) {
+        if (obj == null) {
+            return null;
+        }
+        try {
+            return OBJECT_MAPPER.writeValueAsBytes(obj);
+        } catch (JsonProcessingException e) {
+            return null;
+        }
+    }
+
     /**
      * 不会将 @MaskField 注解修饰的字段处理成指定模糊字段(******)
      *
--- a/server/odc-common/src/main/java/com/oceanbase/odc/common/json/JsonUtils.java
+++ b/server/odc-common/src/main/java/com/oceanbase/odc/common/json/JsonUtils.java
@@ -180,6 +180,7 @@ public class JsonUtils {
         try {
             return OBJECT_MAPPER.writeValueAsBytes(obj);
         } catch (JsonProcessingException e) {
+            log.warn("parses obj to bytes failed, reason = {}", e.getMessage());
             return null;
         }
     }
--- a/server/odc-common/src/test/java/com/oceanbase/odc/common/json/JsonUtilsTest.java
+++ b/server/odc-common/src/test/java/com/oceanbase/odc/common/json/JsonUtilsTest.java
@@ -200,7 +200,7 @@ public class JsonUtilsTest {
     }
 
     @Test
-    public void toJsonBytes() {
+    public void toJsonBytes() throws IOException {
         MockObj mockObj = new MockObj();
         mockObj.setList(Arrays.asList(1L, 2L, 3L));
         mockObj.setMsg("test");
@@ -212,7 +212,6 @@ public class JsonUtilsTest {
             MockObj deserializedMockObj = objectMapper.readValue(bos.toByteArray(), MockObj.class);
             Assert.assertEquals(deserializedMockObj.getMsg(), mockObj.getMsg());
             Assert.assertEquals(deserializedMockObj.getList(), mockObj.getList());
-        } catch (IOException ignored) {
         }
     }
 
@@ -241,7 +240,7 @@ public class JsonUtilsTest {
     }
 
     @Data
-    public static class MockObj {
+    private static class MockObj {
 
         private Collection<Long> list;
 
--- a/server/odc-common/src/test/java/com/oceanbase/odc/common/json/JsonUtilsTest.java
+++ b/server/odc-common/src/test/java/com/oceanbase/odc/common/json/JsonUtilsTest.java
@@ -15,8 +15,12 @@
  */
 package com.oceanbase.odc.common.json;
 
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import java.time.OffsetDateTime;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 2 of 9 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
```diff
@@ -682,6 +687,83 @@ public FlowInstanceDetailResp approve(@NotNull Long id,
...
+                    response.getContentByType(new TypeReference<ListResponse<FlowInstanceDetailResp>>() {}).getData()
+                            .getContents(),
+                    new ArrayList<>()));
+        }
+        if (notificationProperties.isEnabled()) {
+            try {
+                eventBuilder
+                        .ofApprovedTasks(listTaskByFlowInstanceIds(flowInstanceIds),
+                                authenticationFacade.currentUserId())
+                        .forEach(e -> broker.enqueueEvent(e));
```
LuckyPickleZZ, line 743: There are multiple database operations in this enqueue method. The performance of multiple enqueues will be poor. 
We'd better add a `batchEnqueue(List<Event>events)` or `enqueue(List<Event>events)` method overload. Batch write events in this method.

server/odc-common/src/main/java/com/oceanbase/odc/common/json/JsonUtils.java
```diff
@@ -173,6 +173,17 @@ public static String toJson(Object obj) {
...
     }
 
+    public static byte[] toBytes(Object obj) {
+        if (obj == null) {
+            return null;
+        }
+        try {
+            return OBJECT_MAPPER.writeValueAsBytes(obj);
+        } catch (JsonProcessingException e) {
+            return null;
```
LioRoger, line 183: is log needed here?

server/odc-common/src/test/java/com/oceanbase/odc/common/json/JsonUtilsTest.java
```diff
@@ -195,6 +199,23 @@ public void convert_jsonToXml() {
...
+        mockObj.setMsg("test");
+        byte[] bytes = JsonUtils.toBytes(mockObj);
+        try (ByteArrayOutputStream bos = new ByteArrayOutputStream()) {
+            bos.write(bytes);
+            bos.flush();
+            ObjectMapper objectMapper = new ObjectMapper();
+            MockObj deserializedMockObj = objectMapper.readValue(bos.toByteArray(), MockObj.class);
+            Assert.assertEquals(deserializedMockObj.getMsg(), mockObj.getMsg());
+            Assert.assertEquals(deserializedMockObj.getList(), mockObj.getList());
+        } catch (IOException ignored) {
```
LioRoger, line 215: if IOException thrown, this unit will always return true

server/odc-common/src/test/java/com/oceanbase/odc/common/json/JsonUtilsTest.java
```diff
@@ -219,6 +240,14 @@ public static class TTT {
         private TT a;
     }
 
+    @Data
+    public static class MockObj {
```
LioRoger, line 244: this inner class should defined as private scope

server/odc-service/src/main/java/com/oceanbase/odc/service/dispatch/RequestDispatcher.java
```diff
@@ -76,6 +79,40 @@ public DispatchResponse forward(@NonNull String ip, @NonNull Integer port) throw
         return forward(ip, port, requestProvider.getRequest(), requestProvider.getRequestBody());
     }
 
+    public DispatchResponse forward(@NonNull String ip, @NonNull Integer port, ByteArrayOutputStream requestBody)
+            throws IOException {
+        return forward(ip, port, requestProvider.getRequest(), requestBody);
+    }
+
+    public DispatchResponse forwardWithRequestBodyData(@NonNull String ip, @NonNull Integer port,
```
LioRoger, line 87: general, method with same name should defined following each other as a method override, move forwardWithRequestBodyData to first place.

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
```diff
@@ -682,6 +687,83 @@ public FlowInstanceDetailResp approve(@NotNull Long id,
...
+        if (CollectionUtils.isEmpty(taskEntities)) {
+            return Collections.emptyList();
+        }
+
+        Map<Long, Long> taskId2FlowInstanceId = listServiceTaskByFlowInstanceIds(flowInstanceIds).stream().filter(
+                t -> t.getTargetTaskId() != null)
+                .collect(Collectors.toMap(ServiceTaskInstanceEntity::getTargetTaskId,
+                        ServiceTaskInstanceEntity::getFlowInstanceId, (exist, value) -> exist));
+
+        Map<String, List<Long>> HostPortKey2FlowInstanceId = new HashMap<>();
```
LioRoger, line 707: variable name should  in Camel-Case

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
```diff
@@ -682,6 +687,83 @@ public FlowInstanceDetailResp approve(@NotNull Long id,
...
+        List<TaskEntity> tasksNotToForward = new ArrayList<>();
+
+        for (TaskEntity taskEntity : taskEntities) {
+            if (taskEntity.getTaskType() == TaskType.IMPORT && !dispatchChecker.isTaskEntityOnThisMachine(taskEntity)) {
```
LioRoger, line 711: i don't get what's this codes do,  why task type is needed

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
ysjemmm: Ok.I know and will change

> There are multiple database operations in this enqueue method. The performance of multiple enqueues will be poor. We'd better add a `batchEnqueue(List<Event>events)` or `enqueue(List<Event>events)` method overload. Batch write events in this method.

server/odc-common/src/test/java/com/oceanbase/odc/common/json/JsonUtilsTest.java
ysjemmm: I see that the static classes defined in this class before are all public, but private is more rigorous

ysjemmm: > this inner class should defined as private scope

I see that the static classes defined in this class before are all public, but private is more rigorous

server/odc-common/src/main/java/com/oceanbase/odc/common/json/JsonUtils.java
ysjemmm: > is log needed here?
It may be necessary to log

server/odc-service/src/main/java/com/oceanbase/odc/service/dispatch/RequestDispatcher.java
ysjemmm: > general, method with same name should defined following each other as a method override, move forwardWithRequestBodyData to first place.

That's a good constraint

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
ysjemmm: > variable name should in Camel-Case

Sorry，this is my problem

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
ysjemmm: > i don't get what's this codes do, why task type is needed

see `FlowInstanceService#approve()`，it is meaning that：
> For import tasks, since the file is uploaded to a machine, the actual execution of the task must also be on that machine. If the approval action is sent to a machine other than the one where the task is located, it needs to be forwarded, otherwise the task will report an error because it cannot find the uploaded file. The reason why asynchronous execution does not have this problem is that {@code objectStorage} is added. If {@code objectStorage} is also added to the import task, this logic is not required

PR #4269, opened 2025-03-07
fix(table permission): When executing sql in an sql window, a permission check on a table involves a database that does not exist

zijiacj:
#### What type of PR is this?
type-bug

When executing sql in an sql window, a permission check on a table involves a database that does not exist

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/permission/DBResourcePermissionHelper.java
```diff
@@ -260,10 +260,13 @@ public List<UnauthorizedDBResource> filterUnauthorizedDBResources(
         // Pre-handle DBResource (fill in databaseId, databaseName, tableId, tableName)
         Set<Long> dsIds = resource2Types.keySet().stream().map(DBResource::getDataSourceId).collect(Collectors.toSet());
         List<DatabaseEntity> dbEntities = databaseRepository.findByConnectionIdIn(dsIds);
+        List<DatabaseEntity> realDbEntities =
```
MarkPotato777, line 263: realDbEntities ---> existingDatabases

zijiacj: ok

MarkPotato777 (approved):
LGTM

PR #4275, opened 2025-03-10
feat(config): add ddl 4 create table and load meta data

yiminpeng:
#### What type of PR is this?
feat

This PR has added the DDL for create table 
and the YAML file for loading metadata.

Fixes #

```docs

```


server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_5__config_organization_configuration.sql
```diff
@@ -0,0 +1,17 @@
+-- drop table config_organization_configuration
+DROP TABLE IF EXISTS `config_organization_configuration`;
+
+-- v4.3.4: create table config_organization_configuration
+CREATE TABLE IF NOT EXISTS `config_organization_configuration`(
+  `id` bigint(20) NOT NULL AUTO_INCREMENT COMMENT 'auto-increment id',
+  `organization_id` bigint(20) NOT NULL COMMENT 'organization id, reference config_organization.id',
```
MarkPotato777, line 7: what is `config_organization.id` ?

server/odc-service/src/main/java/com/oceanbase/odc/service/config/OrganizationConfigKeys.java
```diff
@@ -18,7 +18,7 @@
 public class OrganizationConfigKeys {
     public static final String DEFAULT_QUERY_LIMIT = "odc.sqlexecute.default.queryLimit";
 
-    public static final String DEFAULT_QUERY_COUNT = "odc.sqlexecute.default.queryCount";
+    public static final String DEFAULT_DEFAULT_QUERY_LIMIT = "odc.sqlexecute.default.defaultQueryLimit";
```
MarkPotato777, line 21: why `DEFAULT_DEFAULT_xxx` ?

server/odc-service/src/main/resources/config-meta/organization-config-meta.yml
```diff
@@ -0,0 +1,83 @@
...
+# - category options: user, indivOrganization, teamOrganization, system
+# - type options: string, integer, decimal, boolean, list
+- id: 1
+  category: 'teamOrganization'
+  key: 'odc.sqlexecute.default.queryLimit'
+  type: 'integer'
+  max_value: 100000
+  min_value: 1
+  allowed_values: ~
+  default_value: '1000'
```
MarkPotato777, line 11: why default_value is a string, but max_value and min_value is a number? Same as below

server/odc-service/src/main/java/com/oceanbase/odc/service/config/OrganizationConfigKeys.java
yiminpeng: changed to `DEFAULT_QUERY_LIMIT` and `DEFAULT_MAX_QUERY_LIMIT`

server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_5__config_organization_configuration.sql
yiminpeng: actually is `iam_organization` id

server/odc-service/src/main/resources/config-meta/organization-config-meta.yml
yiminpeng: because I need this value as metadata and the default configuration value if the user does not update the configuration.

MarkPotato777 (approved):

PR #4278, opened 2025-03-11
feat(task): let task v2 force start executor in pull mode

LioRoger:
#### What type of PR is this?

type-feature
PullTaskV2 is always run for task v2 mode, so let process mode run in pull mode to avoid duplicate update of job_job

Fixes #

```docs

```


guowl3 (approved):
lgtm

PR #4279, opened 2025-03-13
fix(task): taskResult event not send before task finished

LioRoger:
#### What type of PR is this?

type-bug
PullResult should trigger event notify if taskResult changed or task finished. previous implement not send Event with taskResult before task finished

Fixes #

```docs

```


guowl3 (approved):
lgtm

PR #4276, opened 2025-03-11
feat(schedule): add schedule task import api

ungreat:
#### What type of PR is this?
type-feature

#### What this PR does / why we need it:

Add schedule task import api.

#### Which issue(s) this PR fixes:

Fixes #

#### Additional documentation e.g., usage docs, etc.:

```docs

```

PR #4281, opened 2025-03-14
feat(schdule): add schedule task import

ungreat:
#### What type of PR is this?
type-feature

Added implementation of the ScheduleImport API.  relate with https://github.com/oceanbase/odc/pull/4276

The main objective of this PR is to enable the import of exported Schedule zip files into ODC and recreate new Schedule tasks. For the same Schedule task within the same file, ODC will only create it once. Additionally, ODC validates the file content against the key provided during import, and modifications to the imported files are strictly prohibited.

The import feature provided by this PR only allows files exported from the same version of ODC to be successfully imported, while cross-version imports are also explicitly prohibited.

The community version does not currently provide the ability to import schedule task.

Fixes #

The core implementation class for Schedule import is scheduleTaskImporter.
The ScheduleExportImportFacade is designed to abstract away the implementation differences between the Community Edition and the Cloud Service.

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/config/ScheduleConfiguration.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/config/ScheduleConfiguration.java
@@ -311,4 +311,21 @@ public class ScheduleConfiguration {
         return executor;
     }
 
+    @Bean(name = "scheduleImportExecutor")
+    public ThreadPoolTaskExecutor scheduleImportExecutor() {
+        ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
+        int minPoolSize = Math.max(SystemUtils.availableProcessors(), 8);
+        executor.setCorePoolSize(minPoolSize);
+        executor.setMaxPoolSize(minPoolSize * 2);
+        executor.setQueueCapacity(10);
+        executor.setThreadNamePrefix("schedule-import-");
+        executor.setWaitForTasksToCompleteOnShutdown(false);
+        executor.setAwaitTerminationSeconds(5);
+        executor.setTaskDecorator(new TraceDecorator<>());
+        executor.setRejectedExecutionHandler(new ThreadPoolExecutor.AbortPolicy());
+        executor.initialize();
+        log.info("scheduleImportExecutor initialized");
+        return executor;
+    }
+
 }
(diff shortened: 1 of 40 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/config/ScheduleConfiguration.java
```diff
@@ -311,4 +311,21 @@ public ThreadPoolTaskExecutor queryProfileMonitorExecutor() {
         return executor;
     }
 
+    @Bean(name = "scheduleImportExecutor")
+    public ThreadPoolTaskExecutor scheduleImportExecutor() {
+        ThreadPoolTaskExecutor executor = new ThreadPoolTaskExecutor();
+        int minPoolSize = Math.max(SystemUtils.availableProcessors(), 8);
```
LioRoger, line 317: import should be sync,  so i thought it's not necessary

server/odc-service/src/main/java/com/oceanbase/odc/service/exporter/utils/JsonExtractorFactory.java
```diff
@@ -0,0 +1,101 @@
...
+        jsonExtractor.setExportedFile(new ExportedFile(configJson, secret));
+        return jsonExtractor;
+    }
+
+    public static JsonExtractor buildJsonExtractor(ExportedFile exportedFile, String tempPath) throws IOException {
+        JsonExtractor jsonExtractor = new JsonExtractor();
+        // Create a random directory within the specified destination path
+        Path randomDir = Files.createTempDirectory(new File(tempPath).toPath(), "unzipped-");
+
+        // Create a temporary file to save the InputStream contents
```
LioRoger, line 78: avoid duplicate code

ungreat: have extracted to a method.

server/odc-service/src/main/java/com/oceanbase/odc/config/ScheduleConfiguration.java
ungreat: sync will lead to an increase in api response time, and it will be proportional to the scale.

guowl3 (approved):
lgtm

LioRoger (approved):
lgtm

PR #4282, opened 2025-03-14
fix(task): cancel task  may not wait log meta uploaded

LioRoger:
… saveOrUpdateLogMetadata method

#### What type of PR is this?

type-bug
cancel task http call may re-entry, add isTerminating when canceling has called
add timeout for PullTaskResultV2 handle doCanceling status
add transactional annotate for  saveOrUpdateLogMetadata method call

Fixes #

```docs

```


guowl3 (approved):
lgtm

PR #4286, opened 2025-03-17
fix(task): correct executorIdenfifier port to odc server port for process mode

LioRoger:
…cause log get required this port

#### What type of PR is this?

type-bug
the Port in ExecutorIdentifier have different meaning in different run mode.
In process mode, it represent the ODC server port and will be used for log content get request process.
In K8s mode, it represent the Agent port for future request send.
Correct logic to fix previous logic.

Fixes #

```docs

```


guowl3 (approved):
lgtm

PR #4294, opened 2025-03-19
fix(Script): only one script download when select many same name scripts

CHLK:
#### What type of PR is this?
type-bug

fix bug: only one script download when select many same name scripts

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/script/ScriptService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/script/ScriptService.java
@@ -25,8 +25,10 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/script/ScriptService.java
```diff
@@ -332,10 +334,15 @@ public ResponseEntity<InputStreamResource> batchDownload(List<Long> scriptIds) t
...
             }
+            Map<String, Integer> duplicateNameFileCountMap = new HashMap<>();
             for (ScriptMetaEntity objectMetadata : scriptMetas) {
                 StorageObject storageObject =
                         objectStorageFacade.loadObject(objectMetadata.getBucketName(), objectMetadata.getObjectId());
-                File file = new File(downloadDir, objectMetadata.getObjectName());
+                int count = duplicateNameFileCountMap.compute(objectMetadata.getObjectName(),
+                        (k, v) -> v == null ? 1 : v + 1);
+                String fileName =
+                        count > 1 ? objectMetadata.getObjectName() + "(" + count + ")" : objectMetadata.getObjectName();
```
LuckyPickleZZ, line 344: Shouldn't file names have a numeric suffix when there are no duplicate named objects?

LioRoger (approved):
lgtm

LuckyPickleZZ (approved):

PR #4295, opened 2025-03-19
fix(import): schedule import not have default name and lack of migrate sql 

ungreat:
#### What type of PR is this?
type-bug

1、add imported schedule default name.
2、add migrate sql to create `import_file_row_history`

Fixes #

```docs

```


guowl3 (approved):
lgtm

PR #4296, opened 2025-03-19
feat(Sql Check): Supports to get predicted affectedRows of DML sql

ysjemmm:
#### What type of PR is this?
type-feature

 to get predicted affectedRows of DML sql

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
@@ -57,6 +57,7 @@ import org.springframework.transaction.annotation.Transactional;
 import org.springframework.validation.annotation.Validated;
 
 import com.fasterxml.jackson.core.type.TypeReference;
+import com.google.common.collect.Sets;
 import com.oceanbase.odc.common.event.EventPublisher;
 import com.oceanbase.odc.common.i18n.I18n;
 import com.oceanbase.odc.common.json.JsonUtils;
@@ -130,6 +131,7 @@ import com.oceanbase.odc.service.flow.model.FlowMetaInfo;
 import com.oceanbase.odc.service.flow.model.FlowNodeInstanceDetailResp.FlowNodeInstanceMapper;
 import com.oceanbase.odc.service.flow.model.FlowNodeStatus;
 import com.oceanbase.odc.service.flow.model.FlowNodeType;
+import com.oceanbase.odc.service.flow.model.PreCheckTaskResult;
 import com.oceanbase.odc.service.flow.model.QueryFlowInstanceParams;
 import com.oceanbase.odc.service.flow.processor.EnablePreprocess;
 import com.oceanbase.odc.service.flow.task.BaseRuntimeFlowableDelegate;
@@ -539,11 +541,23 @@ public class FlowInstanceService {
     }
 
     public FlowInstanceDetailResp detail(@NotNull Long id) {
-        return mapFlowInstanceWithReadPermission(id, flowInstance -> {
+        FlowInstanceDetailResp flowInstanceDetailResp = mapFlowInstanceWithReadPermission(id, flowInstance -> {
             FlowInstanceMapper instanceMapper = mapperFactory.generateMapperByInstance(flowInstance, false);
             FlowNodeInstanceMapper nodeInstanceMapper = mapperFactory.generateNodeMapperByInstance(flowInstance, false);
             return instanceMapper.map(flowInstance, nodeInstanceMapper);
         });
+        // Only DatabaseChange supports to get DML affectedRows
+        if (flowInstanceDetailResp != null && flowInstanceDetailResp.getType() == TaskType.ASYNC) {
+            Optional<TaskEntity> preCheckOp = getPreCheckTaskByFlowInstanceId(id);
+            if (preCheckOp.isPresent()) {
+                PreCheckTaskResult preCheckTaskResult = JsonUtils.fromJson(preCheckOp.get().getResultJson(),
+                        PreCheckTaskResult.class);
+                if (preCheckTaskResult != null && preCheckTaskResult.getSqlCheckResult() != null) {
+                    flowInstanceDetailResp.setAffectedRows(preCheckTaskResult.getSqlCheckResult().getAffectedRows());
+                }
+            }
+        }
+        return flowInstanceDetailResp;
     }
 
     @Transactional(rollbackFor = Exception.class)
@@ -752,6 +766,21 @@ public class FlowInstanceService {
         Verify.singleton(taskIds, "Multi task for one instance is not allowed, id " + id);
         Long taskId = taskIds.iterator().next();
         return taskService.detail(taskId);
+    }
+
+    public Optional<TaskEntity> getPreCheckTaskByFlowInstanceId(@NonNull Long flowInstanceId) {
+        List<ServiceTaskInstanceEntity> serviceTasks = serviceTaskRepository.findByFlowInstanceIdIn(
+                Sets.newHashSet(flowInstanceId))
+                .stream()
+                .filter(s -> s.getTaskType() == TaskType.PRE_CHECK || s.getTaskType() == TaskType.SQL_CHECK)
+                .collect(Collectors.toList());
+        if (CollectionUtils.isEmpty(serviceTasks)) {
+            return Optional.empty();
+        }
+        Verify.singleton(serviceTasks,
+                "Multi preCheckTask or sqlCheckTask for one instance is not allowed, flowInstanceId " + flowInstanceId);
+        TaskEntity task = taskService.detail(serviceTasks.get(0).getTargetTaskId());
+        return Optional.of(task);
     }
 
     private void checkCreateFlowInstancePermission(CreateFlowInstanceReq req) {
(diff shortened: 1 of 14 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
```diff
@@ -539,11 +541,23 @@ public List<FlowInstanceEntity> listByIds(@NonNull Collection<Long> ids) {
...
 
     public FlowInstanceDetailResp detail(@NotNull Long id) {
-        return mapFlowInstanceWithReadPermission(id, flowInstance -> {
+        FlowInstanceDetailResp flowInstanceDetailResp = mapFlowInstanceWithReadPermission(id, flowInstance -> {
             FlowInstanceMapper instanceMapper = mapperFactory.generateMapperByInstance(flowInstance, false);
             FlowNodeInstanceMapper nodeInstanceMapper = mapperFactory.generateNodeMapperByInstance(flowInstance, false);
             return instanceMapper.map(flowInstance, nodeInstanceMapper);
         });
+        // Only DatabaseChange supports to get DML affectedRows
+        if (flowInstanceDetailResp != null && flowInstanceDetailResp.getType() == TaskType.ASYNC) {
```
MarkPotato777, line 550: It's wield to put `affectedRows` in `FlowInstance`? Any other better solution?

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/BaseSqlChecker.java
```diff
@@ -159,8 +166,40 @@ private List<OffsetString> splitByCommentProcessor(String sqlScript) {
         return sqls;
     }
 
+    public long getAffectedRows(@NonNull List<OffsetString> offsetStrings) {
```
MarkPotato777, line 169: My suggestion is that do not put this customized method into the `SqlChecker`, which would compromise the abstraction and encapsulation of the code. How about creating a new class to do this customized thing? You may make the `SqlChecker` as your field or parameter in your new class, and use it to complete your customized business.

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/SqlCheckRule.java
```diff
@@ -39,4 +39,8 @@ public interface SqlCheckRule {
 
     List<DialectType> getSupportsDialectTypes();
 
+    default SqlCheckRule getRule() {
+        return this;
```
MarkPotato777, line 43: Why do you need this method?

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/SqlCheckService.java
```diff
@@ -205,4 +226,18 @@ public List<Rule> fullFillRiskLevel(List<Rule> rules, @NonNull List<CheckViolati
         return violatedRules;
     }
 
+    private Predicate<Rule> getRulePredicate() {
+        return rule -> {
+            RuleMetadata metadata = rule.getMetadata();
+            if (metadata == null) {
+                return false;
+            } else if ((!Boolean.TRUE.equals(rule.getEnabled()))
+                    && !"${com.oceanbase.odc.builtin-resource.regulation.rule.sql-check.restrict-sql-affected-rows.name}"
```
MarkPotato777, line 235: `&&` or `||`?

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
ysjemmm: > It's wield to put `affectedRows` in `FlowInstance`. Any other better solution?

Unless it is divided into two interfaces, it can be displayed in the details of the flow instance

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/BaseSqlChecker.java
ysjemmm: > My suggestion is that do not put this customized method into the `SqlChecker`, which would compromise the abstraction and encapsulation of the code. How about creating a new class to do this customized thing? You may make the `SqlChecker` as your field or parameter in your new class, and use it to complete your customized business.

I will care

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/SqlCheckRule.java
ysjemmm: > Collaborator
> ###

I want to calc affectedRows by `BaseAffectedRowExceedLimit`,  but it is wrapped in a `SqlCheckRuleWrapper`，I can not get `BaseAffectedRowExceedLimit` from  `SqlCheckRuleWrapper` beacuse it is `private`, so I'm going to add a method to the `SqlCheckRuleWrapper` parent class, which is `SqlCheckRule`, to get subClass of current-SqlCheckRule

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
MarkPotato777: Is it better to put this logic to `FlowInstanceMapper`?

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/AffectedRowCalculator.java
```diff
@@ -0,0 +1,122 @@
...
+                    affectedRows += calculator.getStatementAffectedRows(statement);
+                } catch (Exception e) {
+                    log.warn("Get affected rows failed", e);
+                }
+            }
+        }
+        return affectedRows;
+    }
+
+    private List<OffsetString> splitSql(String sqlScript) {
```
MarkPotato777, line 87: Duplicated code. Extract it to a util class.

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
ysjemmm: > Is it better to put this logic to `FlowInstanceMapper`?

I will look

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/AffectedRowCalculator.java
ysjemmm: > Duplicated code. Extract it to a util class.

I will care

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/SqlCheckUtil.java
```diff
@@ -253,4 +255,28 @@ public static String getDbVersion(ConnectionConfig config, DataSource dataSource
         }
     }
 
+    public static List<OffsetString> splitSql(@NonNull String sqlScript, @NonNull DialectType dialectType,
```
MarkPotato777, line 258: Same with `BaseSqlChecker#splitByCommentProcessor`? Can it be replaced?

ysjemmm: > Same with `BaseSqlChecker#splitByCommentProcessor`? Can it be replaced?

Base the class of `BaseSqlChecker`, due to constructing a method of tool, can replace `BaseSqlChecker#splitByCommentProcessor` with `SqlCheckUtils#splitSql`

ysjemmm: > Same with `BaseSqlChecker#splitByCommentProcessor`? Can it be replaced?

I don't think it's necessary to introduce BaseSqlChecker to use this method

ysjemmm: > Same with `BaseSqlChecker#splitByCommentProcessor`? Can it be replaced?

The business-independent common method I think can be structured as a tool method

server/odc-service/src/main/java/com/oceanbase/odc/service/sqlcheck/SqlCheckService.java
```diff
@@ -205,4 +231,21 @@ public List<Rule> fullFillRiskLevel(List<Rule> rules, @NonNull List<CheckViolati
         return violatedRules;
     }
 
+    private Predicate<Rule> getRulePredicate(boolean filterAll) {
```
MarkPotato777, line 234: This method is hard to understand:
1. The parameter is wield. It seems unnecessary. If the caller wants to filter out nothing, they could just pass `r -> true`, which is clearer.
2. The method name does not match your implementation. The method actually does filter the `sql-affected-rows rule`, while the method name is just `getRulePredicate`.

ysjemmm: > This method is hard to understand:
> 
> 1. The parameter is wield. It seems unnecessary. If the caller wants to filter out nothing, they could just pass `r -> true`, which is clearer.
> 2. The method name does not match your implementation. The method actually does filter the `sql-affected-rows rule`, while the method name is just `getRulePredicate`.

I see what you mean. Because of the encapsulation of these methods, initially this method only filters out `SqlCheckRules` that have already been started, but now I have added a condition that filters all `SqlCheckRules` so that I can get the `sql-affected-rows rule`. But maybe I can just filter out the `sql-affected-rows rule` right here

zijiacj (approved):
lgtm

PeachThinking (approved):
lgtm

guowl3 (approved):

PR #4301, opened 2025-03-21
fix(cloud): fail to start odc when oss location is finance region

LuckyPickleZZ:
#### What type of PR is this?
type-bug


MarkPotato777 (approved):
LGTM

PR #4318, opened 2025-03-27
fix(sql-execute): remove length check when querying sql audit

LuckyPickleZZ:
#### What type of PR is this?
type-bug

OB 4.2.5 has added an internal variable, which is named ''eable_sql_audit_query_sql'. If this variable is true, then the query SQL in the SQL audit view is empty.


LioRoger (approved):
lgtm

PR #4319, opened 2025-03-27
fix(permission): authentication will wrongly fail when resource ids are the same but resource types are different

MarkPotato777:
#### What type of PR is this?

type-bug
This PR fixed two issues of permission framework:
1. when quering resource roles, it does not query by organization id, which may cause to horizontal authetication bug;
2. when resource ids are the same, but they are different resource types, it has logic error when authentication.

Fixes #

```docs

```


PeachThinking (approved):
lgtm

PeachThinking (approved):
lgtm

PR #4325, opened 2025-03-31
fix(session): fix query limit conflict

yiminpeng:
#### What type of PR is this?
It solves the adaptation problem of the query limit, and ensures that there will be no exceptions when front interaction and internal call.

```docs

```


LuckyPickleZZ (approved):
lgtm

PR #4331, opened 2025-04-02
feat(materialized view):add more display information for materialized view 

zijiacj:
#### What type of PR is this?
type-feature

Fixes #

```docs

```


libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/editor/DBMViewEditor.java
```diff
@@ -0,0 +1,119 @@
...
+
+    @Override
+    public boolean editable() {
+        return true;
+    }
+
+    @Override
+    public String generateUpdateObjectListDDL(Collection<DBMaterializedView> oldObjects,
+            Collection<DBMaterializedView> newObjects) {
+        throw new UnsupportedOperationException("Not supported yet");
```
LioRoger, line 91: as it's unsupported why this function is needed
if there are too much derived implement as exception thrown, you should consider if that class design is suitable for your situation
base class should implement basic logic。logic may derived use abstract method to do that

zijiacj: Consistent with the implementation of the `DBTableEditor`, only provides editing operations on a single object, not multiple objects.

MarkPotato777 (approved):
LGTM

PR #4332, opened 2025-04-02
fix(schedule): the partition plan statistics are incorrect

ysjemmm:
#### What type of PR is this?
type-bug

At present, partition plans are still managed by the Flow module, and there is no associated Schedule module, resulting in inaccurate statistics


guowl3 (approved):
lgtm

zijiacj (approved):
lgtm

PR #4335, opened 2025-04-07
feat(session): support client session

ungreat:
#### What type of PR is this?
type-feature

Remove the restriction of `odc.session.kill-query-or-session.max-supported-ob-version`;

Duplicate SIDs are removed during kill session。

Fixes #

```docs

```


server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_9__remove_kill_sessesion_max_support_version.sql
```diff
@@ -0,0 +1 @@
+update `config_system_configuration` set `value` = 0 where `key` = 'odc.session.kill-query-or-session.max-supported-ob-version';
```
MarkPotato777, line 1: is `value` = 0 --> `value` = '0' better?

ungreat: It's more reasonable.

LioRoger (approved):
lgtm

PR #4338, opened 2025-04-08
fix(schedule): landing page displayed number of partition plans enabled is incorrect

ysjemmm:
#### What type of PR is this?
type-bug

- Adjust the logic for obtaining the number of enabled partitions by querying the flow_isntance table
- Adjusted to get all sub-tasks of a periodic parent task instead of grouping them. For example, the sub-tasks of a periodic data archiving task include data cleaning + data archiving. The actual number of tasks should be 2 instead of 1
- Adjust the structure of the code to make it more concise


guowl3 (approved):
lgtm

zijiacj (approved):
lgtm

PR #4339, opened 2025-04-08
fix(audit): add intsance migrate audit meta data

ungreat:
#### What type of PR is this?
type-bug

add instance migrate audit event metadata

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/audit/util/AuditEventMapper.java
```diff
@@ -25,6 +26,8 @@
 public interface AuditEventMapper {
     AuditEventMapper INSTANCE = Mappers.getMapper(AuditEventMapper.class);
 
+    @Mapping(target = "actionName", expression = "java(entity.getAction().getLocalizedMessage())")
+    @Mapping(target = "typeName", expression = "java(entity.getType().getLocalizedMessage())")
```
MarkPotato777, line 30: Why do we need i18n here? I remember that the i18n here will be done by the front end. Can the front end identify this if you do i18n in the response?

ungreat: now there's internationalization on both the front and the back, it would be more appropriate to internationalize the backend here.

MarkPotato777 (approved):
LGTM

server/odc-service/src/main/java/com/oceanbase/odc/service/audit/model/AuditEventMeta.java
```diff
@@ -45,11 +45,21 @@ public class AuditEventMeta {
      */
     private AuditEventType type;
 
+    /**
+     * Audit event name
+     */
+    private String typeName;
```
LioRoger, line 51: what's different between typeName and type, can those variable can be inferred by each other?

PR #4341, opened 2025-04-08
feat(config): Add a data source key configuration item in the organization configuration.

yiminpeng:
#### What type of PR is this?
feat

This PR add a data source key configuration item in the organization configuration.

-->
```docs

```

### DIFF
--- a/server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/V4349OrganizationSecretMigrate.java
+++ b/server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/V4349OrganizationSecretMigrate.java
@@ -16,7 +16,6 @@
 package com.oceanbase.odc.migrate.jdbc.common;
 
 import java.util.Arrays;
-import java.util.Base64;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
@@ -33,6 +32,7 @@ import com.oceanbase.odc.core.migrate.JdbcMigratable;
 import com.oceanbase.odc.core.migrate.Migratable;
 import com.oceanbase.odc.metadb.iam.OrganizationEntity;
 
+import cn.hutool.core.codec.Caesar;
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
@@ -65,7 +65,7 @@ public class V4349OrganizationSecretMigrate implements JdbcMigratable {
         String sql = "update iam_organization set secret=? where id=?";
         List<Object[]> parameters = organizationList.stream()
                 .map(organization -> new Object[] {
-                        Base64.getEncoder().encodeToString(organization.getSecret().getBytes()), organization.getId()})
+                        Caesar.encode(organization.getSecret(), 8), organization.getId()})
                 .collect(Collectors.toList());
         AtomicReference<Exception> thrown = new AtomicReference<>(null);
         Long total = transactionTemplate.execute(status -> {
--- a/server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/V4349OrganizationSecretMigrate.java
+++ b/server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/V4349OrganizationSecretMigrate.java
@@ -52,11 +52,7 @@ public class V4349OrganizationSecretMigrate implements JdbcMigratable {
         log.info("organization secret migrate started, organizationCount={}", organizationList.size());
         int total = 0;
         for (OrganizationEntity organization : organizationList) {
-
-            log.info("previous organization secret, organization secret={}", organization.getSecret());
             organization.setSecret(Base64.getEncoder().encodeToString(organization.getSecret().getBytes()));
-            log.info("new organization secret, organization secret={}", organization.getSecret());
-
             total += migrateForOrganization(organization);
         }
         log.info("organization secret migrated, organizationCount={}, total={}", organizationList.size(), total);
--- a/server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/V4349OrganizationSecretMigrate.java
+++ b/server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/V4349OrganizationSecretMigrate.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright (c) 2025 OceanBase.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.oceanbase.odc.migrate.jdbc.common;
+
+import java.util.Base64;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import javax.sql.DataSource;
+
+import org.springframework.jdbc.core.BeanPropertyRowMapper;
+import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.jdbc.datasource.DataSourceTransactionManager;
+import org.springframework.transaction.support.TransactionTemplate;
+import org.springframework.util.CollectionUtils;
+
+import com.oceanbase.odc.core.migrate.JdbcMigratable;
+import com.oceanbase.odc.core.migrate.Migratable;
+import com.oceanbase.odc.metadb.iam.OrganizationEntity;
+
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+//@Migratable(version = "4.3.4.9", description =
+//    "migrate the organization secret to this stored in confusion")
+public class V4349OrganizationSecretMigrate implements JdbcMigratable {
+
+    private JdbcTemplate jdbcTemplate;
+    private TransactionTemplate transactionTemplate;
+
+    @Override
+    public void migrate(DataSource dataSource) {
+        this.jdbcTemplate = new JdbcTemplate(dataSource);
+        this.transactionTemplate = new TransactionTemplate(new DataSourceTransactionManager(dataSource));
+        List<OrganizationEntity> organizationList = listOrganizations();
+        if (CollectionUtils.isEmpty(organizationList)) {
+            log.info("organization secret migrate skipped, organization list is empty");
+            return;
+        }
+        log.info("organization secret migrate started, organizationCount={}", organizationList.size());
+        int total = 0;
+        for (OrganizationEntity organization : organizationList) {
+
+            log.info("previous organization secret, organization secret={}", organization.getSecret());
+            organization.setSecret(Base64.getEncoder().encodeToString(organization.getSecret().getBytes()));
+            log.info("new organization secret, organization secret={}", organization.getSecret());
+            log.info("organization secret decode, organization secret decode={}",
+                new String(Base64.getDecoder().decode(organization.getSecret())));
+            total += migrateForOrganization(organization);
+        }
+        log.info("organization secret migrated, organizationCount={}, total={}", organizationList.size(), total);
+    }
+
+    private List<OrganizationEntity> listOrganizations() {
+        String sql = "select `id`, `secret` from iam_organization";
+        return jdbcTemplate.query(sql, new BeanPropertyRowMapper<>(OrganizationEntity.class));
+    }
+
+    private int migrateForOrganization(OrganizationEntity organization) {
+        String sql = "update iam_organization set secret=? where id=?";
+        AtomicInteger row = new AtomicInteger();
+        transactionTemplate.execute(status -> {
+            try {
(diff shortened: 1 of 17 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/config/OrganizationConfigService.java
```diff
@@ -176,4 +218,51 @@ private Map<String, Configuration> internalQuery(Long organizationId) {
...
+            return;
+        }
+        // The key is equal to the old secret
+        if (Objects.equals(customKey, organizationEntity.get().getSecret())) {
+            return;
+        }
+        OrganizationConfigEntity customKeyInDB = organizationConfigDAO.queryByOrganizationIdAndKey(
+                organizationId, DEFAULT_CUSTOM_DATA_SOURCE_ENCRYPTION_KEY);
+        // The key is not configured for the first time
+        if (customKeyInDB != null) {
```
MarkPotato777, line 234: Why do you only care about the first update?

server/odc-service/src/main/java/com/oceanbase/odc/service/config/OrganizationConfigService.java
```diff
@@ -176,4 +218,51 @@ private Map<String, Configuration> internalQuery(Long organizationId) {
...
+            return;
+        }
+        log.info("Start migrate existed datasource password, organizationId={}", organizationId);
+        List<ConnectionConfig> connectionList = connectionService.listByOrganizationId(organizationId);
```
MarkPotato777, line 238: This logic should not be put here as it's all about the `ConnectionConfig`. Generally, the organization config module should not be dependent on the connection module because the configuration module is closer to the low level.

server/odc-service/src/main/java/com/oceanbase/odc/service/config/OrganizationConfigService.java
```diff
@@ -176,4 +218,51 @@ private Map<String, Configuration> internalQuery(Long organizationId) {
...
+                            connectionService.getEncryptedConfig(decryptedConfig, customKey);
+                    return mapper.modelToEntity(reEncryptedConfig);
+                })
+                .collect(Collectors.toList());
+        connectionService.batchUpdateConnectionConfig(reEncryptedList);
+        log.info("Success migrate existed datasource password, organizationId={}", organizationId);
+    }
+
+    private void updateOrganizationSecret(Long organizationId, String customKey) {
+        String secret = customKey.isEmpty() ? PasswordUtils.random(32)
```
MarkPotato777, line 252: `customKey.isEmpty()` may cause NPE.

server/odc-service/src/main/java/com/oceanbase/odc/service/config/OrganizationConfigService.java
```diff
@@ -176,4 +218,51 @@ private Map<String, Configuration> internalQuery(Long organizationId) {
...
+                : Base64.getEncoder().encodeToString(customKey.getBytes());
+        int updateRows = organizationRepo.updateOrganizationSecretById(organizationId, secret);
```
MarkPotato777, line 254: Why do you need to update if the key is empty?

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionEncryption.java
```diff
@@ -74,4 +73,22 @@ TextEncryptor getEncryptor(ConnectionConfig connection) {
         return encryptionFacade.organizationEncryptor(connection.getOrganizationId(), connection.getSalt());
     }
 
+    private void checkCipherAndSalt(ConnectionConfig connection) {
```
MarkPotato777, line 76: Why is the method named 'check"?

yiminpeng: `null` judgment, or to call it `fillCipherAndSaltIfNull` ?

server/odc-service/src/main/java/com/oceanbase/odc/service/config/OrganizationConfigService.java
yiminpeng: The logic here is confused with the previous treatment of individual space, in fact every update requires migrating these connections.

server/odc-service/src/main/java/com/oceanbase/odc/service/config/OrganizationConfigService.java
```diff
@@ -176,4 +218,51 @@ private Map<String, Configuration> internalQuery(Long organizationId) {
                 .collect(Collectors.toMap(Configuration::getKey, c -> c));
     }
 
+    private void migrateExistedDataSourcePassword(Long organizationId, String customKey) {
+        Optional<OrganizationEntity> organizationEntity = organizationRepo.findById(organizationId);
+
+        if (!organizationEntity.isPresent()) {
```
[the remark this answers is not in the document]
yiminpeng: Got it.

server/odc-service/src/main/java/com/oceanbase/odc/service/config/OrganizationConfigService.java
```diff
@@ -128,12 +144,22 @@ public List<Configuration> queryListDefault() {
     public List<Configuration> batchUpdate(@NotNull Long organizationId, @NotNull Long userId,
             @NotEmpty List<Configuration> configurations) {
         validateConfiguration(configurations);
+        String customDataSourceKey = getCustomDataSourceKey(configurations);
```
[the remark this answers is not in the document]
yiminpeng: individual  space does not have this configuration item, here I need to add the logic of returning in advance when the current space is individual.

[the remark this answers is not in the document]
yiminpeng: Got it.

server/odc-service/src/main/java/com/oceanbase/odc/service/config/OrganizationConfigService.java
```diff
@@ -128,12 +144,22 @@ public List<Configuration> queryListDefault() {
...
+        migrateExistedDataSourcePassword(organizationId, customDataSourceKey);
+        updateOrganizationSecret(organizationId, customDataSourceKey);
+
         List<OrganizationConfigEntity> organizationConfigEntities = configurations.stream()
-                .map(record -> record.convert2DO(organizationId, userId))
+                .map(record -> {
```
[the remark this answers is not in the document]
yiminpeng: reasonable, It's been changed.

server/odc-service/src/main/java/com/oceanbase/odc/service/config/OrganizationConfigService.java
yiminpeng: `getCustomDataSourceKey` already checked `null` situation. if the `customDataSourceKey` is present, and the value actually can not be null ? default value is `""`

server/odc-service/src/main/java/com/oceanbase/odc/service/config/OrganizationConfigService.java
yiminpeng: At present, all I can think of is to move the whole logic to `ConnectionService` to make the code a little more concise, but the essence is still dependent on the connection module, I have not thought of how to deal with it.

server/odc-service/src/main/java/com/oceanbase/odc/service/integration/util/EncryptionUtil.java
```diff
@@ -150,4 +151,9 @@ public static String convertCertificateToPem(X509Certificate certificate) throws
                 + base64EncodedCert
                 + "\n" + CERTIFICATE_KEY_SUFFIX;
     }
+
+    public static String encodeByBase64(String rawString) {
```
ungreat, line 155: may use com.oceanbase.odc.common.util.EncodeUtils#base64DecodeFromString

yiminpeng: got it.

server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/V4349OrganizationSecretMigrate.java
```diff
@@ -0,0 +1,82 @@
...
+        String sql = "update iam_organization set secret=? where id=?";
+        AtomicInteger row = new AtomicInteger();
+        transactionTemplate.execute(status -> {
+            try {
+                row.set(jdbcTemplate.update(sql, organization.getSecret(),
+                        organization.getId()));
+            } catch (Exception e) {
+                log.error("organization secret migrate failed, organizationId={}, error={}", organization.getId(),
+                        e.getMessage());
+                status.setRollbackOnly();
```
MarkPotato777, line 76: 1. What will happen if you swallow the exception?
2. What if just a part of the data has been migrated because of an exception? Is your migration repeatable?

server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/V4349OrganizationSecretMigrate.java
```diff
@@ -0,0 +1,82 @@
...
+        this.transactionTemplate = new TransactionTemplate(new DataSourceTransactionManager(dataSource));
+        List<OrganizationEntity> organizationList = listOrganizations();
+        if (CollectionUtils.isEmpty(organizationList)) {
+            log.info("organization secret migrate skipped, organization list is empty");
+            return;
+        }
+        log.info("organization secret migrate started, organizationCount={}", organizationList.size());
+        int total = 0;
+        for (OrganizationEntity organization : organizationList) {
+            organization.setSecret(EncodeUtils.base64EncodeToString(organization.getSecret().getBytes()));
```
MarkPotato777, line 55: Base64 is not a good idea because
1) It may be time-consuming
2) It's not an encryption algorithm. It could be easily decrpted even if the attacters do not get your source code.

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
```diff
@@ -794,6 +794,38 @@ public ConnectionConfig getDecryptedConfig(@NotNull Long id) {
...
+    }
+
+    @SkipAuthorize("internal usage")
+    public ConnectionConfig getEncryptedConfig(@NotNull ConnectionConfig decryptedConfig, String customKey) {
+        return connectionEncryption.encryptPasswordsByCustomKey(decryptedConfig, customKey);
+    }
+
+    @Transactional(rollbackFor = Exception.class)
+    @SkipAuthorize("internal usage")
+    public void batchUpdateConnectionConfig(List<ConnectionEntity> connections) {
```
MarkPotato777, line 826: `@Transactional(rollbackFor = Exception.class)` will not take effect as you expected.

server/odc-service/src/main/java/com/oceanbase/odc/service/config/OrganizationConfigService.java
```diff
@@ -128,6 +142,9 @@ public List<Configuration> queryListDefault() {
     public List<Configuration> batchUpdate(@NotNull Long organizationId, @NotNull Long userId,
             @NotEmpty List<Configuration> configurations) {
         validateConfiguration(configurations);
+        String customDataSourceKey = getCustomDataSourceKey(configurations);
+        migrateExistedDataSourcePassword(organizationId, customDataSourceKey);
```
MarkPotato777, line 146: Why `batchUpdate` has no transaction control while `migrateExistedDataSourcePassword` has? Does it make sense? What if the above updating config fail? Will `migrateExistedDataSourcePassword` rollback?

server/odc-service/src/main/java/com/oceanbase/odc/service/config/model/Configuration.java
```diff
@@ -64,7 +69,9 @@ public OrganizationConfigEntity convert2DO(Long organizationId, Long userId) {
         entity.setLastModifierId(userId);
         entity.setCreatorId(userId);
         entity.setKey(this.key);
-        entity.setValue(this.value);
+        entity.setValue(Objects.equals(entity.getKey(), DEFAULT_CUSTOM_DATA_SOURCE_ENCRYPTION_KEY)
```
MarkPotato777, line 72: Why `DEFAULT_CUSTOM_DATA_SOURCE_ENCRYPTION_KEY`??

yiminpeng: The value of the data source key needs to be obfuscated and then stored, so that's it. What exactly is your problem?

server/odc-service/src/main/java/com/oceanbase/odc/service/config/OrganizationConfigService.java
yiminpeng: Also need.

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
yiminpeng: changed to `TransactionTemplate`.

server/odc-migrate/src/main/java/com/oceanbase/odc/migrate/jdbc/common/V4349OrganizationSecretMigrate.java
yiminpeng: 1. throw a RuntimeException.
2. Use batch updates and ensure that only both success and failure are allowed.

MarkPotato777 (approved):

guowl3 (approved):

PR #4342, opened 2025-04-09
fix(sql check ): Getting the number of sql affected rows does not support the old ob version

zijiacj:
#### What type of PR is this?
type-bug

The result set of the sql plan of the old version ob is inconsistent with that of the new version ob, resulting in failure to obtain the number of affected rows

Fixes #

```docs

```


ysjemmm (approved):
LGTM

MarkPotato777 (approved):

PR #4344, opened 2025-04-09
fix(riskLevel): the riskLevelDescriber under the two databases can be distinguished 

ysjemmm:
…ject and database with the same name can be distinguished

#### What type of PR is this?
type-bug

The `RiskLevelDescriber` of `two databases` cannot be completely distinguished based on the following conditions: `databaseName`, `projectName`, `environmentID`, `environmentName`, and `taskType`. You may need to add a unique identification `databaseId`

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/regulation/approval/ApprovalFlowConfigSelector.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/regulation/approval/ApprovalFlowConfigSelector.java
@@ -32,7 +32,7 @@ import com.oceanbase.odc.service.regulation.risklevel.RiskLevelService;
 import com.oceanbase.odc.service.regulation.risklevel.model.RiskDetectRule;
 import com.oceanbase.odc.service.regulation.risklevel.model.RiskLevel;
 import com.oceanbase.odc.service.regulation.risklevel.model.RiskLevelDescriber;
-import com.oceanbase.odc.service.regulation.risklevel.model.RiskLevelIdentifier;
+import com.oceanbase.odc.service.regulation.risklevel.model.RiskLevelDescriberIdentifier;
 
 /**
  * @Author: Lebie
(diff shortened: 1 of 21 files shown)


LioRoger (dismissed):
i have one more question, database in one db can have two or more database id, what risk level describer process these records

ysjemmm:
> i have one more question, database in one db can have two or more database id, what risk level describer process these records
The database id uniquely identifies the database. I understand that a riskLevelDescriber is created for each database, and the corresponding relation of the database -> riskLevelDescriber is found by the database id

server/odc-service/src/main/java/com/oceanbase/odc/service/regulation/risklevel/model/RiskLevelDescriber.java
```diff
@@ -77,6 +79,7 @@ public String describe(ConditionExpression expression) {
 
     public static RiskLevelDescriber of(@NonNull Database db, String taskType) {
         RiskLevelDescriber describer = RiskLevelDescriber.builder()
+                .databaseId(db.getId())
```
MarkPotato777, line 82: What's the function if the user cannot config database id in risk detect rules on the frontend?

ysjemmm: > What's the function if the user cannot config database id in risk detect rules on the frontend?

The front-end does not need to configure the database id in the risk detection rule. The id is only to ensure that `RiskLevelDescriber` can be used to locate which `database` the `describer` belongs to

server/odc-service/src/main/java/com/oceanbase/odc/service/regulation/approval/ApprovalFlowConfigSelector.java
```diff
@@ -76,13 +77,13 @@ public RiskLevel select(RiskLevelDescriber describer) {
     }
 
     @SkipAuthorize("internal usage")
-    public Map<RiskLevelDescriber, RiskLevel> batchSelect(Collection<RiskLevelDescriber> describers) {
-        if (CollectionUtils.isEmpty(describers)) {
+    public Map<RiskLevelIdentifier, RiskLevel> batchSelect(Collection<RiskLevelIdentifier> riskLevelIdentifiers) {
```
MarkPotato777, line 80: It's hard to understand. Why do I need to pass a 'risk level identifer' to select a risk level?

server/odc-service/src/main/java/com/oceanbase/odc/service/regulation/risklevel/model/RiskLevelIdentifier.java
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * @Author: ysj
+ * @Date: 2025/4/10 16:29
+ * @Since: 4.3.4
+ * @Description:
+ */
+@Data
+@Accessors(chain = true)
+@EqualsAndHashCode
+public class RiskLevelIdentifier {
```
MarkPotato777, line 32: Hard to understand. How could  'database id' & 'risk level describer' identify a risk level?

guowl3 (approved):
lgtm

MarkPotato777 (approved):

PR #4347, opened 2025-04-09
security: lack permission check when listing another user's resources

MarkPotato777:
#### What type of PR is this?

type-security
When listing data sources of another user, it doesn't check if current user have the read permission to that user.
This PR added this permission check.

Fixes #

```docs

```


LioRoger (approved):
lgtm

PR #4348, opened 2025-04-09
fix(notification): fix metadata and message resending

LuckyPickleZZ:
1. Sometimes a message could be sent twice. So we should fail it when it's timeout.
2. There is no message for  online schema change and data archive .


MarkPotato777 (approved):
LGTM

LioRoger (approved):
lgtm

PR #4349, opened 2025-04-10
fix(database): Supports synchronization of individual space to all current organization databases

ysjemmm:
#### What type of PR is this?
type-bug

For individual space, fixed a bug that would not sync all databases under the current organization


server/odc-service/src/main/java/com/oceanbase/odc/metadb/connection/DatabaseRepository.java
```diff
@@ -49,6 +49,9 @@ public interface DatabaseRepository extends JpaRepository<DatabaseEntity, Long>,
     List<DatabaseEntity> findByProjectIdInAndExistedAndObjectSyncStatusNot(Collection<Long> projectIds, Boolean existed,
             DBObjectSyncStatus dbObjectSyncStatus);
 
+    List<DatabaseEntity> findByOrganizationIdAndExistedAndObjectSyncStatusNot(Long organizationId, Boolean existed,
```
LioRoger, line 52: check if this sql has index

ysjemmm: > check if this sql has index

care

MarkPotato777 (approved):
LGTM

guowl3 (approved):

PeachThinking (approved):
lgtm

PR #4352, opened 2025-04-10
fix(config): fix the query limit does not take effect as expected

yiminpeng:
#### What type of PR is this?
fix

1. add the bottom line strategy of query limit
2. fix the `DumpTask` misquote configuration

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/config/OrganizationConfigService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/config/OrganizationConfigService.java
@@ -156,6 +156,9 @@ public class OrganizationConfigService {
         String maxQueryLimit = configMap.get(DEFAULT_MAX_QUERY_LIMIT).getValue();
         String queryLimit = configMap.get(DEFAULT_QUERY_LIMIT).getValue();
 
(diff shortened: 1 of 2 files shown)


LuckyPickleZZ (dismissed):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/config/OrganizationConfigService.java
```diff
@@ -156,6 +156,9 @@ private void validateConfiguration(List<Configuration> configurations) {
         String maxQueryLimit = configMap.get(DEFAULT_MAX_QUERY_LIMIT).getValue();
         String queryLimit = configMap.get(DEFAULT_QUERY_LIMIT).getValue();
 
+        if (Integer.parseInt(queryLimit) < 0 || Integer.parseInt(maxQueryLimit) < 0) {
```
MarkPotato777, line 159: Why could it be negative? If it could be negative, is it possible that it exceeds Int?

MarkPotato777: Why not use your 'config meta' to validate this?

MarkPotato777 (approved):

guowl3 (approved):

PeachThinking (approved):
lgtm

PR #4358, opened 2025-04-11
fix(Script): add duplicated filename tag before suffix of filename

CHLK:
#### What type of PR is this?
type-bug

 add duplicated filename tag before suffix of filename

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/script/ScriptService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/script/ScriptService.java
@@ -367,7 +367,7 @@ public class ScriptService {
     }
 
     private String addTagAfterFilename(String fileName, String tag) {
-        int lastDotIndex = fileName.lastIndexOf(".");
+        int lastDotIndex = fileName.indexOf(".");
         if (lastDotIndex != -1) {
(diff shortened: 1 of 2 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/script/ScriptService.java
```diff
@@ -364,4 +365,13 @@ public ResponseEntity<InputStreamResource> batchDownload(List<Long> scriptIds) t
         return WebResponseUtils.getFileAttachmentResponseEntity(
                 new InputStreamResource(new FileInputStream(zipFileStr)), new File(zipFileStr).getName());
     }
+
+    private String addTagAfterFilename(String fileName, String tag) {
```
MarkPotato777, line 369: The former format is like 'name(1)', but now it's like 'name 1'. Is it expected?

CHLK: yes，tester wishes to change to this format

MarkPotato777 (dismissed):
LGTM

LioRoger (approved):
lgtm

PR #4360, opened 2025-04-14
fix(partition-plan): the task generate time string value error

kiko-art:
#### What type of PR is this?
type-bug

 the task generate time string value error

Fixes #

```docs

```


LuckyPickleZZ (approved):
lgtm

PR #4361, opened 2025-04-14
fix(object-storage): use doesBucketExist for gcs

LuckyPickleZZ:
#### What type of PR is this?
type-bug

`AmazonS3#doesBucketExistV2` is not compatible with google cloud storage , we should use `doesBucketExist` instead.


MarkPotato777 (approved):
LGTM

CHLK (approved):
LGTM

PR #4362, opened 2025-04-14
fix(db-browser):the ddl of produce has two semicolons

PeachThinking:

Fixes #

```docs

```


MarkPotato777 (approved):
LGTM

PR #4363, opened 2025-04-14
fix(collaboration): access denied when synchronizing obcloud organization

LuckyPickleZZ:
#### What type of PR is this?
type-bug

Synchronizing organization would fail for access denied.


PeachThinking (approved):
lgtm

MarkPotato777 (approved):

PR #4370, opened 2025-04-16
fix(resource): let resource destroy concurrent safe

LioRoger:
#### What type of PR is this?

type-bug
destroy may invoked multi times, cause it may be invoked by multi ODC node.
add database lock and status check to avoid this issue.

Fixes #

```docs

```


guowl3 (approved):
lgtm

ungreat (approved):

PR #4372, opened 2025-04-16
fix(flow): can view project when locate at applyDatabasePermission or applyTablePermission 

ysjemmm:
#### What type of PR is this?
type-bug


MarkPotato777 (approved):
LGTM

PR #4373, opened 2025-04-16
fix(session): Supports fuzzy matching of table names and view names

ysjemmm:
#### What type of PR is this?
type-bug

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/db/DBIdentitiesService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/db/DBIdentitiesService.java
@@ -24,6 +24,7 @@ import java.util.TreeMap;
 import org.apache.commons.collections4.CollectionUtils;
 import org.springframework.stereotype.Service;
 
+import com.oceanbase.odc.common.util.ObjectUtil;
 import com.oceanbase.odc.common.util.StringUtils;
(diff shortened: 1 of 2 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/db/DBIdentitiesService.java
```diff
@@ -36,19 +37,23 @@
...
         }
         if (types.contains(DBObjectType.TABLE)) {
-            listTables(schemaAccessor, all);
+            listTables(schemaAccessor, identityNameLike, all);
         }
         if (types.contains(DBObjectType.EXTERNAL_TABLE)) {
-            listExternalTables(schemaAccessor, all);
+            listExternalTables(schemaAccessor, identityNameLike, all);
         }
         if (types.contains(DBObjectType.MATERIALIZED_VIEW)) {
```
zijiacj, line 58: The materialized view may also need to add a fuzzy match of name

ysjemmm: > The materialized view may also need to add a fuzzy match of name

ok

MarkPotato777 (approved):
LGTM

guowl3 (approved):

zijiacj (approved):
lgtm

PR #4384, opened 2025-04-18
feat(shecule): add batch terminal

ungreat:
#### What type of PR is this?
type-feature

support batch terminal schedule and schedule task;
support batch terminal flow instance;

Fixes #

```docs

```

### DIFF
--- a/server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/FlowInstanceController.java
+++ b/server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/FlowInstanceController.java
@@ -260,13 +260,13 @@ public class FlowInstanceController {
     }
 
     @ApiOperation(value = "getBatchCancelResult", notes = "获取批量终止结果")
-    @RequestMapping(value = "/asyncResult", method = RequestMethod.GET)
+    @RequestMapping(value = "/asyncCancelResult", method = RequestMethod.GET)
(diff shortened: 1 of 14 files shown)


guowl3 (dismissed):

server/odc-server/src/main/java/com/oceanbase/odc/server/web/controller/v2/FlowInstanceController.java
```diff
@@ -251,4 +253,22 @@ public SuccessResponse<PartitionPlanConfig> getPartitionPlan(@PathVariable Long
...
+
+    @ApiOperation(value = "getBatchCancelResult", notes = "获取批量终止结果")
+    @RequestMapping(value = "/asyncCancelResult", method = RequestMethod.GET)
+    public SuccessResponse<List<BatchTerminateFlowResult>> getBatchCancelResult(String terminateId) {
+        return Responses.single(flowInstanceService.getBatchCancelResult(terminateId));
+    }
+
+    @ApiOperation(value = "getBatchCancelLog", notes = "获取批量终止日志")
+    @RequestMapping(value = "/asyncCancelLog", method = RequestMethod.GET)
+    public SuccessResponse<String> getBatchCancelLog(String terminateId) {
```
LioRoger, line 270: i don't agree with this implementation,  cancel should be a  re-entry operation. if one node down,  what this action has done could not inferred by database log,
by the way, there is not interface passed by argument of terminated id instead of flow instance id

LioRoger (approved):

PR #4386, opened 2025-04-18
fix(db browser):desc view without schema

zijiacj:
#### What type of PR is this?
type-bug

The absence of the `schema `leads to an error reporting that the view does not exist

Fixes #

```docs

```


libs/db-browser/src/main/java/com/oceanbase/tools/dbbrowser/schema/oracle/OBOracleSchemaAccessor.java
```diff
@@ -535,9 +535,13 @@ private DBView fillColumnInfoByDesc(DBView view) {
                     e.getMessage());
         }
         OracleSqlBuilder sb = new OracleSqlBuilder();
-        sb.append("desc ");
-        sb.identifier(view.getViewName());
 
+        sb.append("desc ");
+        if (StringUtils.isNotBlank(view.getSchemaName())) {
```
MarkPotato777, line 540: Directly use `SqlBuilder#identifier`, which has already included the empty schema name logic.

ysjemmm (approved):
lgtm

MarkPotato777 (approved):
LGTM

PR #4394, opened 2025-04-22
enhancement(config): organization configuration adaptive external integration

yiminpeng:
#### What type of PR is this?

enhancement

1. When updating the organization configuration, the corresponding configuration integration is updated synchronously.
2. Solve the problem of inconsistent multiple nodes in the cloud environment. 

```docs

```

### DIFF
--- a/server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_10__alter_iam_organization_add_column.sql
+++ b/server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_10__alter_iam_organization_add_column.sql
@@ -3,6 +3,6 @@ alter table `iam_organization`
 add column `secret_new` varchar(256) default null
 comment 'confused organization secret use caesar';
 
-alter table `iam_organization` MODIFY `secret` varchar(256) null;
+alter table `iam_organization` MODIFY `secret` varchar(256) default null;
 -- copy secret to iam_organization.secret_new
 update `iam_organization` set `secret_new` = `secret`;
--- a/server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_10__alter_iam_organization_add_column.sql
+++ b/server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_10__alter_iam_organization_add_column.sql
@@ -3,5 +3,6 @@ alter table `iam_organization`
 add column `obfuscated_secret` varchar(256) default null
 comment 'organization secret after migrate';
 
+alter table `iam_organization` MODIFY `secret` varchar(256) null;
 -- copy secret to iam_organization.obfuscated_secret
 update `iam_organization` set `obfuscated_secret` = `secret`;
--- a/server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_10__alter_iam_organization_add_column.sql
+++ b/server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_10__alter_iam_organization_add_column.sql
@@ -1,7 +1,7 @@
--- add column custom_secret to iam_organization
+-- add column obfuscated_secret to iam_organization
 alter table `iam_organization`
-add column `custom_secret` varchar(256) default null
+add column `obfuscated_secret` varchar(256) default null
 comment 'organization secret after migrate';
 
 -- copy secret to iam_organization.custom_secret
-update `iam_organization` set `custom_secret` = `secret`;
+update `iam_organization` set `obfuscated_secret` = `secret`;
--- a/server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_10__alter_iam_organization_add_column.sql
+++ b/server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_10__alter_iam_organization_add_column.sql
@@ -1,0 +1,7 @@
+-- add column secret_before_migrate to iam_organization
+alter table `iam_organization`
+add column `secret_before_migrate` varchar(256) default null
+comment 'organization secret before migrate';
+
+-- copy secret to iam_organization.secret_before_migrate
+update `iam_organization` set `secret_before_migrate` = `secret`;
--- a/server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_10__alter_iam_organization_add_column.sql
+++ b/server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_10__alter_iam_organization_add_column.sql
@@ -1,7 +1,7 @@
--- add column secret_before_migrate to iam_organization
+-- add column custom_secret to iam_organization
 alter table `iam_organization`
-add column `secret_before_migrate` varchar(256) default null
-comment 'organization secret before migrate';
+add column `custom_secret` varchar(256) default null
+comment 'organization secret after migrate';
 
--- copy secret to iam_organization.secret_before_migrate
-update `iam_organization` set `secret_before_migrate` = `secret`;
+-- copy secret to iam_organization.custom_secret
+update `iam_organization` set `custom_secret` = `secret`;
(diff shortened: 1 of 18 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/service/config/OrganizationConfigService.java
```diff
@@ -217,7 +224,7 @@ private Map<String, Configuration> internalQuery(Long organizationId) {
                 .collect(Collectors.toMap(Configuration::getKey, c -> c));
     }
 
-    private void migrateExistedDataSourcePassword(Long organizationId, String customKey) {
+    private void migrateExistedSecretAndDataSourcePassword(Long organizationId, String customKey) {
```
ungreat, line 227: seem not just migrate datasource ?  This method name can cause misunderstandings

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
```diff
@@ -842,8 +842,9 @@ public ConnectionConfig getEncryptedConfig(@NotNull ConnectionConfig decryptedCo
     }
 
     @SkipAuthorize("internal usage")
-    public void batchUpdateConnectionConfig(List<ConnectionEntity> connections) {
+    public int batchUpdateConnectionConfig(List<ConnectionEntity> connections) {
```
ungreat, line 845: This method is kind of  wired. `repository.saveAll` actual calls the save method within a for loop,  it is different from batchUpdate semantically. Additionally, why do we need to return this int?

server/odc-service/src/main/java/com/oceanbase/odc/service/encryption/EncryptionFacadeImpl.java
```diff
@@ -106,8 +106,13 @@ public TextEncryptor userEncryptor(Long userId, String salt) {
     public TextEncryptor organizationEncryptor(Long organizationId, String salt) {
         Organization organization = organizationService.get(organizationId).orElseThrow(
                 () -> new NotFoundException(ResourceType.ODC_ORGANIZATION, "organizationId", organizationId));
-        String secret = Caesar.decode(organization.getSecret(), 8);
-        return getEncryptor(secret, salt);
+        // use new secret if secret has been migrated
+        if (organizationService.isOrganizationSecretMigrated(organizationId)
```
ungreat, line 110: why we need this if condition here?

server/odc-service/src/main/java/com/oceanbase/odc/metadb/iam/OrganizationEntity.java
```diff
@@ -84,6 +84,12 @@ public class OrganizationEntity {
     @Column(name = "secret", nullable = false)
     private String secret;
```
ungreat, line 85: do we still need this field in the entity？ or @Column to the new field ? 
we do not need two secret field in this entity.

server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_10__alter_iam_organization_add_column.sql
```diff
@@ -0,0 +1,7 @@
+-- add column custom_secret to iam_organization
+alter table `iam_organization`
+add column `custom_secret` varchar(256) default null
+comment 'organization secret after migrate';
+
+-- copy secret to iam_organization.custom_secret
+update `iam_organization` set `custom_secret` = `secret`;
```
ungreat, line 7: may comment `secret` is deprecate and remove the filed next version. 
`custom_secret` seemly not appropriate，it's generated by random defaultly.

server/odc-service/src/main/java/com/oceanbase/odc/service/git/GitIntegrationService.java
```diff
@@ -121,6 +122,26 @@ public GitRepository update(@NotNull Long projectId, @NotNull Long repoId, @NotN
...
 
+    @Transactional(rollbackFor = Exception.class)
+    @SkipAuthorize("odc internal usage")
+    public int updateGitRepoPersonalToken(@NotNull Long organizationId, String customSecret) {
+        List<GitRepositoryEntity> entities = gitRepoRepository.findByOrganizationId(organizationId);
+        if (entities.isEmpty()) {
+            return 0;
+        }
+        List<GitRepositoryEntity> saved = entities.stream().peek(entity -> {
+            TextEncryptor encryptor = getEncryptor(entity.getOrganizationId(), entity.getSalt());
```
ungreat, line 133: It actually implies that the organization‘s secret field has not been updated.  This implicit logic will make the method not robust enough. It is recommend that passing both oldSecret and newSecret as the signature of the method. And may we can have a method to migrate a encode text from oldSecret to the new one.

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/OrganizationService.java
```diff
@@ -170,4 +173,10 @@ public Optional<Organization> getByIdentifier(@NotNull String identifier) {
     public Optional<Organization> get(@NotNull Long id) {
         return organizationRepository.findById(id).map(Organization::ofEntity);
     }
+
+    public boolean isOrganizationSecretMigrated(@NotNull Long id) {
```
ungreat, line 177: do we need this condition?

server/odc-service/src/main/java/com/oceanbase/odc/service/iam/model/Organization.java
```diff
@@ -54,6 +54,8 @@ public class Organization implements Serializable, SecurityResource, SingleOrgan
     private String displayName;
     @JsonIgnore
     private String secret;
+    @JsonIgnore
+    private String customSecret;
```
ungreat, line 58: same as entity, we need two secret field here ?

server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_10__alter_iam_organization_add_column.sql
```diff
@@ -0,0 +1,7 @@
+-- add column custom_secret to iam_organization
+alter table `iam_organization`
+add column `custom_secret` varchar(256) default null
+comment 'organization secret after migrate';
```
ungreat, line 4: may comment it's not plaintext

server/odc-service/src/main/java/com/oceanbase/odc/service/config/OrganizationConfigService.java
yiminpeng: `migrateExistedEncryptionRelated2OrganizationSecret` maybe

server/odc-service/src/main/java/com/oceanbase/odc/service/connection/ConnectionService.java
yiminpeng: Not necessary, already moved.

server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_10__alter_iam_organization_add_column.sql
yiminpeng: may `obfuscated_secret` could be a good name.

server/odc-migrate/src/main/resources/migrate/common/V_4_3_4_10__alter_iam_organization_add_column.sql
yiminpeng: got it.

server/odc-service/src/main/java/com/oceanbase/odc/service/git/GitIntegrationService.java
yiminpeng: may we can use a better method name.

MarkPotato777 (approved):

guowl3 (approved):

PR #4400, opened 2025-04-25
fix(database): distinguish the case of schemaName

ysjemmm:
#### What type of PR is this?
type-bug

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/metadb/connection/DatabaseSpecs.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/metadb/connection/DatabaseSpecs.java
@@ -27,6 +27,7 @@ import org.springframework.data.jpa.domain.Specification;
 import com.oceanbase.odc.common.jpa.SpecificationUtil;
 import com.oceanbase.odc.common.util.StringUtils;
 import com.oceanbase.odc.core.shared.constant.ConnectType;
+import com.oceanbase.odc.service.common.util.SpringContextUtil;
 import com.oceanbase.odc.service.connection.database.model.DatabaseType;
 
 /**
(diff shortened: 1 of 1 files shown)


server/odc-service/src/main/java/com/oceanbase/odc/metadb/connection/DatabaseSpecs.java
```diff
@@ -41,8 +42,17 @@ public static Specification<DatabaseEntity> idIn(Collection<Long> ids) {
     }
 
     public static Specification<DatabaseEntity> nameLike(String name) {
+        if (SpringContextUtil.isActive("clientMode")) {
```
LioRoger, line 45: clinet mode is not enough to infer what database it is

guowl3 (approved):

server/odc-service/src/main/java/com/oceanbase/odc/metadb/connection/DatabaseSpecs.java
```diff
@@ -45,6 +45,11 @@ public static Specification<DatabaseEntity> nameLike(String name) {
                 : builder.like(root.get("name"), "%" + StringUtils.escapeLike(name) + "%");
     }
 
+    public static Specification<DatabaseEntity> nameLikeWithoutCase(String name) {
+        return (root, query, builder) -> StringUtils.isBlank(name) ? builder.conjunction()
+                : builder.like(builder.lower(root.get("name")), "%" + StringUtils.escapeLike(name.toLowerCase()) + "%");
```
LioRoger, line 50: assume database is case sensitive, but value is stored in uppercase,  does this method can matcher that?

ysjemmm: > assume database is case sensitive, but value is stored in uppercase, does this method can matcher that

Now it is necessary to modify all of them to "case-insensitive".

PR #4404, opened 2025-04-27
fix(sql): Support automatic completion based on schemaName, tableNameLike or viewNameLike

ysjemmm:
#### What type of PR is this?
type-bug


ysjemmm:
-

PR #4406, opened 2025-04-28
enhancement(materialized view):Display all the databases when creating materialized views or views

zijiacj:
#### What type of PR is this?
type-enhancement

1. Previously, odc would send multiple sql requests to obtain tables based on the number of schemas. Currently, it has been changed to sending only one sql request
2. Previously, if there were no base tables in the schema, the schema would not be displayed to the user. Currently, it has been changed to be displayed as well

Fixes #

```docs

```


MarkPotato777 (approved):
LGTM

LioRoger (approved):
lgtm

PR #4407, opened 2025-04-28
fix(flow):  apply project tickets not shown when it's under approval

MarkPotato777:
#### What type of PR is this?

type-bug
1. fix the bug that the Apply Project tickets are not shown when created and under approval.
2. add some logs for logical table extract tasks.

Fixes #

```docs

```

### DIFF
--- a/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
+++ b/server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
@@ -670,8 +670,7 @@ public class FlowInstanceService {
             // APPLY_PROJECT_PERMISSION, then return empty directly
             if (CollectionUtils.isEmpty(joinedProjectIds)
                     && authenticationFacade.currentOrganization().getType() == OrganizationType.TEAM
-                    && !taskTypes.contains(
-                            TaskType.APPLY_PROJECT_PERMISSION)) {
+                    && !taskTypes.contains(TaskType.APPLY_PROJECT_PERMISSION)) {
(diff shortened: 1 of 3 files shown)


zijiacj (approved):
lgtm

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
```diff
@@ -639,11 +639,13 @@ private Page<FlowInstanceEntity> listAll(@NotNull Pageable pageable, @NotNull Qu
...
                 .and(FlowInstanceViewSpecs.idEquals(targetId))
                 .and(FlowInstanceViewSpecs.groupByIdAndTaskType());
+        Set<TaskType> taskTypes;
         if (params.getType() != null) {
+            taskTypes = Sets.newHashSet(params.getType());
             specification = specification.and(FlowInstanceViewSpecs.taskTypeEquals(params.getType()));
         } else {
             // Task type which will be filtered independently
-            List<TaskType> types = Arrays.asList(
+            taskTypes = Sets.newHashSet(
```
LioRoger, line 648: for constant sets, recommend declare it as static and unmodified collection

LioRoger (approved):

guowl3 (approved):

server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowInstanceService.java
MarkPotato777: Nice suggestion, and I will take care next time.

PR #4412, opened 2025-04-29
fix(task): k8s parse resource identifier failed

LioRoger:
#### What type of PR is this?

type-bug
if use StringUtils, splitByWholeSeparator should be used if we want split by characters but not a single character

Fixes #

```docs

```


guowl3 (approved):
lgtm

PR #4414, opened 2025-04-29
fix(audit): Fix the database operation type in the audit event that does not record database info

yiminpeng:
#### What type of PR is this?

fix

Fix the database operation type in the audit event that does not record database information

```docs

```


MarkPotato777 (approved):
LGTM

PR #4415, opened 2025-04-30
fix(db browser): fix the syntax error of fuzzy query

zijiacj:
#### What type of PR is this?
type-bug

1. fix the sql error of the fuzzy query in the materialized view
2. modify the method signature of the materialized view
3. add tests for fuzzy query

Fixes #

```docs

```


MarkPotato777 (approved):
LGTM

PR #4416, opened 2025-04-30
fix(session): fix SQL execute generate result failed caused by material view

yiminpeng:
#### What type of PR is this?
fix

Fixed the issue where `queryTableOrViewData.getMorResults` failed due to duplicate keys in the `toMap` method, which in turn affected the failure of some sql console rule.

```docs

```


zijiacj (dismissed):
lgtm

LioRoger (approved):

PR #4418, opened 2025-05-06
fix(schedule): To stat the count of all state schedule for each scheduleType type

ysjemmm:
#### What type of PR is this?
type-bug

Add a field to count the number of all state schedule for each scheduleType type


guowl3 (approved):

zijiacj (approved):
lgtm

PR #4420, opened 2025-05-06
fix(db browser):failed to get table in doris

zijiacj:
#### What type of PR is this?
type-bug

odc is not compatible with the `duplicate index`, resulting in failure

Fixes #

```docs

```


MarkPotato777 (approved):
LGTM

PR #4432, opened 2025-05-13
fix(objects): support awscn

ungreat:
#### What type of PR is this?
type-feature

support awscn

Fixes #

```docs

```


LioRoger (approved):
recommend collect all this code into one place, let modifier know which code should be changed if there is new provider should be added

PR #4436, opened 2025-05-14
feat(storage): introduce azure client

LioRoger:
#### What type of PR is this?

type-feature
introduce azure blob service client

Fixes #

```docs

```


pom.xml
```diff
@@ -114,7 +114,7 @@
         <jettison.version>1.5.4</jettison.version>
         <!-- ob-loader-dumper involved -->
         <hadoop-common.version>3.3.6</hadoop-common.version>
-        <netty.version>4.1.94.Final</netty.version>
+        <netty.version>4.1.108.Final</netty.version>
```
MarkPotato777, line 117: Why do we need to upgrade the netty?

LioRoger: 1. Azure has dependency required, 4.1.94.Final is not compatible with azure blob client
2. 4.1.94.Final has security issue
3. so i choose upgrade netty version instead of degrade azure version

PR #4441, opened 2025-05-15
fix(connection): The Oracle datasource's connect/socket timeout is not set correctly which may cause to connection leak.

MarkPotato777:
#### What type of PR is this?

type-bug
The Oracle datasource's connect/socket timeout is not set correctly, which may cause to connection leak.

Fixes #

```docs

```


guowl3 (approved):

LuckyPickleZZ (approved):
lgtm

PR #4447, opened 2025-05-23
fix(task): detail task may failed by log file expired

ungreat:
#### What type of PR is this?
type-bug

ignore log file NotFoundException when detail task

Fixes #

```docs

```


server/odc-service/src/main/java/com/oceanbase/odc/service/flow/FlowTaskInstanceService.java
```diff
@@ -798,4 +801,13 @@ private void setDownloadUrlsIfNecessary(Long taskId, List<? extends FlowTaskResu
         }
     }
 
+    private TaskDownloadUrls getTaskDownloadUrls(Long taskId) {
```
zijiacj, line 804: It is recommended to use @Nullable to mark the return value or use Option to handle the situation where the return value is null. This is more intuitive for the caller.

zijiacj (approved):
lgtm

PR #4449, opened 2025-05-27
fix(osc): remove increment progress check

LioRoger:
#### What type of PR is this?

type-bug

Oms removed increment progress compute, that will cause ODC osc check if OMS if ready failed.

Fixes #

```docs

```


LuckyPickleZZ (commented):
lgtm

LuckyPickleZZ (approved):

PR #4454, opened 2025-05-29
feat(flow):  support cancel sql plan flow task  when executing rollback sql generation

LioRoger:
#### What type of PR is this?

type-feature
previous logic will not permit cancel in flowing status
1. flow userTask and serviceTask in created status
2. if user task has done but service task has not created
this patch let those status be canceled by flowing change
1. remove flow process instance if not scheduled
2. check if flow has be canceled before execute service task

by the way, sql plan task is not interrupted in previous implement. introduce breakpoint to let it know flow has been canceled

Fixes #

```docs

```


guowl3 (approved):

MarkPotato777 (approved):
LGTM

zijiacj (approved):
lgtm

PR #4456, opened 2025-05-29
fix(security): upgrade python and pip version

ungreat:
#### What type of PR is this?
type-bug

upgrade python and pip version

Fixes #

```docs

```


LioRoger (approved):
lgtm

PR #4462, opened 2025-06-03
fix(pl debug): add configuration parameters for pl debug

zijiacj:
#### What type of PR is this?
type-bug

The default timeout for pl dubug set by odc is 2 minutes, but in fact it exceeded 2 minutes, resulting in an error. 
Make the timeout of pl debug configurable for certain time-consuming procedures

Fixes #

```docs

```


MarkPotato777 (approved):
LGTM

MarkPotato777 (approved):

PR #4475, opened 2025-06-11
fix(log): getTaskResult may failed by log file expired

ungreat:
#### What type of PR is this?
type-bug

patch from 202502bp

Fixes #

```docs

```


MarkPotato777 (approved):
LGTM

